PR #59, opened 2023-07-23
Remove paths from configuration

jdeschamps:
- Paths (train, val, predict) are now passed as parameters to the functions
- No model path in the configuration
- Model path can be used to create an Engine
- Mean/std are taken from configuration only

### DIFF
--- a/src/careamics_restoration/config/data.py
+++ b/src/careamics_restoration/config/data.py
@@ -24,6 +24,7 @@ class SupportedExtensions(str, Enum):
 
     TIFF = "tiff"
     TIF = "tif"
+    NPY = "npy"
 
     @classmethod
     def _missing_(cls, value: object):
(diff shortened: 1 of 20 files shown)


src/careamics_restoration/config/data.py
```diff
@@ -31,8 +24,6 @@ class SupportedExtensions(str, Enum):
 
     TIFF = "tiff"
     TIF = "tif"
-    NPY = "npy"  # TODO remove numpy after we reupload the dataset as zarr
```
CatEek, line 34: Is there any reason to remove it now, while it's still present in the dataset class ?

CatEek (commented):
LGTM otherwise

src/careamics_restoration/config/data.py
jdeschamps: You caught me, I really wanted to get rid of this. But makes no sense because we are still using it! 

Added back here: https://github.com/CAREamics/careamics-restoration/pull/59/commits/150bbc2cf9aa7ea743bd717924b03fd52c0d330f

PR #62, opened 2023-07-25
Bioimage export

mese79:


src/careamics_restoration/bioimage/io.py
```diff
@@ -0,0 +1,102 @@
...
+from bioimageio.core import load_resource_description
+from bioimageio.core.build_spec import build_model
+from bioimageio.spec.model.raw_nodes import Model
+
+from careamics_restoration.config.config import (
+    Configuration,
+)
+
+PYTORCH_STATE_DICT = "pytorch_state_dict"
+CAREAMICS_CONFIG = "careamics_config.yml"
```
jdeschamps, line 14: Where is this used? And why?

src/careamics_restoration/bioimage/rdfs/n2v.yaml
```diff
@@ -0,0 +1,29 @@
...
+  -
+    - kwargs:
+        axes: yx
+        mean: [46912.574]
+        mode: fixed
+        std: [16847.809]
+      name: zero_mean_unit_variance
+output_axes: ["bcyx"]
+postprocessing:
+  -
```
jdeschamps, line 23: is the single "-" a typo?

src/careamics_restoration/engine.py
```diff
@@ -524,3 +530,77 @@ def __del__(self) -> None:
...
+            f"{self.cfg.experiment_name}_best.pth"
+        ).absolute()
+        checkpoint = torch.load(checkpoint_path)
+        # save the model state_dict separately.
+        weight_path = workdir.joinpath("model_weights.pth")
+        torch.save(checkpoint["model_state_dict"], weight_path)
+        # get in/out samples' files
+        test_inputs, test_outputs = self._get_sample_io_files()
+
+        specs = get_default_model_specs(self.cfg.algorithm.loss)
```
jdeschamps, line 558: `algorithm.loss` is a bad proxy for the algorithm name... This will undoubtedly break soon (not all loss will be named after an algorithm, we might also support other losses).

Currently we only support n2v, so I would just simplify everything and then find a better way once we have multiple algorithms.

jdeschamps: By simplify everything, I mean build the rdf directly for N2V, without trying to support other algorithms. That means no "name" parameter passed to the functions.

src/careamics_restoration/bioimage/io.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # add documentation (md file)
+    doc = Path(__file__).parent.joinpath("docs").joinpath(f"{name}.md")
+    if doc.exists() > 0:
+        rdf["documentation"] = str(doc.absolute())
+
+    return rdf
+
+
+def get_default_model_specs(name: str) -> dict:
```
jdeschamps, line 42: I am not a fan of loading a default rdf and modifying it, because the default has then a lot of details that are potentially wrong and might find their way into the exported rdf.

Also, why bother loading something that we can have as a dict returned by a function?

So I would just pack the really important information into a dict returned by a function `get_default_n2v_rdf`

src/careamics_restoration/engine.py
```diff
@@ -524,3 +530,77 @@ def __del__(self) -> None:
...
+        if model_specs is not None:
+            specs.update(model_specs)
+
+        specs.update(
```
jdeschamps, line 562: We should add a tag `2D` or `3D`, depending on the axes.

jdeschamps: Also axes should be set by the actual `Engine`, not be default.

src/careamics_restoration/bioimage/docs/n2v.md
```diff
@@ -0,0 +1,5 @@
+## Noise2Void
```
jdeschamps, line 1: Citation could also just be a dict returned by `get_n2v_citation`

tests/bioimage/test_io.py
```diff
@@ -0,0 +1,53 @@
...
+    torch.save(checkpoint, checkpoint_path)
+    # output zip file
+    zip_file = tmp_path.joinpath("tmp_model.zip")
+    # export the model (overriding the weight_uri)
+    engine.save_as_bioimage(zip_file)
+
+    # put zip file in the cache for the import test
+    request.config.cache.set("bioimage_model", str(zip_file))
+
+    assert zip_file.exists()
```
jdeschamps, line 42: The exported bioimage.io.zip should also be validated!

jdeschamps (changes requested):
It's getting there!

One additional comment: currently it doesn't seem that one can start the engine from a path pointing to a bioimage.io.zip file created by CAREamics. You should implement the corresponding `elif` in `create_model`.

src/careamics_restoration/bioimage/io.py
mese79: I should remove the `CAREAMICS_CONFIG`. It was for the previous way of exporting.

mese79:
> currently it doesn't seem that one can start the engine from a path pointing to a bioimage.io.zip file. 

There is a method `import_bioimage_model` in `./bioimage/io.py` that one can use to import a zip model to get a path to the checkpoint file save inside the zip model. From there an `Engine` can be created.

src/careamics_restoration/bioimage/rdfs/n2v.yaml
mese79: No, that's because the `postprocessing` and the `preprocessing` must be a list of actual processing, so we can have more than one pre/post processings.

src/careamics_restoration/engine.py
mese79: Yep, agree. I just couldn't find anything else to point to the "N2V".

jdeschamps: Hard-code it for now, later on we will have criterion based on the Engine to decide the algorithm.

For instance N2V2 has the N2V loss but a different model architecture.

src/careamics_restoration/bioimage/io.py
```diff
@@ -0,0 +1,102 @@
...
+    ------
+        A bioimage raw Model
+    """
+    workdir = config.working_directory
+    # attach the best checkpoint to the bioimage model
+    checkpoint_path = workdir.joinpath(f"{config.experiment_name}_best.pth").absolute()
+    # build model zip
+    raw_model = build_model(
+        root=str(Path(model_specs["output_path"]).parent.absolute()),
+        attachments={"files": [str(checkpoint_path)]},
```
jdeschamps, line 68: This means that we save twice the weights or model state_dict. I don't have a good notion or how heavy they can be, but I am uneasy with this idea in general.

It would be better to save:
- configuration
- optimizer state_dict
- scheduler state_dict
- grad_scaler state_dict

as additional files, rather than the whole checkpoint.

src/careamics_restoration/bioimage/io.py
mese79: The current `n2v.yaml` inside the `rdfs` folder contains almost all required fields. The `documentation` field should point to a `.md` file that I put it in the `docs` folder (so later we can have separate docs for different models). Also for the `covers` there is a folder.  
So these fields added to the default model's rdf dict, after loading the related yaml file.  
We can hard-code all of these somewhere, but I think having a long dict of nested dicts/lists also is not that clean...but if it is just for now and for "n2v", it could be an option.

mese79: All the info inside the `n2v.yaml` file are required, plus the `documentation` that after loading the `yaml` file gets added to the model's rdf dict.  
We can hard-code them all, but having a long dict of nested dicts/lists is not very clean too, and I think it is not actual functioning code but some meta-data filling code.  
Anyway, It can be done that way, specially now, that we only have one model.

src/careamics_restoration/bioimage/docs/n2v.md
mese79: `n2v.md` is for the `documentation` field

src/careamics_restoration/engine.py
mese79: I see, ok.

src/careamics_restoration/bioimage/docs/n2v.md
jdeschamps: Hmm yes the build_model requires paths... In that case you are right, let's keep the md in here.

jdeschamps:
> > currently it doesn't seem that one can start the engine from a path pointing to a bioimage.io.zip file.
> 
> There is a method `import_bioimage_model` in `./bioimage/io.py` that one can use to import a zip model to get a path to the checkpoint file save inside the zip model. From there an `Engine` can be created.

Somehow I thought I answered that, but I can find my previous message... Here I go again:

Importing the model to then create the `Engine` is a bit cumbersome. It would be better to just have to do `engine = Engine(model_path="path/to/bioimage.io.zip")`.

For that you can just add an `elif` clause in `create_model` that calls your import function.

src/careamics_restoration/engine.py
mese79: In the careamics config, the axes could be `STCZYX`, and I assume that we don't have the "batch" axes there, right?  
Because data for a conv net sould be at least 4D, and if I replace the "S" for "b" we get "byx" which is just 3D.

jdeschamps: @CatEek 

The channel dimension should be there as well (bcyx). That might be a problem because we do "pre-processing" on the axes and add the S and C dimensions if necessary. But the bioimage.io core might not.

Although I am pretty sure this was also the case in the napari plugin. So maybe that is all good. Maybe look at how things are done in the napari plugin because it should be quite similar.

tests/bioimage/test_io.py
mese79: I handled it this way:
```python
    rdf = load_resource_description(zip_file)
    assert isinstance(rdf, nodes.ResourceDescription)
```

mese79:
> For that you can just add an `elif` clause in `create_model` that calls your import function.

I edited `create_model` this way:

```python
if not model_path.exists or model_path.suffix not in [".pth", ".zip"]:
            raise ValueError(f"Invalid model path: {model_path}")

        if model_path.suffix == ".zip":
            model_path = import_bioimage_model(model_path)

```

[create_model](https://github.com/CAREamics/careamics-restoration/blob/0521ffaf3e0e519a0b18ae39db803d3528c5e2fd/src/careamics_restoration/models/model_factory.py#L68)

src/careamics_restoration/engine.py
```diff
@@ -532,12 +532,21 @@ def save_as_bioimage(
         if model_specs is not None:
             specs.update(model_specs)
 
+        # set in/out axes from config
```
jdeschamps, line 535: mean/std also need to be updated. I couldn't find it here.

jdeschamps (changes requested):

src/careamics_restoration/engine.py
mese79: Oh, that I forgot. The mean/std in pre/post processing should be replaced by mean/std from the config?

jdeschamps: Yes :)

jdeschamps: Just like in https://github.com/juglab/napari-n2v/blob/40317fefbb41e0b44df354440c61616b8d13bbf1/src/napari_n2v/utils/io_utils.py#L245

mese79: Done!
To pass the test, I let the mean & std have some values in default specs, then if we have a trained model `config.data.mean` should have some value and so the replacement happens.

PR #63, opened 2023-07-27
Cleanup

CatEek:
Fixed (almost) all mypy and all ruff

The only 4 mypy errors left are
```
src/careamics_restoration/prediction_utils.py:29: error: Argument 1 to <tuple> has incompatible type "ellipsis"; expected "ndarray[Any, dtype[integer[Any]]] | ndarray[Any, dtype[bool_]]"  [arg-type]
src/careamics_restoration/prediction_utils.py:30: error: Argument 2 to <tuple> has incompatible type "*list[slice]"; expected "ndarray[Any, dtype[integer[Any]]] | ndarray[Any, dtype[bool_]]"  [arg-type]
src/careamics_restoration/config/config.py:224: error: Argument "exclude_optionals" to "model_dump" of "Data" has incompatible type "bool"; expected "dict[Any, Any]"  [arg-type]
src/careamics_restoration/config/config.py:233: error: Argument "exclude_optionals" to "model_dump" of "Prediction" has incompatible type "bool"; expected "dict[Any, Any]"  [arg-type]
```

Ones in the config I have no idea what's wrong.
Ones in the prediction_utils probably have something to do with python versions. Because syntax itself should be correct


src/careamics_restoration/engine.py
```diff
@@ -91,14 +91,18 @@ def __init__(
...
+        # Create model, optimizer, lr scheduler and gradient scaler and load everything
+        # to the specified device
         (
             self.model,
             self.optimizer,
             self.lr_scheduler,
             self.scaler,
             self.cfg,
-        ) = create_model(config=self.cfg, model_path=model_path)
+        ) = create_model(config=self.cfg, model_path=model_path, device=self.device)
```
CatEek, line 105: Also found a bug, no Idea how smoke test was running before for case with loading state_dicts

veegalinova: Good point, there is no smoke test checking the model with state_dict, I'll add it in a separate PR

CatEek: It's there actually. But take a look

veegalinova (approved):

PR #67, opened 2023-08-08
Global RNG generator

veegalinova:
Create a single numpy `default_rng` instance with a fixed random seed to import in all modules for added reproducibility.


veegalinova:
@CatEek Can you please check if your target metrics are reproducing after this patch?

CatEek:
Checked. Unfortunately, this thing breaks something again. Maybe we can think in more detail later

PR #76, opened 2023-09-04
Prediction refactoring

CatEek:
1. Removed rich logger, added custom keras style
2. Predict refac

Haven't done any performance optimization :( 
Probable need to add tests for new logger


src/careamics_restoration/engine.py
```diff
@@ -201,48 +198,68 @@ def train(
...
-                overall_progress=True,
-            ):  # loop over the dataset multiple times
-                train_outputs = self._train_single_epoch(
+            train_stats = []
+            eval_stats = []
+
+            # loop over the dataset multiple times
+            for epoch in range(self.cfg.training.num_epochs):
+                try:
+                    epoch_size = epoch_size
```
jdeschamps, line 209: `epoch_size` is defined nowhere

tests/utils/test_logging.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.mark.parametrize(
+    "max_value, epoch, num_epochs, mode",
+    [(10, 0, 0, "predict"), (None, 0, 0, "predict")],
+)
+def test_progress_bar_update(max_value, epoch, num_epochs, mode):
+    progress_bar = ProgressBar(
+        max_value=max_value, epoch=epoch, num_epochs=num_epochs, mode=mode
+    )
+    for step in range(2):
+        progress_bar.update(step)
```
jdeschamps: `update()` is missing `batch_size` argument? That breaks the CI

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     'zarr',
     'rich',
     'rich-pixels',
+    'ipython'
```
jdeschamps: I would not put ipython as dependency, since it is not necessary to run the library.

We already have put jupyter in the optional [notebook] and [all] dependencies (pip install "careamics[all]") for people wanting to use the notebooks.

You could replace jupyter by ipython in those optional dependencies if you agree!

jdeschamps: Ah I see that it's actually used in the code for the wandb logger? Then rather in the [wandb] and [all] optional groups.

So that we can install it using "pip install "careamics[wandb]"

src/careamics_restoration/engine.py
jdeschamps: Was this solved?

CatEek: That's why it's wrapper into try except. Logic is the 1st epoch progress bar has no length, starting from the 2nd it will have the length

Issue #80, opened 2023-09-14
BioImage.io export [BUG]


CatEek:
Method `save_as_bioimage` of Engine produces this for certain output shapes

```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[9], line 1
----> 1 engine.save_as_bioimage(engine.cfg.experiment_name + "bioimage.zip")

File [~/projects/caremics/src/careamics_restoration/engine.py:762](https://vscode-remote+ssh-002dremote-002bvdi.vscode-resource.vscode-cdn.net/home/igor.zubarev/projects/caremics/examples/3D/~/projects/caremics/src/careamics_restoration/engine.py:762), in Engine.save_as_bioimage(self, output_zip, model_specs)
    759     specs = self._generate_rdf(model_specs)
    761     # Build model
--> 762     raw_model = build_zip_model(
    763         path=output_zip,
    764         config=self.cfg,
    765         model_specs=specs,
    766     )
    768     return raw_model
    769 else:

File [~/projects/caremics/src/careamics_restoration/bioimage/io.py:166](https://vscode-remote+ssh-002dremote-002bvdi.vscode-resource.vscode-cdn.net/home/igor.zubarev/projects/caremics/examples/3D/~/projects/caremics/src/careamics_restoration/bioimage/io.py:166), in build_zip_model(path, config, model_specs)
    163     model_specs["tags"].append("2D")
    165 # build model zip
--> 166 raw_model = build_model(
    167     output_path=Path(path).absolute(),
    168     **model_specs,
    169 )
    171 # remove the temporary files
    172 weight_path.unlink()

[...]

PR #81, opened 2023-09-23
fixed unmatched axes with input shape for bioimage.io export

mese79:
A fix for #80 .  
First I get the `axes` in config, after adding `channel` and `batch` to it, the input sample will be made matching the `axes`.  
But still these sample input/output are dummy and not reflecting the real data.  
Another caveat is that for this dummy input **channel** is assumed **1**. I couldn't find anywhere to get the real input channel(s).


jdeschamps:
Branches heavily diverged, so modifications added to #85

PR #83, opened 2023-09-26
Rename to careamics

jdeschamps:
- Rename package to "careamics"
- Update all imports, tests and notebooks
- Include the poster branch with the in memory dataset
- Fix tests and mypy

:warning: After that, you will need to reinstall the package in your environment.


jdeschamps:
I feel there are some oversight in the dataset testing, since there are a lot of optional parameters there is room for errors (which were obvious when fixing the mypy errors).

jdeschamps:
There are a few ruff errors in the notebooks now. Could you have a look?

[edit]: I am going to push the fix

PR #85, opened 2023-10-04
Fix docs

jdeschamps:
- [x] Add numpydocs pre-commit to check docstrings
- [x] Make seeding optional in Engine
- [x] A bit of refactoring 
- [x] Fix all docs
- [x] Include #81 
- [x] Remove seeding


jdeschamps:
@CatEek Can you check the seeding in `Engine.__init__` (line 227)? I have made it optional, but since I don't understand the cudnn deterministic mode, I left a todo because the flow seems illogical to me.

PR #115, opened 2024-05-06
Mc/feat/u net channel ind

melisande-c:
### Add independent channel training

- Add class to adapt models for independent channel training `IndChannelModel`
    - Creates a separate model for each channel and stores within a list
    - During forward pass the input is split by channel and passed to each network
- Add option in `UNetModel` configuration class to toggle independent channels, default is set to True.
- Add notebook in examples to compare results from training with the channels combined or independent


src/careamics/models/unet.py
```diff
@@ -233,6 +234,118 @@ def forward(self, *features: List[torch.Tensor]) -> torch.Tensor:
...
+            "num_classes": num_classes,
+            "in_channels": in_channels,
+            "depth": depth,
+            "num_channels_init": num_channels_init,
+            "use_batch_norm": use_batch_norm,
+            "dropout": dropout,
+            "pool_kernel": pool_kernel,
+            "final_activation": final_activation,
+        }
+        # TODO: This could be part of the `model_factory` function
```
jdeschamps, line 321: Agreed, that would probably make more sense to have it refactored.

The "`independent_channel` could be a parameter of the `AlgorithmConfig` (it changed name recently) which would be passed on to `model_factory` along the model config (https://github.com/CAREamics/careamics/blob/main/src/careamics/lightning_module.py#L59).

That way we can also pass the name of the in/out channel according to the model to the `ind_channel_model`.

jdeschamps:
Awesome!!!

- Let's put the notebooks in https://github.com/CAREamics/careamics-examples
- I agree with your todo (see comment) 😄 
- Some tests are not passing.
- The ruff pre-commit is not passing, some really small stuff...
- We are slowly converging on the mypy errors, currently I ignore the datasets and patching, but maybe check the files that you modified!
- The BMZ export will break, we would need an `if else` for the situation where we have independent channels: https://github.com/CAREamics/careamics/blob/f09dd94955b49984ce237d7de302e9c79a7c89c3/src/careamics/model_io/bioimage/model_description.py#L247
 Both the `callable` and the `kwargs` need to be updated!

melisande-c:
> Awesome!!!
> 
> * Let's put the notebooks in https://github.com/CAREamics/careamics-examples
> * I agree with your todo (see comment) 😄

Some refactoring is definitely necessary, it is a little ugly at the moment. I wanted to check the concept of this pattern was ok to add to careamics.

> * Some tests are not passing.
> * The ruff pre-commit is not passing, some really small stuff...
> * We are slowly converging on the mypy errors, currently I ignore the datasets and patching, but maybe check the files that you modified!

Yes I saw some tests were not passing, but some were not passing before my changes I think? I will go through them. I also need to add some tests for the new class.

> * The BMZ export will break, we would need an `if else` for the situation where we have independent channels: https://github.com/CAREamics/careamics/blob/f09dd94955b49984ce237d7de302e9c79a7c89c3/src/careamics/model_io/bioimage/model_description.py#L247
>   

Yes I was anticipating that changing the class might break some things. Personally I would try to avoid if else statements if possible; however, I can't think of a neater solution at this time.

>   Both the `callable` and the `kwargs` need to be updated!

src/careamics/models/unet.py
melisande-c: I was thinking this too. At the moment it's in the model because it evolved from adapting the UNet architecture. I can start refactoring.

melisande-c:
Close because alternative independent channels implementation using grouped convolutions is neater

melisande-c:
Close because alternative independent channels implementation using grouped convolutions is neater

PR #123, opened 2024-05-27
Fix some pre-commit errors

jdeschamps:
### Description

Fix ruff and some mypy errors.

Notes:
- `NDFlip` has been renamed `XYFlip` since it is not flipping z since an earlier PR

---

**Please ensure your PR meets the following requirements:**

- [x] Code builds and passes tests locally, including doctests
- [x] New tests have been added (for bug fixes/features)
- [ ] Pre-commit passes
- [x] PR to the documentation exists (for bug fixes / features)

PR #124, opened 2024-05-28
Remove prediction transform

jdeschamps:
### Description

It is not clear whether there is a use-case for prediction transforms. Currently, only normalization is allowed and instantiated. This PR simply removes the possibility to set your own transforms during prediction, and enforces `Normalize` as the sole prediction transform.

In addition, this PR removes the inheritance between the `PathIterableDataset` and the `IterablePredictionDataset`, as the latter being a child of the former lead to the transforms being instantiated multiple times, and the `mean` and `std` to be calculated over the prediction dataset and updated into the `InferenceModel` (while the inference model was talking it from the `Configuration`...).

This simplified the code in the `InferenceModel` as well as in the datasets.

Note that this is independent of the tta transforms.

- **What**: Remove the prediction transforms and enforce normalization alone.
- **Why**: The feature was unused and with no clear use-case, leading to unnecessarily complex code.
- **How**: The `transforms` were removed throughout the prediction functions, `NormalizeModel` was applied as `patch_transform`.

### Changes Made

- **Modified**: 
    - `InferenceModel` does not have the `transforms` parameter anymore.
    - `CAREamist.predict` and `PredictDataWrapper` do not have the `transforms` parameter anymore.
    - `_iter_over_files` refactored outside of the iterable datasets
    - `IterablePredictionDataset` is no longer a child of `PathIterableDataset`

### Breaking changes

Any code with the `transforms` keyword for prediction will break. Any code relying on the prediction statistics coming from the prediction images will not longer have the same normalization.

This should fix the last mypy error in https://github.com/CAREamics/careamics/pull/123.

---

- [ ] Pre-commit passes


melisande-c (approved):
Makes sense to me! 

In future if any more functionality needs to be shared between `PathIterableDataset` and `IterablePredictionDataset`, other than the function `_iterate_over_files`, then we could implement a mixin class. However right now that would be unnecessarily complex, and just converting the method to a function as you have done makes sense!

jdeschamps:
> Makes sense to me!
> 
> In future if any more functionality needs to be shared between `PathIterableDataset` and `IterablePredictionDataset`, other than the function `_iterate_over_files`, then we could implement a mixin class. However right now that would be unnecessarily complex, and just converting the method to a function as you have done makes sense!

that's a great idea!

PR #128, opened 2024-06-03
Fix transforms

jdeschamps:
### Description

I erroneously removed the probability of applying transforms from the flips and rotations, leading to a situation where the original patch is never shown to the network. This PR adds the probability parameter back into the transforms. It also adds a new transform that only flips a single axis (as opposed to either X or Y), as this can be useful for particular applications (e.g. cryoCARE).

- **What**: Add `p` parameter to transforms, create new `XorYFlip` transform with parameter `flip_x` (`True` for X, `False` for Y).
- **Why**: This fixes a problem arising from the transforms always applying, a preventing the original patch to be trained on.
- **How**: Parameter is simply added to the existing classes, new class for the `XorYFlip` transform.

### Changes Made

- **Added**: `XorYFlip` and `XorYFlipModel`.
- **Modified**: `SupportedTransforms`, `XYFlip`, `XYFlipModel`, `XYRandomRotate90`, `XYRandomRotate90Model`, tests.

Another way to create the `XorYFlip` transform would have been to add parameters to the existing `XYFlip` to make the flip happen on a single axis. I thought that a `flip_x` (`True` for X, `False` for Y) parameter in a new class is easier for users than asking for an axis value `Optional[Literal[-2, -1]]` to represent the axis to flip in `XYFlip`.

Current transform list for flipping only X or Y using `XorYFlip`:

```python 
transforms =  [
    NormalizeModel(mean=mean, std=std),
    XorYFlipModel(flip_x=True),
    N2VManipulateModel(),
]
```

Potential alternative if implementing it in `XYFlip`:

```python 
transforms =  [
    NormalizeModel(mean=mean, std=std),
    XYFlipModel(axis=-1), # parameters `axis=None` by default to flip both X and Y
    N2VManipulateModel(),
]
```

Or:

```python 
transforms =  [
    NormalizeModel(mean=mean, std=std),
    XYFlipModel(flip_one_axis=True, flip_x=True), # parameters `flip_one_axis=False` by default to flip both X and Y
    N2VManipulateModel(),
]
```
Any opinion?

---

- [x] Pre-commit passes
- [x] PR to the documentation exists (for bug fixes / features): 
     - https://github.com/CAREamics/careamics.github.io/pull/6
     - https://github.com/CAREamics/careamics-examples/pull/1


melisande-c:
Regarding the X or Y flip transform, I prefer taking the functionality into the existing `XYFlip` class, using the second option you have presented to do this. I think it is perfectly reasonable in the scope of the `XYFlip` class to be able to specify which axis to flip, and it reduces code duplication.

Also small naming note, it's technically an exclusive-or operation so I would call it `XxorYFlip` if it is kept.

jdeschamps:
> Also small naming note, it's technically an exclusive-or operation so I would call it `XxorYFlip` if it is kept.

Ah good point, but now I am tempted to keep the class just for the geeky name!

Let me try to one class solution then!

jdeschamps:
> Regarding the X or Y flip transform, I prefer taking the functionality into the existing `XYFlip` class, using the second option you have presented to do this. I think it is perfectly reasonable in the scope of the `XYFlip` class to be able to specify which axis to flip, and it reduces code duplication.

Here is a new proposal:

``` python
# Flip only Y
XYFlip(
    flip_x=False,
)

# Flip only X
XYFlip(
    flip_y=False,
)

# raises error
XYFlip(
    flip_x=False,
    flip_y=False
)
```

melisande-c:
> 
> Here is a new proposal:
> 
> ```python
> # Flip only Y
> XYFlip(
>     flip_x=False,
> )
> 
> # Flip only X
> XYFlip(
>     flip_y=False,
> )
> 
> # raises error
> XYFlip(
>     flip_x=False,
>     flip_y=False
> )
> ```

I had this thought too, I think it's clear for users and has a simple implementation!

melisande-c (approved):

PR #134, opened 2024-06-06
Channel-dependent normalization

CatEek:
### Description

Input patches and targets should be normalized separately, because they may have very different pixel value ranges. Also normalization was done for all channels at once, which might not be correct for cases where 2 channels should not be mixed.

- **What**: Added separate normalization for images and targets and per channel normalization
- **Why**: Normalizing everything with the same stats is not correct 

### Changes Made

- Normalization func + some changes in the datasets, training and prediction loops.
- Reelevant tests

Please ensure your PR meets the following requirements:

 

- [x]  New tests have been added (for bug fixes/features)
- [x]  Pre-commit passes
- [ ]  No change to the documentation needed


src/careamics/config/data_model.py
```diff
@@ -235,11 +239,41 @@ def std_only_with_mean(self: Self) -> Self:
             If std is not None and mean is None.
         """
         # check that mean and std are either both None, or both specified
-        if (self.mean is None) != (self.std is None):
+        if (self.image_mean and not self.image_std) or (
```
jdeschamps, line 235: Why not keep `if (self.image_mean is None) != (self.image_std is None):`? I find this a lot more readable.

src/careamics/config/data_model.py
```diff
@@ -90,8 +92,10 @@ class DataConfig(BaseModel):
     axes: str
 
     # Optional fields
-    mean: Optional[float] = None
-    std: Optional[float] = None
+    image_mean: Optional[List[float]] = Field(default=[], min_length=0, max_length=32)
```
jdeschamps: It should probably be `image_means`, plural. Same for the others!

src/careamics/config/data_model.py
```diff
@@ -235,11 +239,41 @@ def std_only_with_mean(self: Self) -> Self:
...
+                raise ValueError(
+                    "Mean and std must be either both None, or both specified for each "
+                    "target channel."
+                )
+
+            if len(self.target_mean) > 0 and len(self.image_mean) != len(
+                self.target_mean
+            ):
+                raise ValueError(
+                    "Statistics of input and target channels must be the same."
```
jdeschamps: Why? What if we are training 3 channels input to 2 channels target? Or single channel to two channels?

There are examples in CARE, and I guess muSplit.

src/careamics/dataset/dataset_utils/dataset_utils.py
```diff
@@ -99,3 +99,22 @@ def reshape_array(x: np.ndarray, axes: str) -> np.ndarray:
         _x = np.expand_dims(_x, new_axes.index("S") + 1)
 
     return _x
+
+
+def compute_normalization_stats(image: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    """
+    Compute mean and standard deviation of an array.
```
jdeschamps, line 106: not only mean and std, but for each channel! 

The main docstring needs more details

src/careamics/dataset/iterable_dataset.py
```diff
@@ -145,15 +146,47 @@ def __init__(
         self.read_source_func = read_source_func
 
         # compute mean and std over the dataset
-        if not data_config.mean or not data_config.std:
-            self.mean, self.std = self._calculate_mean_and_std()
+        # Only checking the image_mean because the DatasetConfig class ensures that
```
jdeschamps: `DataConfig` rather than `DatasetConfig` in the comment

src/careamics/dataset/iterable_dataset.py
```diff
@@ -145,15 +146,47 @@ def __init__(
...
+        # if image_mean is provided, image_std is also provided
+        if not self.data_config.image_mean:
```
jdeschamps, line 76: Are you testing that image mean is `None` or an empty list here? Because both are possible given the type declaration in the Pydantic model.

What if users pass an empty list?

src/careamics/dataset/iterable_dataset.py
```diff
@@ -145,15 +146,47 @@ def __init__(
...
-            self.mean = data_config.mean
-            self.std = data_config.std
+            self.patches_data = PatchedOutput(
+                None,
+                None,
+                self.data_config.image_mean,
+                self.data_config.image_std,
+                (
+                    self.data_config.target_mean
+                    if hasattr(self.data_config, "target_mean")
```
jdeschamps: Can we use a `is_supervised` boolean created in the `__init__` rather than `hasattr`. That would make the whole code a lot easier to read.

src/careamics/dataset/iterable_dataset.py
```diff
@@ -145,15 +146,47 @@ def __init__(
...
+                    else None
+                ),
+                (
+                    self.data_config.target_std
+                    if hasattr(self.data_config, "target_std")
+                    else None
+                ),
+            )
+        if hasattr(self.data_config, "set_mean_and_std"):
```
jdeschamps: Why the `hasattr` here? A `data_config` always has `set_mean_and_std`.

src/careamics/dataset/iterable_dataset.py
```diff
@@ -145,15 +146,47 @@ def __init__(
...
+            self.data_config.set_mean_and_std(
+                image_mean=list(self.patches_data.image_means),
+                image_std=list(self.patches_data.image_stds),
+                target_mean=(
+                    list(self.patches_data.target_means)
+                    if self.patches_data.target_means is not None
```
jdeschamps: `target_means` has a default value `[]` in `DataConfig`, why would it be None here?

src/careamics/config/data_model.py
```diff
@@ -90,8 +92,10 @@ class DataConfig(BaseModel):
     axes: str
 
     # Optional fields
-    mean: Optional[float] = None
-    std: Optional[float] = None
+    image_mean: Optional[List[float]] = Field(default=[], min_length=0, max_length=32)
+    image_std: Optional[List[float]] = Field(default=[], min_length=0, max_length=32)
+    target_mean: Optional[List[float]] = Field(default=[], min_length=0, max_length=32)
```
jdeschamps: Why declare the lists as `Optional` but give them a default `[]` value?

That means that we always need to specifically set them to `None` at the beginning. How do we differentiate between an voluntary empty list and a default value?

src/careamics/dataset/iterable_dataset.py
```diff
@@ -145,15 +146,47 @@ def __init__(
...
+                f"std: {self.patches_data.image_stds}"
+            )
 
-            # update mean and std in configuration
-            # the object is mutable and should then be recorded in the CAREamist
-            data_config.set_mean_and_std(self.mean, self.std)
         else:
-            self.mean = data_config.mean
-            self.std = data_config.std
+            self.patches_data = PatchedOutput(
```
jdeschamps: I don't see the point of using a Dataclass just for half of its entries.

Rather make a second Dataclass (statistics) and make the PatchedOutput inherit from it.

```python
from dataclasses import dataclass
from typing import List 
import numpy as np

@dataclass
class Statistics:
    means: List[float]
    stds: List[float]

@dataclass
class PatchedImage(Statistics):
    array: np.ndarray


PatchedImage(means=[1, 2, 3], stds=[4, 5, 6], array=np.array([1, 2, 3]))
```

And return a PatchedImage for the source and targets images.

Other alternative:

```python
from dataclasses import dataclass
from typing import List 
import numpy as np

@dataclass
class Statistics:
    means: List[float]
    stds: List[float]

@dataclass
class PatchedImage:
    array: np.ndarray
    statistics: Statistics

PatchedImage(array=np.array([1, 2, 3]), statistics=Statistics(means=[1, 2, 3], stds=[4, 5, 6]))
```

src/careamics/dataset/iterable_dataset.py
```diff
@@ -167,25 +200,49 @@ def _calculate_mean_and_std(self) -> Tuple[float, float]:
...
+            image_stds[ch] /= num_samples
+            if target is not None:
+                target_means[ch] /= num_samples
+                target_stds[ch] /= num_samples
 
         logger.info(f"Calculated mean and std for {num_samples} images")
-        logger.info(f"Mean: {result_mean}, std: {result_std}")
-        return result_mean, result_std
+        logger.info(f"Mean: {image_means}, std: {image_stds}")
+        return PatchedOutput(
```
jdeschamps: Again here, we could just return `-> Tuple[Statistics, Optional[Statistics]]`

src/careamics/transforms/normalize.py
```diff
@@ -67,26 +85,41 @@ def __call__(
...
             Transformed patch and target.
         """
-        norm_patch = self._apply(patch)
-        norm_target = self._apply(target) if target is not None else None
+        norm_patch = np.zeros_like(patch, dtype=np.float32)
+        norm_target = (
+            np.zeros_like(target, dtype=np.float32) if target is not None else None
+        )
+
+        for i in range(patch.shape[0]):
```
jdeschamps: It would be better to use broadcasting to do the normalization:

```python
array = np.arange(3 * 4 * 4).reshape(3, 4, 4)

means = np.mean(array, axis=(1, 2))[:, np.newaxis, np.newaxis]
stds = np.std(array, axis=(1, 2))[:, np.newaxis, np.newaxis]

norm_array = (array - means) / stds
norm_array.mean(axis=(1, 2)), norm_array.std(axis=(1, 2))
```

jdeschamps: We just need to adjust the number of `np.newaxis` depending on whether there is a `Z` dim. We know this from the length of the shape.

src/careamics/transforms/normalize.py
```diff
@@ -150,12 +182,18 @@ def __call__(
...
             Transformed patch and target.
         """
-        norm_patch = self._apply(patch)
-        norm_target = self._apply(target) if target is not None else None
+        norm_array = np.zeros_like(patch, dtype=np.float32)
 
-        return norm_patch, norm_target
+        # Iterating over the batch dimension
+        for i in range(patch.shape[0]):
+            for ch in range(patch.shape[1]):
```
jdeschamps: Same here, use broadcasting for the channels

tests/transforms/test_compose.py
```diff
@@ -59,18 +61,20 @@ def test_compose_n2v(ordered_array):
...
     # transform lists
-    mean, std = 0.5, 0.5
+    means, stds = compute_normalization_stats(image=array)
+
 
     transform_list_pydantic = [
-        NormalizeModel(mean=mean, std=std),
-        XYFlipModel(seed=seed),
+        NormalizeModel(image_means=means, image_stds=stds),
+        NDFlipModel(seed=seed),
```
jdeschamps: `XYFlipModel` is the new name, I must have missed that in the merge 😞 

This must trigger a failure in the tests!

jdeschamps (changes requested):
I did the merge with the most recent main, and that introduced some errors. So I pushed some fixes now, but not all tests pass.

Mypy is definitely not passing...

The main points in the review are the following:
- dataclasses use
- improving the normalization with broadcasting
- type declaration for the stats in the Pydantic model

jdeschamps:
There is still one test no passing related to the BMZ, potentially because it was before using a tensor [-1, 1] with meant that the normalization was not mattering so much for the result.

Now we are again with small differences between CAREamics prediction and BMZ one. Let's understand where it comes from.

src/careamics/config/data_model.py
CatEek: Looks like a merge artifact

src/careamics/dataset/iterable_dataset.py
CatEek: I'd honestly rather use hasattr. Also maybe i'm missing something, what is_supervised are you referring to ?

src/careamics/dataset/iterable_dataset.py
CatEek: That was because of the inheritance from training ds

src/careamics/dataset/iterable_dataset.py
CatEek: `bool([]) == False`

src/careamics/dataset/iterable_dataset.py
CatEek: 1st option might create inconsistencies in unpacking the func output later. 2nd one is good, will change

src/careamics/config/data_model.py
CatEek: Looks like a merge artifact

src/careamics/config/data_model.py
jdeschamps: Nope

https://github.com/CAREamics/careamics/blob/15ccfb8ca9b4dfb311c169520a3059d10f863c12/src/careamics/config/data_model.py#L93

src/careamics/config/data_model.py
jdeschamps: Nope

https://github.com/CAREamics/careamics/blob/15ccfb8ca9b4dfb311c169520a3059d10f863c12/src/careamics/config/data_model.py#L240


But it could be an automated tool running in VSCode (black?).

src/careamics/config/data_model.py
CatEek: Good point, fixed

src/careamics/dataset/dataset_utils/dataset_utils.py
CatEek: Added

src/careamics/dataset/iterable_dataset.py
CatEek: fixed

jdeschamps (approved):

PR #136, opened 2024-06-07
Feature: Add option to save predictions to disk

melisande-c:
### Description

Adding a convenient method for users to save predictions directly to disk.

- **What**: `CAREamist` now has a method `predict_to_disk`. Users can call this to predict and save the results to disk; a list of paths to where the predictions are saved will be returned.
- **Why**: Currently, when using the `CAREamist.predict` method, all predictions are saved in memory. With a large number of files this can lead to memory issues. For the CLI it is essential to have some method to save predictions. Additionally it will be convenient for users who wish to save predictions.
- **How**: 
   - A new `CAREamicsPredictionWriter` class is added to the callback list stored in `CAREamist` that is used to initialise a lightning `Trainer`.
   - `CAREamicsModule` has a toggle whether to save predictions. In the method `predict_to_disk` it is set to True at the start of the function, then set back to False at the end of the function.
   - (WIP) In `CAREamicsPredictionWriter`, if the save predictions toggle is on, the function `write_on_batch_end` will save the prediction – if the last tile has been processed and the prediction has been stitched.

### Changes Made

- **Added**: 
   - Added a `CAREamicsPredictionWriter` that inherits from the lightning class [`BasePredictionWriter`](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.BasePredictionWriter.html#lightning.pytorch.callbacks.BasePredictionWriter).
   - Added src/careamics/prediction/save_utils.py containing `get_save_func` function.
   - Added src/careamics/prediction/save_tiff.py containing `save_tiff` function.
- **Modified**: 
   - Added a `predict_to_disk` method to `CAREamist`. 
   - Extracted creation of `CAREamicsPredictData` in `CAREamist` to method `_create_pred_datamodule` to be used by both the methods `predict` and `predict_to_disk`.
   - Added `save_prediction_args` attribute to `CAREamicsModule`.
   - Added `set_save_prediction_args` method to `CAREamicsModule`.
   - Added `set_save_predictions` method to `CAREamicsModule`.
- **Removed**: 

### Related Issues

- Fixes #
- Resolves #
- Closes #

### Breaking changes

Describe any breaking change.

#### Current issues/questions

- The `CAREamicsPredictionLoop` is still caching all the predictions in the `self.predicted_array` list. This means for a large number of files there will still be memory issues. If the predictions are saved, `self.predicted_array` needs to be cleared.
- Should `predict_to_disk` be combined with `predict` with argument or kept separate?
- Spelling: "disc" (UK) or "disk" (American)
- Have a file_io/file_utils package (not in datasets package) containing `save_utils`, `read_utils` and save and read funcs.
- (minor): need to add `get_save_extension` function because the `SupportedData.get_extension` returns a pattern.

---

- [ ] Pre-commit passes


jdeschamps:
Sorry only, getting to this now... I guess the other PR takes precedence, correct?

> * Should `predict_to_disk` be combined with `predict` with argument or kept separate?
> 

My opinion is that because they have a different outcome, they should be different functions.


> * Spelling: "disc" (UK) or "disk" (American)
> 

Ahah, tough one! I usually like UK spelling better (bias from my own native language...), but on that one I am too used to "disk"... 

Because the scientific world is overwhelmingly tipped towards the US spelling, maybe we should stick to that.

> * Have a file_io/file_utils package (not in datasets package) containing `save_utils`, `read_utils` and save and read funcs.
> 

Agreed!

src/careamics/careamist.py
```diff
@@ -580,76 +652,217 @@ def predict(
...
+        """ """
+        self.model.set_save_prediction_args(
+            save_predictions=True,
+            save_type=save_type,
+            save_extension=save_extension,
+            save_predict_func=save_predict_func,
+            save_predict_func_kwargs=save_predict_func_kwargs,
+            predict_dir=predict_dir,
+            force=force,
+        )
```
melisande-c, line 702: @jdeschamps what do you think about this? I find it kind of annoying. Can you think of a better solution?

melisande-c:
> My opinion is that because they have a different outcome, they should be different functions.

Ok nice, I agree!

> Ahah, tough one! I usually like UK spelling better (bias from my own native language...), but on that one I am too used to "disk"...
>
> Because the scientific world is overwhelmingly tipped towards the US spelling, maybe we should stick to that.

Begrudgingly, I agree

src/careamics/careamist.py
jdeschamps: So if I understand correctly:
- `CallBack` is saving data
- Saving settings live in the `LightningModule`
- Prediction call set the settings in the `LightningModule`

It's quite convoluted and I don't really like that the settings live in the module, since they really don't belong there...

## Alternative 1

Could `CAREamist` keep a reference of the `CallBack` and modify the settings directly there?

## Alternative 2

Could we make an internal `_generate_prediction` method that yields prediction by prediction (or batch of tiles per batch of tiles), and have all the stitching and saving logic happening in the `CAREamist` (or refactor somewhere else, but called there)?

Then we could have a function to run through the generator in order to save predictions, and one function to run through the generator in order to stitch and then save the predictions? (or however it makes sense to organize this).

Here we would go back into the not using Ligthning callback.

melisande-c: Yes, I agree it's convoluted and gross.

> It's quite convoluted and I don't really like that the settings live in the module, since they really don't belong there...

The reason I put the settings in the module is because the `write_on_batch_end` hook gets passed the trainer and module as arguments, so the settings can be accessed from there, see below.
https://github.com/CAREamics/careamics/blob/d90aa0e8906fb942cc35f384be83bf42d0624d84/src/careamics/lightning_prediction_writer.py#L29-L49 
However your alternative 1 would definitely solve this problem, by just having the settings saved in the callback directly!

For alternative two this means we would not really be using the lightning `trainer.predict` at all (we would but not properly). This is what the prediction loop does. It would work but would mean we would not be able to use any of the potential benefits of Lightning down the line, e.g. distributed predictions etc.

jdeschamps: Let's try with the callback then!

melisande-c:
Closing this PR because it is quite far behind main now. I will open a new one taking forward what was discussed here.

PR #137, opened 2024-06-07
Drop support for Python3.8

jdeschamps:
### Description

Drop support for Python3.8 as it will not receive security patches any longer this year.

- **What**: Drop Python 3.8.
- **Why**: Python 3.8 will not be supported any more this year.
- **How**: Adapt `pyproject.toml` and CI.

### Changes Made

- **Modified**: `pyproject.toml` and`ci.yml`.

---

- [x] Pre-commit passes


src/careamics/config/tile_information.py
```diff
@@ -17,26 +17,26 @@ class TileInformation(BaseModel):
...
     @classmethod
-    def no_singleton_dimensions(cls, v: Tuple[int, ...]):
+    def no_singleton_dimensions(cls, v: tuple[int, ...]):
         """
         Check that the array shape does not have any singleton dimensions.
 
         Parameters
         ----------
-        v : Tuple[int, ...]
+        v : tuple[int, ...]
```
melisande-c: How do you feel about:
```python
"""
v : tuple of integers
"""
```
or 
```python
"""
v : tuple of (int, ...)
"""
```
Similar to the [Pandas docstring guide](https://pandas.pydata.org/docs/development/contributing_docstring.html#parameter-types) although they don't have an example for unknown length tuples.

src/careamics/config/tile_information.py
```diff
@@ -73,23 +73,23 @@ def only_if_tiled(cls, v: bool, values: ValidationInfo):
...
+    ) -> Optional[tuple[int, ...]]:
         """
         Check that the coordinates are not `None` if tiling is enabled.
 
         The method also return `None` if tiling is not enabled.
 
         Parameters
         ----------
-        v : Optional[Tuple[int, ...]]
+        v : tuple[int, ...] or None
```
melisande-c: I think
```python
v : tuple of integers, optional
```

(This is mostly a comment on `optional` vs `or None`, see [other comment](https://github.com/CAREamics/careamics/pull/137/files#r1633192490) for tuple style.)

src/careamics/config/configuration_model.py
```diff
@@ -279,7 +279,7 @@ def set_3D(self, is_3D: bool, axes: str, patch_size: List[int]) -> None:
             Whether the algorithm is 3D or not.
         axes : str
             Axes of the data.
-        patch_size : List[int]
+        patch_size : list[int]
```
melisande-c, line 282: Could be
```python
"""
patch_size : list of int
"""
```

src/careamics/dataset/iterable_dataset.py
```diff
@@ -91,38 +91,36 @@ class PathIterableDataset(IterableDataset):
...
-    mean : Optional[float], optional
+    mean : float or None, optional
         Expected mean of the dataset, by default None.
-    std : Optional[float], optional
+    std : float or None, optional
         Expected standard deviation of the dataset, by default None.
-    patch_transform : Optional[Callable], optional
+    patch_transform : Callable or None, optional
         Patch transform callable, by default None.
     """
```
melisande-c, line 53: Again, I don't think the `"or None"` is necessary for optional params.

src/careamics/dataset/in_memory_dataset.py
```diff
@@ -47,8 +47,8 @@ class InMemoryDataset(Dataset):
     def __init__(
         self,
         data_config: DataConfig,
-        inputs: Union[np.ndarray, List[Path]],
-        input_target: Optional[Union[np.ndarray, List[Path]]] = None,
+        inputs: Union[np.ndarray, list[Path]],
+        input_target: Optional[Union[np.ndarray, list[Path]]] = None,
```
melisande-c, line 49: Maybe should be `numpy.array` not `np.array`.

Also the [`numpy.typing`](https://numpy.org/devdocs/reference/typing.html#) package had the [`NDArray`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray) and [`ArrayLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) types for type hinting.

src/careamics/config/configuration_model.py
```diff
@@ -524,7 +524,7 @@ def load_configuration(path: Union[str, Path]) -> Configuration:
 
     Parameters
     ----------
-    path : Union[str, Path]
+    path : str or Path
```
melisande-c, line 527: Maybe python package of `Path` should be explicit i.e.
```python
"""
path : str or pathlib.Path
"""
```

src/careamics/config/data_model.py
```diff
@@ -180,19 +178,19 @@ def axes_valid(cls, axes: str) -> str:
...
-    ) -> List[TRANSFORMS_UNION]:
+        cls, transforms: list[TRANSFORMS_UNION]
+    ) -> list[TRANSFORMS_UNION]:
         """
         Validate N2VManipulate transform position in the transform list.
 
         Parameters
         ----------
-        transforms : List[Transformations_Union]
+        transforms : list[Transformations_Union]
```
melisande-c, line 188: can this be
```python
"""
transforms : list of careamics.config.transformations.TransformModel
"""
```
it's a bit long though 😅. So alternative could be something like
```python
"""
transforms : list of careamics TransformModel
    (see careamics.config.transformations.TransformModel)
"""
```

I know this isn't really a function users will be using but type aliases in docstrings should be clearer

src/careamics/config/tile_information.py
jdeschamps: I like the first one, makes it quite readable!

src/careamics/dataset/iterable_dataset.py
melisande-c: Also I think the "by default None", can be removed – it is implied by "optional".

melisande-c (approved):
Anyway, we can change docstrings as we go along! Just thought I would add some comments here with examples

src/careamics/config/tile_information.py
jdeschamps: I am a bit torn here, because `, optional` means that you don't need to provide a value, but `Optional` means that it can be `None`. The following code is correct:

```
def my_func(a: Optional[int]) -> int:
    if a is None:
        return 0
    return a
``` 

and in there `a` is not `optional`.


Ofc, in most cases `Optional` will have a default None value... So I guess we can differentiate the two cases as follows:

```
"""
a: int or None
    A, if None then A is interpreted as 0
"""
```

vs 


```
"""
a: int, optional
    A, default value is None.
"""
```

This is not an interesting example, but there will be cases where `None` has a meaning and is not the default value. Although we should probably avoid these cases in user relevant API...

In this case, the validator should not be used directly so I would stick to the `a: int or None`, because `None` really means that we didn't pass any tile.

src/careamics/dataset/iterable_dataset.py
jdeschamps: In this case I do agree since they have a default `None`!

src/careamics/config/tile_information.py
melisande-c: I think often the standard practice is:

```python
def func(a: Optional[int]=None):
    """
    Parameters
    ----------
    a : int, optional
        Description of parameter `a`.
    """
```
vs
```python
def func(a: int=1):
    """
    Parameters
    ----------
    a: int, default=1
        Description of parameter `a`.
    """
```

src/careamics/dataset/in_memory_dataset.py
jdeschamps: Yes that would make a lot of sense!

I would tend to use `NDArray` because `ArrayLike` is a large Union that includes int and such.

src/careamics/config/data_model.py
jdeschamps: Love the second one!

src/careamics/config/tile_information.py
melisande-c: Ok looking through the function more closely maybe this is an exceptional case. But generally I would follow what I put above.

jdeschamps:
I really like the approach of making the doc more human readable. It worries me a little that we cannot automate it (that would be quite the endeavour). We could think at some point to have a small scanning tool that detects a bunch of the strings in the doc, at least this could give us some indications if not point to all errors.

On the other hand we are reaching a point where the code base will become more stable so it shouldn't be too much of an issue.

What we do need to do is to write a doccstring guidelines in https://careamics.github.io/0.1/guides/dev_resources/.

melisande-c:
> I really like the approach of making the doc more human readable. It worries me a little that we cannot automate it (that would be quite the endeavour). We could think at some point to have a small scanning tool that detects a bunch of the strings in the doc, at least this could give us some indications if not point to all errors.

Yes, it's something that's not too bad if done when writing the code, but going back and changing everything will take some work 😅. I do think it's important that docs are easy to interpret when read by new users though!

> What we do need to do is to write a doccstring guidelines in https://careamics.github.io/0.1/guides/dev_resources/.

Sounds good!

src/careamics/config/data_model.py
jdeschamps: Although we would need to check whether mkdocstring actually manage to compile the second version.

src/careamics/config/tile_information.py
jdeschamps: Works for me!

src/careamics/config/data_model.py
jdeschamps: It doesn't pass the numpydoc pre-commit unfortunately.

PR #139, opened 2024-06-07
Make normalization mandatory

jdeschamps:
### Description

The current code base applies denormalization automatically (in the prediction step of the `LightningModule` and in the BMZ export), and the datasets all compute the statistics of the data (mean and std). But `Normalization` can be disabled through the Pydantic model!

This PR makes `Normalization` mandatory by adding it to the list of transforms directly in the datasets. This simplifies the configuration (no more adding mean/std in two places) and ensures that normalization is always applied.

One way to still disable it, is to choose 0 mean and 1 std.

- **What**: Make normalization mandatory in the code.
- **Why**: Normalization is understood in the code as being always applied, yet it can be disabled through the configuration.
- **How**: `Normalization` is removed from the list of transforms and automatically added when creating the `Compose` in the datasets.

### Changes Made

- **Modified**: `DataModel`, `SupportedTransforms`, the datasets.
- **Removed**: `NormalizationModel` from `DataModel` and `SupportedTransforms`.

### Breaking changes

Any code specifically calling `Normalization` in the configuration.

---

- [x] Pre-commit passes


melisande-c (approved):
Will this make old configurations that specify normalisation break too? Maybe we could add a legacy config handler to convert old configs to match the new code, maybe unnecessary.

Otherwise looks good!

jdeschamps:
> Will this make old configurations that specify normalisation break too? Maybe we could add a legacy config handler to convert old configs to match the new code, maybe unnecessary.
> 
> Otherwise looks good!

Totally!

I would say that until we do an official v0.1.0 release, we should not worry about back compatibility of the config. Now, this will be important in the future and we will need a solution.

There is a version number in the configuration for this reason!

PR #141, opened 2024-06-12
Refactor: Tiling applied post prediction (in contrast to during loop)

### LINKED ISSUE #140: Refactoring: Decouple tiling from prediction loop
### Description
I think the tiling logic should probably be removed from the prediction loop.
#### Why
- Currently the prediction loop's `run` function is copied from the original Lightning `_PredictionLoop` class with the tiling logic added in-between the Lightning code. This is hard to maintain if Lightning ever change their code.
- Currently I'm pretty sure `CAREamicsPredictionLoop` is incompatible with `trainer.predict(*args, **kwargs, return_predictions=False)`.
- Once the save to disk function is added, when a full (stitched) prediction is saved we want to free up memory so that the prediction can be run on a large number of files. This is hard to add into the current implementation.
- When zarr datasets are added, tiles can be written into the correct place in the file without waiting for the complete set of tiles. This is hard to add into the current implementation.

#### Tiling as a Callback
Lightning already has a [`BasePredictionWriter`](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.BasePredictionWriter.html#lightning.pytorch.callbacks.BasePredictionWriter) Callback. There is a `write_on_batch_end` hook that could handle writing to zarr files or caching tiles until the last tile to save to tiff. The outputs of `trainer.predict` can also be changed so that it is the full prediction and not the tiles. I have implemented a version of this as a demo in this [branch](https://github.com/CAREamics/careamics/tree/mc/refac/move_tiling_to_callback), where I move the current tiling logic to a callback. (This would have to change a lot to accommodate writing predictions).
#### Tiling in `CAREamicsModule`
I don't like this option as much because it doesn't feel like what the `LightingModules` are for. `LightningModule`s  have all the same hooks as `Callback`, `on_predict_batch_end` etc. So all the tiling logic could move there.

### Discusion points
Please comment likes/dislikes with either solution or a new solution and any other thoughts!

- Does `trainer.predict` have to output the stitched predictions? 
   - To change the outputs of `trainer.predict`, in `on_predict_epoch_end` we have to do: 
      ```python
      trainer.predict_loop._predictions = stitched_predictions
      ```
      (see implementation in mentioned branch above) which I don't like because it feels a bit hacky.
   - predictions keep the tiling information alongside them, why not just do:
     ```python
     predictions = trainer.predict(model=model, datamodule=datamodule, ckpt_path=checkpoint)
     stitched_preds = stitched_predictions(predictions)
     ```
    - Users using the `CAREamist` class never need to see this (happens in `CAREamist.predict`). Users not using the `CAREamist` class have more control, which is why they might not be using it.

- Something to consider: 
   - Users not using `CAREamist` class, if they want tiled predictions, currently have to do:
     ```python
     trainer = Trainer(*args, **kwargs)
     trainer.prediction_loop = CAREamicsPredictionLoop(trainer)
     predictions = trainer.predict(model=model, datamodule: CAREamicsDataModule=datamodule)
     ```
   - With the Callback option:
     ```python
     trainer = Trainer(*args, **kwargs, callbacks=[TiledPredictionCallback, ...])
     predictions = trainer.predict(model=model, datamodule: CAREamicsDataModule=datamodule)
     ```
   - With the LightningModule option
     ```python
     trainer = Trainer(*args, **kwargs)
     predictions = trainer.predict(model=model, datamodule: CAREamicsDataModule=datamodule)
     ```
   - Although I have just realised this is redundant if stitching is applied afterwards (as described in point above). However they would still have to add the `PredictionWriterCallback` for saving tiles.

### LINKED ISSUE #143: Batched prediction [BUG]
**Describe the bug**
Batched prediction does not work for both tiled prediction and non-tiled prediction, for different reasons.

**To Reproduce**
***Non-tiled prediction***
```python
import numpy as np

from careamics.config import create_n2v_configuration

    experiment_name="PredBatchingTest", 
    data_type="array",
    axes="SYX",
    patch_size=[64, 64],
    batch_size=4,
    num_epochs=1,
)
images = np.random.random((4, 512, 512))

engine = CAREamist(source=config)
engine.train(train_source=images)
pred = engine.predict(source=images, batch_size=2)
```
Results in error
```console
  ...
  File "/home/melisande.croft/Documents/Repos/careamics/src/careamics/lightning_prediction_loop.py", line 108, in run
    last_tiles = [t.last_tile for t in self.tile_information]
  File "/home/melisande.croft/Documents/Repos/careamics/src/careamics/lightning_prediction_loop.py", line 108, in <listcomp>
    last_tiles = [t.last_tile for t in self.tile_information]
AttributeError: 'Tensor' object has no attribute 'last_tile'
```

This is happening because in `CAREamanics.predict_step` the batch unpacking assumes batch size is 1, see below. If `batch` has a shape (4, C, Y, X) then `x` be a tensor of shape (1, C, Y, X) and `aux` will be a tuple of length 3 containing tensors of shape (1, C, Y, X).
https://github.com/CAREamics/careamics/blob/60adc764f016eade2e6978a8fdcb4f6acb2d5658/src/careamics/lightning_module.py#L136-L151

***Tiled prediction***
```python
import numpy as np

from careamics.config import create_n2v_configuration

    experiment_name="PredBatchingTest", 
    data_type="array",
    axes="SYX",
    patch_size=[64, 64],
    batch_size=4,
    num_epochs=1,
)
images = np.random.random((4, 512, 512))

engine = CAREamist(source=config)
engine.train(train_source=images)
pred = engine.predict(source=images, batch_size=2, tile_size=(64, 64))
```
Doesn't result in an error but produces outputs:
predictions, 0 & 2:
bad_tiling
predictions, 1 & 3
output1

Cause unknown.

Please share any insight into this strange behaviour!

melisande-c:
### Description

Using the regular Lightning prediction loop, and tiled datasets, the tile information is also returned when calling `trainer.predict`. This means there is no reason tiling cannot be applied after prediction instead of during the prediction loop. This means we don't have to write a custom loop, simplifying the code and reducing coupling with Lightning. This will also make life easier when we add saving predictions (tiff or zarr).

**What**: Removes `CAREamicsPredictionLoop` (where tiling was previously implemented) and applies tiling in `CAREamist.predict` after calling `trainer.predict` with the regular Lightning prediction loop.

**Why**: Altering the Lightning prediction loop was overcomplicated, hard to maintain and made adding an option to save predictions more difficult.

**How**: predictions are returned with tiling information so predicted tiles are stitched together at the end of prediction.

### Changes Made

- **Added**:
   - multi-image stitch prediction function (takes name of old stitching function, `stitched_prediction`)
   - New `prediction_utils` package.
   - module `prediction_outputs` in new `prediction_utils` package.
- **Modified**: 
   - `Trainer` in `CAREamist` no longer has prediction_loop replaced
   - At the end of `CAREamist.predict` predictions are stitched and/or converted to match old `CAREamistPredictionLoop` outputs.
   - Creation of `CAREamicsPredictData` has been moved to new `prediction_utils` package
   - `stitch_prediction` function has been moved to `prediction_utils` package
- **Removed**: `CAREamistPredictionLoop` 
- **Tests** 
   - Added: test for new `stitched_prediction` function.
   - Added: test for prediction output conversion.
   - Modified: moved `stitched_prediction` tests to match new file structure of src.
   - Modified: `test_predict_on_array_tiled` and `test_predict_arrays_no_tiling` – parametrised with `samples` and `batch_size`; squeeze `train_array` when asserting size equality.

### Related Issues

- Resolves #140: Now a `BasePredictionWriter` Callback can be written more easily.
- Resolves #143: New logic stops tiles from being skipped at the end of batches.

### Breaking changes

Any code that instantiated a Lightning `Trainer` and added the `CAREamistPredictionLoop` .
There might be some unforeseen changes to dimensions of prediction outputs that the tests do not catch, i.e. adding S & C dims.

### TODO: (for future)
- Change prediction so that list is always output (currently, if there is only 1 prediction it will not output a list).
- Dimensions of outputs always have all SC(Z)YX, or match input dimensions. Currently, there is some inconsistency between tiled and not tiled prediction output dimensions.

- [x] Pre-commit passes


src/careamics/careamist.py
```diff
@@ -643,16 +634,41 @@ def predict(
...
+                for _, tile_info_list in predictions
+                for tile_info in tile_info_list
+            ]
+            predictions = torch.concatenate([preds for preds, _ in predictions])
+            # convert to numpy
+            predictions = predictions.cpu().numpy()
+            predictions = stitch_prediction(predictions, tile_infos)
+        else:
+            # convert to numpy
+            predictions = [pred.cpu().numpy() for pred in predictions]
```
jdeschamps: So not sure whether this happens somewhere else in the code, but if the predictions have all the same size we could concatenate them in a single array.

src/careamics/careamist.py
```diff
@@ -643,16 +634,41 @@ def predict(
...
+
+        # don't return list if single image
+        if len(predictions) == 1:
```
jdeschamps: Ignoring the previous comment: would it not make sense for the API to be clearer and always return a list of arrays?

Currently (and even more with the previous comment), users will have to check whether they received a list or a single numpy array. I would find this annoying.

I don't have a super strong opinion here.

jdeschamps (approved):

src/careamics/careamist.py
melisande-c: I added this because it was the previous behaviour of the CAREamics prediction loop, but I am happy to remove! or think about input -> output shapes and types a bit harder

src/careamics/careamist.py
melisande-c: I am changing a lot of stuff, this PR is actually a proper draft. I have put it up more to get an opinion on applying tiling after prediction before I put a lot of work in! So this is pretty rough and minimally functioning

melisande-c: Maybe comment on #140 if you have thoughts on this! Thanks!

src/careamics/careamist.py
jdeschamps: Yes I figured!

src/careamics/careamist.py
```diff
@@ -646,16 +637,42 @@ def predict(
...
                 )
 
+        # TODO: tidy up: move to a seperate function
+        predictions = self.trainer.predict(
+            model=self.model, datamodule=self.pred_datamodule, ckpt_path=checkpoint
+        )
+        # not sure if necessary
+        if len(predictions) == 0:
+            return predictions
+        is_tiled = tile_size is not None
```
CatEek: Can't we get tiling param from the datamodule ?

melisande-c: Yeah tbh, I did this pretty quick as proof of concept! Still a draft, I will fix up after I have tackled some other bugs

src/careamics/prediction_utils/prediction_outputs.py
```diff
@@ -0,0 +1,165 @@
...
+    if tiled:
+        predictions_comb = combine_batches(predictions, tiled)
+        # remove sample dimension (always 1) `stitch_predict` func expects no S dim
+        tiles = [pred[0] for pred in predictions_comb[0]]
+        tile_infos = predictions_comb[1]
+        predictions_output = stitch_prediction(tiles, tile_infos)
+    else:
+        predictions_output = combine_batches(predictions, tiled)
+
+    # TODO: add this in? Returns output with same axes as input
```
jdeschamps, line 44: Shouldn't the result of stitch prediction have a "S" dimensions if the input had any? If singleton then I guess it doesn't?

The output should already be shaped correctly. Now what we could do is to reshape back to the real input (which may have the axes in a weird order)... But for now it is better to stick to that.

jdeschamps: Oh I see now, this is not the `reshape` that we have elsewhere in the code. This is to add S and C back.

If the images are just returned to the user or saved to the disk, we are probably better off not returning singleton dims. It is annoying in terms of coherence, but much nicer for downstream.

jdeschamps (approved):
Looks good!

I am wondering a bit what the consequences are for the Lightning API. I guess they just need to call the `stitch` method. Since I anyway need to create examples and tests, I will make a PR in that direction as soon as this is merged.

src/careamics/prediction_utils/prediction_outputs.py
melisande-c: @jdeschamps 

I think there are some inconsistencies with output dims depending on whether the prediction was tiled or not. If the prediction was tiled, then because the function `stitch_prediction` doesn't expect the S dim, if there was only one image, then the output will also not have an S dim. However, if the prediction is not tiled, then the S dimension is not removed, even if there is only one image.

I have left it like this because I am pretty sure it is recreating how the code was previously behaving, but I can merge this PR and then create a new PR to fix this issue.

jdeschamps: Let's make an issue out of it so that we don't loose track of it! Also with respect to the list or not list return value.

I think for now it is ok to merge!

CatEek (approved):
lgtm

PR #152, opened 2024-06-18
Feature: Allow Noise2Noise with different in and out channels

jdeschamps:
### Description

Until now, Noise2Noise was only possible with the same number of input and output channels. But one can envision have multiple channels in (e.g. time frames) and supervising with a single frame.

This PR relaxes the constraints on Noise2Noise: channels in and out can now be different, and UNet is not mandatory!

- **What**: Remove `algorithm_model` constraint on `N2N`, and change the configuration convenience function.
- **Why**: We might try to run Noise2Noise with different in and out channels.
- **How**: See what.

### Changes Made

- **Modified**: `algorithm_model` and `create_n2n_configuration`, and the corresponding tests.

---

- [x] Pre-commit passes
    - https://github.com/CAREamics/careamics.github.io/pull/9
    - https://github.com/CAREamics/careamics-examples/pull/3

### DIFF
--- a/src/careamics/config/algorithm_model.py
+++ b/src/careamics/config/algorithm_model.py
@@ -134,21 +134,6 @@ class AlgorithmConfig(BaseModel):
                     "sure that `in_channels` and `num_classes` are the same."
                 )
 
-        # N2N
-        if self.algorithm == "n2n":
-            # n2n is only compatible with the UNet model
-            if not isinstance(self.model, UNetModel):
-                raise ValueError(
-                    f"Model for algorithm {self.algorithm} must be a `UNetModel`."
-                )
-
-            # n2n requires the number of input and output channels to be the same
-            if self.model.in_channels != self.model.num_classes:
-                raise ValueError(
-                    "N2N requires the same number of input and output channels. Make "
-                    "sure that `in_channels` and `num_classes` are the same."
-                )
-
         if self.algorithm == "care" or self.algorithm == "n2n":
             if self.loss == "n2v":
                 raise ValueError("Supervised algorithms do not support loss `n2v`.")
--- a/src/careamics/config/configuration_factory.py
+++ b/src/careamics/config/configuration_factory.py
@@ -243,19 +243,23 @@ def create_n2n_configuration(
     use_augmentations: bool = True,
     independent_channels: bool = False,
     loss: Literal["mae", "mse"] = "mae",
-    n_channels: int = 1,
+    n_channels_in: int = 1,
+    n_channels_out: int = -1,
     logger: Literal["wandb", "tensorboard", "none"] = "none",
     model_kwargs: Optional[dict] = None,
 ) -> Configuration:
     """
-    Create a configuration for training Noise2Noise.
+    Create a configuration for training CARE.
 
     If "Z" is present in `axes`, then `path_size` must be a list of length 3, otherwise
     2.
 
-    If "C" is present in `axes`, then you need to set `n_channels` to the number of
+    If "C" is present in `axes`, then you need to set `n_channels_in` to the number of
     channels. Likewise, if you set the number of channels, then "C" must be present in
     `axes`.
+
+    To set the number of output channels, use the `n_channels_out` parameter. If it is
+    not specified, it will be assumed to be equal to `n_channels_in`.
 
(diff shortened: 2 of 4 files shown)


jdeschamps:
Requires https://github.com/CAREamics/careamics/pull/149 to be merged for tests to pass.

src/careamics/config/algorithm_model.py
```diff
@@ -134,21 +134,6 @@ def algorithm_cross_validation(self: Self) -> Self:
                     "sure that `in_channels` and `num_classes` are the same."
                 )
 
-        # N2N
-        if self.algorithm == "n2n":
-            # n2n is only compatible with the UNet model
-            if not isinstance(self.model, UNetModel):
-                raise ValueError(
-                    f"Model for algorithm {self.algorithm} must be a `UNetModel`."
-                )
```
melisande-c, line 143: Did you mean to also remove this check?

jdeschamps: Yes!

The thought is that (as far as I thought) there is no reason to make N2N not compatible with the future VAE.

melisande-c: I see. In the convenience function `create_n2n_configuration`, that in turn calls `_create_supervised_configuration` a UNet is initialised (see code ref below). When the option of different models are fully implemented, should it be possible for the model to be selected by the user through the convenience function?

https://github.com/CAREamics/careamics/blob/8c344920387d964c28839c53152d286ad53af9fa/src/careamics/config/configuration_factory.py#L87-L98

src/careamics/config/configuration_factory.py
```diff
@@ -243,20 +243,24 @@ def create_n2n_configuration(
...
     loss: Literal["mae", "mse"] = "mae",
-    n_channels: int = 1,
+    n_channels_in: int = 1,
+    n_channels_out: int = -1,
     logger: Literal["wandb", "tensorboard", "none"] = "none",
     model_kwargs: Optional[dict] = None,
 ) -> Configuration:
     """
-    Create a configuration for training Noise2Noise.
+    Create a configuration for training CARE.
```
melisande-c: Has this been changed to CARE from Noise2Noise because of the relaxation in constraints? Is CARE the umbrella term for supervised algorithms and Noise2Noise specifically uses a UNet? (This is just a question for my benefit).

src/careamics/config/algorithm_model.py
jdeschamps: I would be in favor of keeping the convenience function like this while the VAE are not available, rather than having a parameter that will throw a `NotImplementedError` (which would be thrown if people try to use a VAE in a self crafter configuration).

Once the VAE is available, we would indeed need to reflect this in the convenience functions. (which means we would need to remember that, but worst case it comes up as an issue at some point).

Does that make sense?

src/careamics/config/configuration_factory.py
jdeschamps: Argh, that's a mistake! I probably copied the whole docstring, rather than the just the parameters.

CARE = any supervised image restoration in microscopy (the benefit of being the first paper to do it at scale for scientific bioimages), actually the original principle is really any method that leverages structural priors for image restoration (so not supervised could also be argued). We have a tendency to use it for the supervised part.

Noise2Noise = self-supervised denoising by training noisy image to noisy image supervision

src/careamics/config/algorithm_model.py
melisande-c: yes, I meant change it only once other options are available

src/careamics/config/configuration_factory.py
melisande-c: Ok thanks for the clarification! That is what I had originally thought :)

src/careamics/config/configuration_factory.py
```diff
@@ -293,8 +299,11 @@ def create_n2n_configuration(
     Returns
     -------
     Configuration
-        Configuration for training Noise2Noise.
+        Configuration for training CARE.
```
melisande-c: Wrong name here as well

jdeschamps: Damn it 😄

melisande-c (approved):

PR #153, opened 2024-06-19
Feature: Welford std approximation

CatEek:
### Description

Changed std computation for iterative dataset to use Welford algorithm. 

- **What**: Added Welford algo into iterable dataset
- **Why**:  Provides precise approximation of an std for a dataset of files

### Changes Made

- calculate_mean_std fun in iterable dataset module
- Relevant tests


src/careamics/dataset/dataset_utils/dataset_utils.py
```diff
@@ -121,3 +121,69 @@ def compute_normalization_stats(image: np.ndarray) -> Tuple[np.ndarray, np.ndarr
     # Define the list of axes excluding the channel axis
     axes = tuple(np.delete(np.arange(image.ndim), 1))
     return np.mean(image, axis=axes), np.std(image, axis=axes)
+
+
+def update_iterative_stats(
+    count: np.ndarray, mean: np.ndarray, m2: np.ndarray, new_values: np.ndarray
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """Update the mean and variance of an array iteratively.
```
melisande-c: Method should be referenced in docstring or at least comments.

src/careamics/dataset/dataset_utils/__init__.py
```diff
@@ -13,7 +15,12 @@
 ]
 
 
-from .dataset_utils import compute_normalization_stats, reshape_array
+from .dataset_utils import (
+    compute_normalization_stats,
```
jdeschamps: I think I'd prefer `from .data_utils.welford import ...`

For methods that are clearly part of a particular algorithm, I would like this to be clearly identified, even in the import!

src/careamics/dataset/dataset_utils/dataset_utils.py
jdeschamps: Agree, there should be mention of Welford algorithm and maybe a link to the wiki page (https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance)

src/careamics/dataset/iterable_dataset.py
```diff
@@ -125,36 +131,72 @@ def _calculate_mean_and_std(self) -> StatsOutput:
...
+                    count, mean, m2, sample_channels
+                )
 
             if target is not None:
-                target_mean, target_std = compute_normalization_stats(target)
-                target_means.append(target_mean)
-                target_stds.append(target_std)
+                target_init_mean, _ = compute_normalization_stats(target)
+                target_channels = np.array(np.split(target, target.shape[1], axis=1))
+                if num_samples == 0:
```
jdeschamps: This should be refactored somewhere else, especially because the whole procedure is called twice!

In the spirit of the previous comment, `update_iterative_stats` and the other method could become `_update_iterative_stats` and then a new method in `dataset_utils.welford` called `welford_mean_and_std(array, ...)` could do the calculation

tests/dataset/test_iterable_dataset.py
```diff
@@ -181,7 +174,67 @@ def test_compute_mean_std_transform_iterable(
...
+
+@pytest.mark.parametrize(
+    "shape, axes, patch_size",
+    [
+        ((32, 32), "YX", (8, 8)),
+        ((2, 32, 32), "CYX", (8, 8)),
+        ((32, 32, 32), "ZYX", (8, 8, 8)),
+    ],
+)
+def test_compute_mean_std_transform_iterable_with_targets(
```
jdeschamps: Also call the test Welford

src/careamics/dataset/dataset_utils/dataset_utils.py
```diff
@@ -121,3 +121,69 @@ def compute_normalization_stats(image: np.ndarray) -> Tuple[np.ndarray, np.ndarr
     # Define the list of axes excluding the channel axis
     axes = tuple(np.delete(np.arange(image.ndim), 1))
     return np.mean(image, axis=axes), np.std(image, axis=axes)
+
+
+def update_iterative_stats(
```
jdeschamps: Not sure whether that's "better", but another approach could be:

```python
class WelfordStatistics:

    count: int
    means: NDArray
    m2: NDArray

   def update(new_values) -> None:
        # update here

   def get_stats(self) -> Tuple[NDArray, NDArray]:
        # finalize here
    
```

src/careamics/dataset/dataset_utils/__init__.py
```diff
@@ -3,19 +3,25 @@
 __all__ = [
     "reshape_array",
     "compute_normalization_stats",
+    "update_iterative_stats",
```
jdeschamps: These are not imported here

src/careamics/dataset/dataset_utils/running_stats.py
```diff
@@ -0,0 +1,183 @@
...
+    computed per channel.
+
+    Parameters
+    ----------
+    image : NDArray
+        Input array.
+
+    Returns
+    -------
+    tuple[List[float], List[float]]
```
jdeschamps: can you write:
```
tuple of lists of floats
```

src/careamics/dataset/dataset_utils/running_stats.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Returns
+        -------
+        tuple[NDArray, NDArray]
+            Final mean and standard deviation.
+        """
+        return finalize_iterative_stats(self.count, self.mean, self.m2)
+
+
+# from multiprocessing import Value
```
jdeschamps, line 146: Is this kept for the Zarr?

src/careamics/dataset/dataset_utils/running_stats.py
```diff
@@ -0,0 +1,183 @@
...
+    std = np.array([np.sqrt(m / c) for m, c in zip(m2, count)])
+    if any(c < 2 for c in count):
+        return np.full(mean.shape, np.nan), np.full(std.shape, np.nan)
+    else:
+        return mean, std
+
+
+class WelfordStatistics:
+    """Compute Welford statistics iteratively."""
+
```
jdeschamps, line 97: can we explicitly have a place holder value in the class for
`self.num_samples`, `self.mean`, `self.count` and `self.m2` in an `__init__`? To make it clear! 

We should also write the docstring for the class.

src/careamics/dataset/dataset_utils/running_stats.py
```diff
@@ -0,0 +1,183 @@
...
+        ----------
+        array : NDArray
+            Input array.
+        num_samples : int
+            Current sample number.
+        """
+        self.num_samples = num_samples
+        sample_channels = np.array(np.split(array, array.shape[1], axis=1))
+
+        if self.num_samples == 0:
```
jdeschamps: If `self.mean`, `self.count` and `self.m2` have decent default values (I guess `0`), can't we just call `update_iterative_stats`? (or at least on each channel)

src/careamics/dataset/dataset_utils/running_stats.py
```diff
@@ -0,0 +1,183 @@
...
+                [
+                    np.sum(
+                        np.subtract(channel.flatten(), [self.mean[i]] * self.count[i])
+                        ** 2
+                    )
+                    for i, channel in enumerate(sample_channels)
+                ]
+            )
+        else:
+            self.count, self.mean, self.m2 = update_iterative_stats(
```
jdeschamps, line 129: Is this somewhat running on the channels independently?

src/careamics/dataset/dataset_utils/running_stats.py
melisande-c: I think actually
```
tuple of (list of floats, list of floats)
```

jdeschamps: Even better!

jdeschamps (approved):

PR #161, opened 2024-06-21
Refactoring: lightning API package and smoke tests

jdeschamps:
### Description

I moved the lightning modules into their own package to clarify the imports for the different API, and added smoke tests for the Lightning API. As a result I needed to implement a new function to retrieve the statistics from the `CAREamicsTrainData`, which led me to refactor a bit the datasets and standardize how the statistics are recorded across the two datasets.

The smoke tests check that the Lightning API works for tiled and un-tiled data.

Finally, the lightning modules wrappers have disappeared, and now convenience functions explicit the parameters. This led to renaming of the lightning modules.

- **What**:
    - Moved Lightning modules and callbacks to `careamics.lightning`
    - Added tests for the Lightning API (tiled and un-tiled prediction included)
    - Normalize the way the datasets keep the image statistics.
    - Fixed some errors in `CAREamicsTrainData`
- **Why**: Since we are differentiating between CAREamist and Lightning APIs, they should be imported from different packages.
- **How**: 
    - Moved Lightning modules and callbacks to `careamics.lightning`
    - Added tests for Ligthning API.
    - Removed `StatsOutput`, both datasets now have a `self.image_stats: Stats` and `self.target_stats: Stats`
    - Added `get_data_statistics` to the datasets and `CAREamicsTrainData` and corresponding test

### Changes Made

- **Added**: `test_lightning_api.py`.
- **Modified**: 
    - `ligthning_data_module.py`
    - `Iterable_dataset.py`
    - `in_memory_dataset.py`
    - `patching.py`
    - imports throughout the careamist and other files

To use the Lightning API, users now need to have the following imports:

```python
from pytorch_lightning.callbacks import ModelCheckpoint

from careamics.lightning import (
    CAREamicsModuleWrapper,
    TrainingDataWrapper,
    PredictDataWrapper,
)
from careamics.prediction_utils import convert_outputs # if tiling required
```

(see [full example here](https://github.com/CAREamics/careamics-examples/blob/jd/lightning_api/applications/lightning_api/2D/BSD68_Noise2Void_lightning_api.ipynb))

Since the `PredictDataWrapper` requires passing the statistics, users can now simply call:

```python
means, stds = train_data_wrapper.get_data_statistics()
```

While building the notebook, I uncovered the following error: https://github.com/CAREamics/careamics/issues/162

---

- [x] Pre-commit passes
    - https://github.com/CAREamics/careamics-examples/pull/6
    - https://github.com/CAREamics/careamics.github.io/pull/11


jdeschamps:
This PR might interact with https://github.com/CAREamics/careamics/pull/153

src/careamics/dataset/patching/patching.py
```diff
@@ -20,21 +20,26 @@ class Stats:
...
-    targets: Union[np.ndarray, None]
-    image_stats: Stats
-    target_stats: Stats
+        return list(self.means), list(self.stds)
 
 
 @dataclass
-class StatsOutput:
+class PatchedOutput:
     """Dataclass to store patches and statistics."""
```
melisande-c, line 44: I'm not totally familiar with the patching and I could do with a little more explicit documentation here! i.e. `image_stats` are per channel over all the input samples, similar for `target_stats`. I think documenting the expected dimensions of patches and targets could be helpful too.

src/careamics/dataset/patching/patching.py
```diff
@@ -20,21 +20,26 @@ class Stats:
...
+    def get_statistics(self) -> tuple[list[float], list[float]]:
+        """Return the means and standard deviations.
 
-@dataclass
-class PatchedOutput:
-    """Dataclass to store patches and statistics."""
+        Returns
+        -------
+        tuple of two lists of floats
+            Means and standard deviations.
```
melisande-c, line 34: Again documentation, this is means per channel?

melisande-c (commented):
I have a couple notes on the lightning API that are maybe not relevant to this PR since they weren't altered here. 

1) This is just my opinion so feel free to ignore: but I find all the "`Wrapper`" type classes unnecessary and confusing. These include the `TrainingDataWrapper`, `CAREamicsModuleWrapper` and `PredictDataWrapper`. The purpose of these classes is to initialise the classes they claim to wrap with a different set of parameters and this can easily be achieved with a function. I feel subclassing should be primarily be reserved for extending or modifying the functionality of the parent class. 

For example a `CAREamicsTrainData` factory function could reimplement what is in the `__init__` of `TrainingDataWrapper`, but return a `CAREamicsTrainData` object, as such:
```python
def create_train_datamodule(
  train_data: Union[str, Path, np.ndarray],
  data_type: Union[Literal["array", "tiff", "custom"], SupportedData],
  patch_size: List[int],
  axes: str,
  batch_size: int,
  val_data: Optional[Union[str, Path]] = None,
  transforms: Optional[List[TRANSFORMS_UNION]] = None,
  train_target_data: Optional[Union[str, Path]] = None,
  val_target_data: Optional[Union[str, Path]] = None,
  read_source_func: Optional[Callable] = None,
  extension_filter: str = "",
  val_percentage: float = 0.1,
  val_minimum_patches: int = 5,
  dataloader_params: Optional[dict] = None,
  use_in_memory: bool = True,
  use_n2v2: bool = False,
  struct_n2v_axis: Literal["horizontal", "vertical", "none"] = "none",
  struct_n2v_span: int = 5,
) -> CAREamicsTrainData:
  """Create `CAREamicsTrainData`."""
   # logic from `TrainingDataWrapper.__init__`
  ...
  return CAREamicsTrainData(
    data_config=data_config,
    train_data=train_data,
    val_data=val_data,
    train_data_target=train_target_data,
    val_data_target=val_target_data,
    read_source_func=read_source_func,
    extension_filter=extension_filter,
    val_percentage=val_percentage,
    val_minimum_split=val_minimum_patches,
    use_in_memory=use_in_memory
  )
```

2) Not a priority but I will mention anyway: I think the `CAREamicsTrainData.setup` method is pretty difficult to read with all the nested "if else" statements 😅. Maybe a strategy pattern could help here in future, especially if there are more dataset types to come, i.e. zarr !

jdeschamps:
> I have a couple notes on the lightning API that are maybe not relevant to this PR since they weren't altered here.
> 
>     1. This is just my opinion so feel free to ignore: but I find all the "`Wrapper`" type classes unnecessary and confusing. These include the `TrainingDataWrapper`, `CAREamicsModuleWrapper` and `PredictDataWrapper`. The purpose of these classes is to initialise the classes they claim to wrap with a different set of parameters and this can easily be achieved with a function. I feel subclassing should be primarily be reserved for extending or modifying the functionality of the parent class.
> 
> 
> For example a `CAREamicsTrainData` factory function could reimplement what is in the `__init__` of `TrainingDataWrapper`, but return a `CAREamicsTrainData` object, as such:
> 
> ```python
> def create_train_datamodule(
>   train_data: Union[str, Path, np.ndarray],
>   data_type: Union[Literal["array", "tiff", "custom"], SupportedData],
>   patch_size: List[int],
>   axes: str,
>   batch_size: int,
>   val_data: Optional[Union[str, Path]] = None,
>   transforms: Optional[List[TRANSFORMS_UNION]] = None,
>   train_target_data: Optional[Union[str, Path]] = None,
>   val_target_data: Optional[Union[str, Path]] = None,
>   read_source_func: Optional[Callable] = None,
>   extension_filter: str = "",
>   val_percentage: float = 0.1,
>   val_minimum_patches: int = 5,
>   dataloader_params: Optional[dict] = None,
>   use_in_memory: bool = True,
>   use_n2v2: bool = False,
>   struct_n2v_axis: Literal["horizontal", "vertical", "none"] = "none",
>   struct_n2v_span: int = 5,
> ) -> CAREamicsTrainData:
>   """Create `CAREamicsTrainData`."""
>    # logic from `TrainingDataWrapper.__init__`
>   ...
>   return CAREamicsTrainData(
>     data_config=data_config,
>     train_data=train_data,
>     val_data=val_data,
>     train_data_target=train_target_data,
>     val_data_target=val_target_data,
>     read_source_func=read_source_func,
>     extension_filter=extension_filter,
>     val_percentage=val_percentage,
>     val_minimum_split=val_minimum_patches,
>     use_in_memory=use_in_memory
>   )
> ```
> 
>     2. Not a priority but I will mention anyway: I think the `CAREamicsTrainData.setup` method is pretty difficult to read with all the nested "if else" statements 😅. Maybe a strategy pattern could help here in future, especially if there are more dataset types to come, i.e. zarr !

I absolutely agree, the whole dataset handling has been a mess since they have been implemented and we are slowly refactoring things...

I like the idea of the function rather than the subclassing. It would allow us to come up with better names for the classes as well (since now there would be only a single one).

src/careamics/dataset/patching/patching.py
jdeschamps: Agreed, this part is inherited from another PR that was merged. I'll give it a go.

jdeschamps:
Update: the wrappers have made space for functions, this created a duplication with a method used in CAREamics. The duplication was solved via some refactoring.

src/careamics/dataset/patching/patching.py
jdeschamps: All of this changed in the Welford PR, I made now another pass to the docstrings.

src/careamics/config/configuration_factory.py
```diff
@@ -641,13 +642,13 @@ def create_inference_configuration(
...
+    return {
+        "data_type": data_type or configuration.data_config.data_type,
+        "tile_size": tile_size,
+        "tile_overlap": tile_overlap,
+        "axes": axes or configuration.data_config.axes,
+        "image_means": configuration.data_config.image_means,
+        "image_stds": configuration.data_config.image_stds,
+        "tta_transforms": tta_transforms,
+        "batch_size": batch_size or configuration.data_config.batch_size,
+    }
```
melisande-c: What's the reasoning for not returning the `InferenceConfig` object anymore?

Additionally, is it not possible to have the validation of for the tile size/increment in the case of UNets directly in the Pydantic `InferenceConfig`? Same for existence of `image_means`/`image_stds`

Edit: The validation happens a few lines above what I commentated on.

melisande-c: Ok, I now see that it is to mirror the same pattern in the `create_train_datamodule` and the `create_predict_datamodule`. (Comparison below)

https://github.com/CAREamics/careamics/blob/d0369f7ffb666d87176c8abea4ddd32106b5571f/src/careamics/lightning/train_data_module.py#L648-L649

https://github.com/CAREamics/careamics/blob/d0369f7ffb666d87176c8abea4ddd32106b5571f/src/careamics/lightning/predict_data_module.py#L321-L322

However, if all the validation can be done in `InferenceConfig` then there is no need for the `create_inference_parameters` function. In `CAREamist.predict`, all the parameters can be passed to `create_predict_datamodule` directly, including `image_means` and `image_std` which is accessed from the `cfg` attribute of the `CAREamist` instance. Then the validation happens in the instantiation of the `InferenceConfig`.

https://github.com/CAREamics/careamics/blob/d0369f7ffb666d87176c8abea4ddd32106b5571f/src/careamics/careamist.py#L609-L616

src/careamics/careamist.py
```diff
@@ -592,28 +595,27 @@ def predict(
...
             tile_overlap=tile_overlap,
-            axes=axes,
             data_type=data_type,
+            axes=axes,
             tta_transforms=tta_transforms,
+            batch_size=batch_size,
+        )
+
+        # create the prediction
+        self.pred_datamodule = create_predict_datamodule(
```
melisande-c: I've just noticed do we have separate definitions of "inference" and "prediction" or specific scenarios we use either one for, or are they just interchangeable? Not very important as everyone should understand what they mean but might be good to keep some consistency.

src/careamics/lightning/lightning_module.py
```diff
@@ -202,101 +202,74 @@ def configure_optimizers(self) -> Any:
...
+    model_parameters: Optional[dict] = None,
+    optimizer: Union[SupportedOptimizer, str] = "Adam",
+    optimizer_parameters: Optional[dict] = None,
+    lr_scheduler: Union[SupportedScheduler, str] = "ReduceLROnPlateau",
+    lr_scheduler_parameters: Optional[dict] = None,
+) -> CAREamicsModule:
+    """Return a CAREamics Lithgning module.
 
-    This class exposes parameters used to create an AlgorithmModel instance, triggering
+    This method exposes parameters used to create an AlgorithmModel instance, triggering
```
melisande-c: in standard python terminology this is technically a function, a method is bound to a class 😁 (sorry to be pydantic ...)

src/careamics/config/configuration_factory.py
jdeschamps: I agree that is unnecessarily complex....

Probably best indeed to roll back from the even-earlier refactoring step (where what is now `create_inference_parameters` once lived in `CAREamist`).

src/careamics/careamist.py
jdeschamps: I think it stems from me preferring `prediction` and Igor preferring `inference`, at least at some point in the past.

src/careamics/lightning/lightning_module.py
jdeschamps: True!

jdeschamps:
Removed `create_inference_config` and moved the code into the careamics predict function, now `InferenceConfig` is only instantiated in `create_predict_datamodule`.

And I corrected the vocab. 😉

src/careamics/careamist.py
```diff
@@ -594,25 +598,59 @@ def predict(
...
+        ):
+            raise ValueError("Mean and std must be provided in the configuration.")
+
+        # tile size for UNets
+        if tile_size is not None:
+            model = self.cfg.algorithm_config.model
+
+            if model.architecture == SupportedArchitecture.UNET.value:
+                # tile size must be equal to k*2^n, where n is the number of pooling
+                # layers (equal to the depth) and k is an integer
```
melisande-c, line 623: Ah, I realise now the `InferenceConfig` doesn't have access to the type of architecture so it can't be validated in there 😅, I guess it might be nice to isolate this check as a separate function somewhere but otherwise looks good!

melisande-c (approved):

src/careamics/careamist.py
jdeschamps: Not only the type of architecture, but also the depth of the UNet...

But you had a good point with the double `Inference` instantiation, that was convoluted. Let's see later if we want to move that to some utils package or so.

PR #164, opened 2024-06-23
Fix: Fix TTA and add tests

### LINKED ISSUE #157: Incorrect TTA flip transform [BUG]
**Describe the bug**
TTA transforms perform flip on -1 and -3 axis. See code ref below. -3 axis can be Z or C. We do not want to be performing the flip on C, and we potentially do not want to be performing flip on Z depending on imaging optics.

https://github.com/CAREamics/careamics/blob/eea3ecd6d4668d484a2610ac6cd8cdb38db7259f/src/careamics/transforms/tta.py#L26-L49

**Expected behavior**
Flips should happen on the YX axis only, and potentially configurable Z flips.
Maybe some more documentation in `CAREamist.predict` on why you would or would not use TTA transforms.

jdeschamps:
### Description

Following https://github.com/CAREamics/careamics/issues/157, this PR fixes the axes for the flipping and add tests.

- **What**: Fixing test-time augmentation used during the prediction step.
- **Why**: The axes of some of the transformations were wrong.
- **How**: Explicitly write the augmentations and add tests.

### Changes Made

- **Added**: `test_tta.py`.
- **Modified**: `CAREamist.predict` tta explanations, `ImageRestorationTTA`.

### Related Issues

- Fixes https://github.com/CAREamics/careamics/issues/157

---

- [x] Pre-commit passes


src/careamics/transforms/tta.py
```diff
@@ -1,74 +1,88 @@
...
-            rot90(x, 3, dims=(-2, -1)),
+            # original
+            input_tensor,
+            # rotations
+            rot90(input_tensor, 1, dims=axes),
+            rot90(input_tensor, 2, dims=axes),
+            rot90(input_tensor, 3, dims=axes),
+            # original flipped
+            flip(input_tensor, dims=(axes[0],)),
+            flip(input_tensor, dims=(axes[1],)),
```
melisande-c, line 45: This doesn't actually cover every unique transformation.
There are 8 unique transformations (some combinations result in the same transformation e.g. flip X followed by flip Y is equivalent to rotate 180).
To cover every possible combination, do all the rotations and then the combination of each rotation with a flip.
```python
augmented = [
   # original
   input_tensor,
   # rotations
   rot90(input_tensor, 1, dims=axes),
   rot90(input_tensor, 2, dims=axes),
   rot90(input_tensor, 3, dims=axes),
   # rotations + flip
   flip(input_tensor, dims=(-1, )),
   flip(rot90(input_tensor, 1, dims=axes), dims=(-1, )),
   flip(rot90(input_tensor, 2, dims=axes), dims=(-1, )),
   flip(rot90(input_tensor, 3, dims=axes), dims=(-1, )),
]
```

jdeschamps: Correct, but it is done in the lines that follow 😃 
https://github.com/CAREamics/careamics/blob/cfe269fcdc72d02e2f65490ad656262d5ab79692/src/careamics/transforms/tta.py#L49

```python
        augmented.extend(
            [
                flip(augmented[1], dims=(axes[0],)),
                flip(augmented[1], dims=(axes[1],)),
            ]
        )
```

I didn't want to recompute the rotations again (which might not be computationally faster...).

melisande-c: My mistake! I guess I read it too quickly

> - **How**: Explicitly write the augmentations

I partly assumed this from your description in the PR.

melisande-c (approved):

src/careamics/transforms/tta.py
jdeschamps: Ahah ouch, point taken. 😄

PR #170, opened 2024-06-25
Fix: Inconsistent prediction outputs

### LINKED ISSUE #156: Inconsistent prediction outputs [BUG]
**Describe the bug**
The output dimensions of `CAREamics.predict` are different depending on whether tiling was used to compute the prediction or not. Additionally, if the no. of samples is greater than 1 then the output is a list, if there is only 1 sample then the prediction array is returned, not contained in a list.

If tiling is not used, the dimensions will always be SC(Z)YX; if the input did not have the C dim it is set to 1, and S is always 1 (because predictions are returned as a list i.e. a list of length S (unless S=1) of arrays with shape (1, C, Y, X). If tiling is used then the S dimension is always removed, and if C is 1, then it is removed.

**To Reproduce**
***Tiled vs non-tiled***
```python
import numpy as np

from careamics.config import create_n2v_configuration

    experiment_name="PredBatchingTest", 
    data_type="array",
    axes="SYX",
    patch_size=[8, 8],
    batch_size=1,
    num_epochs=1,
    n_channels=1,
    )
image = np.random.random((1, 32, 32))

engine = CAREamist(source=config)
engine.train(train_source=image)
pred_tiled = engine.predict(source=image, tile_size=(8, 8), tile_overlap=(2,2))
pred = engine.predict(source=image)

print("Tiled shape: ", pred_tiled.shape)
print("Non-tiled shape: ", pred.shape)
```
```
>>> ...
>>> Tiled shape:  (32, 32)
>>> Non-tiled shape:  (1, 1, 32, 32)
```
***List output***
```
import numpy as np

from careamics.config import create_n2v_configuration

    experiment_name="PredBatchingTest", 
    data_type="array",
    axes="SYX",
    patch_size=[8, 8],
    batch_size=1,
    num_epochs=1,
    n_channels=1,
    )
image = np.random.random((2, 32, 32))

engine = CAREamist(source=config)
engine.train(train_source=image)
pred = engine.predict(source=image)

print("Output type: ", type(pred))
print("Output length: ", len(pred))
print("Prediction shape: ", pred[0].shape)
```
```
>>> ...
>>> Output type:  <class 'list'>
>>> Output length:  2
>>> Prediction shape:  (1, 1, 32, 32)
```

**Expected behavior**
Prediction outputs should be consistent. 

I suggest that the outputs should always be contained in a list, even if there is only 1 sample. I would not concatenate the samples because there is a possibility that, when predicting on a set of files, the samples have different YX dimensions.

I suggest that the samples should always have the channel dimension added, even if the input did not have channels; but the S dimension should be removed since it will always be 1, since the channels are returned in a list.

**Causes**
The function `stitch_prediction` assumes there is no S dim, so it is removed from the inputs before calling the function and it is never added back. I am not sure how the C dimension is removed.

melisande-c:
### Description

Prediction outputs were inconsistent. Firstly, if there was only 1 sample the output would not be a list. Secondly, and less predictably, if the prediction was tiled, then singleton channel dimensions would be dropped. This PR makes sure that the output of `CAREamist.predict` is always a list and that the outputs always have the same dimensions.

**What**:
- The `stitched_prediction` function will now output arrays with dimensions `SC(Z)YX` where S will always be singleton.
- Tests have been updated to always expect lists from `CAREamist.predict`. Additionally, instead of squeezing prediction outputs, the input arrays are reshaped when comparing dimensions; this should make the tests a little more robust.

**Why**:
- Having `stitched_prediction` output `SC(Z)YX` arrays simplifies the `prediction_utils.convert_outputs` function, in which, if a prediction has 3 dimensions it can be ambiguous whether the first dimension is Z or C. If the first dimension is Z then we need to add both the S & C singleton dims, and if the first dimension is C then we only need to add the S singleton dimension.

**How**: 
- Solved not very elegantly by expanding the `predicted_image` dimensions, based on the first tile's dimensions, in `stitch_prediction_single`. 
- This can be solved more elegantly by allowing `TileInformation.array_shape` to have singleton dimensions.

### Changes Made

**Modified**: 
  - `careamics.prediction_utils.convert_outputs`
  - `careamics.prediction_utils.stitch_prediction_single`
  - `CAREamsit.export_to_bmz`
  - tests  

### Related Issues

- Fixes #156 
- `stitch_prediction_single` now accepts inputs with dimensions `SC(Z)YX`. There was an incorrect assumption that C was the first axis before this PR.

### Breaking changes

Code working with prediction outputs.

Related to allowing `TileInformation.array_shape` to have singleton dimensions, in my opinion, after data is initially input and reshaped, we should always assume all internal code is working with data in the form SC(Z)YX. This will reduce some ambiguity and confusion. 

The prediction output dimensions could be changed to C(Z)YZ as I initially suggested in #156, since S is always 1. This would mean replacing `np.concatenate` with `np.stack` in a lot of places that combine prediction samples.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -661,7 +661,7 @@ class CAREamist:
         """
         # TODO: add in docs that it is expected that input_array dimensions match
         # those in data_config
-     
+
         output_patch = self.predict(
             input_array,
             data_type=SupportedData.ARRAY.value,
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -659,27 +659,16 @@ class CAREamist:
         data_description : str, optional
             Description of the data, by default None.
         """
-        input_patch = reshape_array(input_array, self.cfg.data_config.axes)
-
-        # axes need to be reformated for the export because reshaping was done in the
-        # datamodule
-        if "Z" in self.cfg.data_config.axes:
-            axes = "SCZYX"
-        else:
-            axes = "SCYX"
-
-        # predict output, remove extra dimensions for the purpose of the prediction
+        # TODO: add in docs that it is expected that input_array dimensions match
+        # those in data_config
+     
         output_patch = self.predict(
-            input_patch,
+            input_array,
             data_type=SupportedData.ARRAY.value,
-            axes=axes,
             tta_transforms=False,
         )
-
-        if isinstance(output_patch, list):
-            output = np.concatenate(output_patch, axis=0)
-        else:
-            output = output_patch
+        output = np.concatenate(output_patch, axis=0)
+        input_array = reshape_array(input_array, self.cfg.data_config.axes)
 
         export_to_bmz(
             model=self.model,
@@ -688,7 +677,7 @@ class CAREamist:
             name=name,
             general_description=general_description,
             authors=authors,
-            input_array=input_patch,
+            input_array=input_array,
             output_array=output,
             channel_names=channel_names,
             data_description=data_description,
--- a/src/careamics/prediction_utils/prediction_outputs.py
+++ b/src/careamics/prediction_utils/prediction_outputs.py
@@ -96,8 +96,8 @@ def _combine_tiled_batches(
     ----------
     predictions : list of (numpy.ndarray, list of TileInformation)
         Predictions that are output from `Trainer.predict`. For tiled batches, this is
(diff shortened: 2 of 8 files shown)


jdeschamps (approved):

jdeschamps:
Looks great!

Let's open another issue for further simplifying the prediction via the `TileInformation` as you suggested.

PR #174, opened 2024-07-04
Feature: Add File IO package

melisande-c:
### Description

This is a precursor for the soon to come save prediction feature. Save functions needed to be added; then because they mirror the structure of the read functions, the read functions have been moved from `dataset.dataset_utils` to a new `file_io` package. This package has `read` and `write` subpackages that mirror each other's structures, for ease of understanding.

Additionally, the `get_read_func` is slightly refactored to match how the `get_write_func` is implemented. Read functions are stored in a module level dictionary with the keys being `SupportedData`, the `get_read_func` indexes this dictionary based on the `data_type` passed. This removes the eventuality of a long list of if/else statements. This wasn't strictly necessary as we do not plan to support a large number of data types, but the option is always there.

An additional extra change that snuck into this PR is renaming `SupportedData.get_extension` to `SupportedData.get_extension_pattern`, and adding a different `SupportedData.get_extension`. The new `SupportedData.get_extension` returns the literal string without the unix wildcard patterns and will be used for saving predictions in a future PR.

- **What**: 
  - Added a `file_io` package to contain functions to read and write image files. 
  - `SupportedData.get_extension` modification and addition.
- **Why**: Partly an aesthetic choice, removes the responsibility of file reading and writing from the `datasets` package in accordance with trying to follow the single-responsibility principle.
- **How**: Added write functions. Moved and slightly refactored read functions.

### Changes Made

- **Added**:
  - `file_io`, `file_io.read`, `file_io.write` packages.
  - `write_tiff` function
  - `get_write_func` function
  - New `SupportedData.get_extension` to return literal extension string
  - Tests for all the above new functions
- **Modified**:
  - Old `SupportedData.get_extension` renamed to `SupportedData.get_extension_pattern`
  - Renamed tests to mirror name change in function
- **Removed**: 
  - File reading from `datasets.dataset_utils`

### Breaking changes

- Code calling read functions from `datasets.dataset_utils` directly.
- Code using `SupportedData.get_extension` directly.

In a future PR `SupportedData.get_extension` and `SupportedData.get_extension_pattern` could be moved to the `file_io` package. These functions do not need to be bound to `SupportedData` and none of the other "support" `Enum` classes have additional methods. It might make sense to store these functions closer to where they are used. This is again a stylistic choice, feel free to share other opinions.

---

- [x] Pre-commit passes


src/careamics/config/support/supported_data.py
```diff
@@ -60,9 +60,9 @@ def _missing_(cls, value: object) -> str:
         return super()._missing_(value)
 
     @classmethod
-    def get_extension(cls, data_type: Union[str, SupportedData]) -> str:
+    def get_extension_pattern(cls, data_type: Union[str | SupportedData]) -> str:
```
jdeschamps: `Union[A | B]` is probably a typo?

Since we are still supporting py9, so we should avoid using pipes for Unions.

src/careamics/config/support/supported_data.py
```diff
@@ -82,3 +82,28 @@ def get_extension(cls, data_type: Union[str, SupportedData]) -> str:
...
+        data_type : str or SupportedData
+            Data type.
+
+        Returns
+        -------
+        str
+            Corresponding extension.
+        """
+        if data_type == cls.ARRAY:
+            raise NotImplementedError(f"Data {data_type} are not loaded from file.")
```
jdeschamps: Ok, this comment is probably going to much into the details... Shouldn't it be "files" plural? (I'll trust the native speaker there, just my feelings). Please disregard if disagree!

src/careamics/config/support/supported_data.py
```diff
@@ -82,3 +82,28 @@ def get_extension(cls, data_type: Union[str, SupportedData]) -> str:
...
+        elif data_type == cls.TIFF:
+            return ".tiff"
+        elif data_type == cls.CUSTOM:
+            # TODO: improve this message
+            raise NotImplementedError("Custom extensions have to be passed elsewhere.")
```
jdeschamps, line 107: Can we give more details in the error message? Right now I am not sure yet where this method is called, but we are pretty consistent about the `extension_filter` parameter for the `get_extension_pattern` method.

jdeschamps: ah nevermind, there is a todo!

src/careamics/file_io/read/get_func.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import Callable, Dict, Protocol, Union
+
+from numpy.typing import NDArray
+
+from careamics.config.support import SupportedData
+
+from .tiff import read_tiff
+
+
+# This is very strict, arguments have to be called img
```
jdeschamps: What does "img" mean in this context?

jdeschamps: oh, it's probably `file_path` here!

jdeschamps (approved):
I like that it defines strictly the signature of a read/write function. Will make the custom data examples much cleaner for users.

We should have an update to the doc asap, before we forget this happened!

src/careamics/config/support/supported_data.py
melisande-c: Oh yeah oops, not sure how this happened

src/careamics/config/support/supported_data.py
melisande-c: I copy and pasted this from the original function above.

https://github.com/CAREamics/careamics/blob/6ff1d4b54f6491e71cc3360148ea946f76ccef1a/src/careamics/config/support/supported_data.py#L78

I think I will change both to "Data {data_type} _is_ not loaded from a file."

src/careamics/file_io/read/get_func.py
melisande-c: Ah yes I think I copy and pasted my comment from the `WriteFunc` description incorrectly, will update

src/careamics/config/support/supported_data.py
melisande-c: Yes this function isn't called anywhere yet, so I decided to leave the message as a future job, especially because there might be some refactoring. 

In future it will be used to add the correct extension when saving predictions. It made it into this PR because I was considering putting it directly into the `write_tiff` function by doing something like `file_path.with_suffix(SupportedData.get_extension(data_type))`. But then I decided against it, and instead thought file paths should be correct before being passed to the save function, because of how custom save functions will be called.

src/careamics/config/support/supported_data.py
jdeschamps: Sounds good!

The code base is probably riddled with approximate english. 😄

PR #179, opened 2024-07-08
Feature: Prediction writer callback initialisation

melisande-c:
### Description

Add the skeleton and initialisation of the `PredictionWriterCallback` class, to allow users to save predictions to disk in the future. This callback will be used by the `CAREamist` class but should also be compatible with users preferring to use the internal `lightning` API.

- **What**: 
  - Add `PredictionWriterCallback` class.
  - Implement `PredictionWriterCallback.__init__` method.
  - Add placeholder for `write_on_batch_end` hook, with docstring description of what is coming.
  - Implement `setup` hook that will create the prediction output directory when the `lightning` predict stage starts.
  - Tests for `PredictionWriterCallback.__init__`.
- **Why**: 
  - This callback will be used to save predictions to disk, this is a feature in the works for a while. This will also allow the predict command to be implemented in the CLI.

### Changes Made

- **Added**: 
  - `PredictionWriterCallback` class to the `lightning.callbacks` package.
  - Tests for the initialisation of `PredictionWriterCallback`.
- **Modified**: 
  - Package/sub-package level imports (__init__.py files.)

#### Future considerations

- There will have to be a method to update `save_type`, `save_extension`, `save_func` & `save_func_kwargs` because in the `CAREamist` class they will be decided after the initialisation of `PredictionWriterCallback`. The `CAREamist` class will have a `predict_to_disk` method that where users can pass these parameters, but the `PredictionWriterCallback` will have to be initialised during `CAREamist` initialisation because it needs to be passed to the trainer.
- For tiling/not tiling & tile caching/writing tiles directly to disk (`zarr`). I want to follow a [strategy pattern](https://refactoring.guru/design-patterns/strategy) so `PredictionWriterCallback` will interact with a strategy interface and it's only responsibility will be to select the correct one.
- As mentioned in the description, this callback should be easily usable by both the `CAREamist` class and users who are using the `lightning` API.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/lightning/callbacks/prediction_writer_callback.py
+++ b/src/careamics/lightning/callbacks/prediction_writer_callback.py
@@ -1,3 +1,5 @@
+"""Module containing PredictionWriterCallback class."""
+
 from typing import Union, Optional, Any
 from pathlib import Path
 
@@ -159,9 +161,10 @@ class PredictionWriterCallback(BasePredictionWriter):
                     "A save extension must be provided for custom data types."
                 )
             else:
-                self.save_func = save_extension
+                self.save_extension = save_extension
         else:
-            self.save_func = self.save_type.get_extension()
+            # kind of a weird pattern -> reason to move get_extension from SupportedData
+            self.save_extension = self.save_type.get_extension(self.save_type)
 
     def setup(self, trainer: Trainer, pl_module: LightningModule, stage: str) -> None:
         """
@@ -176,5 +179,5 @@ class PredictionWriterCallback(BasePredictionWriter):
                 logger.info("Making prediction output directory.")
                 self.dirpath.mkdir()
 
-        
+    # TODO: write hook placeholders with TODO comments 
 
--- a/src/careamics/lightning/callbacks/prediction_writer_callback.py
+++ b/src/careamics/lightning/callbacks/prediction_writer_callback.py
@@ -1,0 +1,139 @@
+from typing import Union, Optional, Any
+from pathlib import Path
+
+from pytorch_lightning import Trainer, LightningModule
+from pytorch_lightning.callbacks import BasePredictionWriter
+
+from careamics.config.support import SupportedData
+from careamics.file_io import get_write_func, WriteFunc
+from careamics.utils import get_logger
+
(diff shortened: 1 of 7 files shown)


src/careamics/lightning/callbacks/prediction_writer_callback.py
```diff
@@ -0,0 +1,242 @@
...
+        The extension that will be added to the save paths.
+    save_func_kwargs : dict of {str, any}
+        Additional keyword arguments that will be passed to the save function.
+    dirpath : pathlib.Path
+        The path to the directory where prediction outputs will be saved.
+    """
+
+    def __init__(
+        self,
+        save_type: Union[SupportedData, str] = "tiff",
```
jdeschamps, line 55: Should we use `Literal["tiff", "custom"]` instead of `str`?

Then the SupportedData is not exactly equivalent, so we could also just put `Liberal["array", "tiff", "custom]`.

jdeschamps (approved):
Looks good, sorry for the delay.

src/careamics/lightning/callbacks/prediction_writer_callback.py
melisande-c: Yes I think maybe for all the user facing functions use `Literal`, have the value converted to `SupportedData` early on, and all internal functions expect `SupportedData` type.

jdeschamps: That would make a lot of sense, yes!

melisande-c:
@jdeschamps 

I have already started on the "write strategy"s in [this branch](https://github.com/CAREamics/careamics/tree/mc/feat/save_strategy). 

It did require a fairly big rework of my idea for `PredictionWriterCallback`; so I can either merge this PR and then make a new PR for the new work, or I can update this PR with the new work. What do you think?

jdeschamps:
I don't really have an opinion. Probably merging and making a new PR is easier?

Whatever is easier for you (writing a new PR comment vs updating it)!

melisande-c:
In the spirit of small PRs I will merge, and write a new one.

PR #180, opened 2024-07-11
LVAE losses tests

CatEek:
### Description

- **What**:  Moved losses/NMs/LL/etc out of lightning module
- **Why**: To be able to train LVAE based algorithms in CAREamics
- **How**:  Coded with my elbows, unfortunately

### Changes Made

 **Added**:  
- Some amount of very questionable quality code which needs to be refactored right away
- Very basic tests which just ensure loss is not nan

### Known problems

- Noise models doesn't work because even the logic is actually not there yet
- Training logic for NMs isn't clear
- Most of the parameters should be documented, some should be refactored or removed 
- The whole pipeline isn't complete

---

- [ ] Pre-commit passes


src/careamics/careamist.py
```diff
@@ -150,9 +151,24 @@ def __init__(
...
+                == SupportedArchitecture.UNET
+            ):
+                self.model: FCNModule = FCNModule(
+                    algorithm_config=self.cfg.algorithm_config,
+                )
+            elif (
+                self.cfg.algorithm_config.model.architecture
+                == SupportedArchitecture.LVAE
+            ):  # TODO fuckin mypy thinks model "already defined on line 158"
+                self.model: VAEModule = VAEModule(
```
jdeschamps: I believe what you can do about this is to define a type at the top of the module:
```
Module = Union[FCNModule, VAEModule]
```

And use that type on line 158.

src/careamics/careamist.py
```diff
@@ -150,9 +151,24 @@ def __init__(
             self.cfg = source
 
             # instantiate model
-            self.model = CAREamicsModule(
-                algorithm_config=self.cfg.algorithm_config,
-            )
+            if (
```
jdeschamps, line 153: It would probably makes sense to refactor somewhere else the module creation since the same if-else statement repeats below!

src/careamics/careamist.py
```diff
@@ -150,9 +151,24 @@ def __init__(
...
+                    algorithm_config=self.cfg.algorithm_config,
+                )
+            elif (
+                self.cfg.algorithm_config.model.architecture
+                == SupportedArchitecture.LVAE
+            ):  # TODO fuckin mypy thinks model "already defined on line 158"
+                self.model: VAEModule = VAEModule(
+                    algorithm_config=self.cfg.algorithm_config,
+                )
+            else:
```
jdeschamps, line 160: The configuration allows custom models that are neither Unet nor LVAE (`Custom` I believe). This breaks that possibility.

We could potentially consider that they should be `FCNModule` by default. Not sure what it entails.

src/careamics/careamist.py
jdeschamps: But for now I think the `CAREamist` should just not be able to create a `LVAEModule` at all, since it is not possible to train it properly.

src/careamics/config/algorithm_model.py
```diff
@@ -92,21 +94,23 @@ class AlgorithmConfig(BaseModel):
...
-    algorithm: Literal["n2v", "care", "n2n", "custom"]  # defined in SupportedAlgorithm
-    """Name of the algorithm, as defined in SupportedAlgorithm."""
-
-    loss: Literal["n2v", "mae", "mse"]
-    """Loss function to use, as defined in SupportedLoss."""
-
-    model: Union[UNetModel, VAEModel, CustomModel] = Field(discriminator="architecture")
-    """Model architecture to use, defined in SupportedArchitecture."""
+    # defined in SupportedAlgorithm
+    algorithm: Literal["n2v", "care", "n2n", "hdn", "musplit", "denoisplit", "custom"]
```
jdeschamps: The overall configuration should not allow for the algorithms that don't work yet!

src/careamics/config/algorithm_model.py
```diff
@@ -146,11 +150,54 @@ def algorithm_cross_validation(self: Self) -> Self:
...
+        noise_model : GaussianMixtureNoiseModel
+            Noise model.
+
+        Returns
+        -------
+        GaussianMixtureNoiseModel
+            Validated noise model.
+        """
+        if noise_model is not None:
+            try:
```
jdeschamps: This is not clear to me, if `GaussianMixtureNoiseModel` is a Pydantic model, why wouldn't it have a `type` (if `type` is defined in the model that is)?

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,171 @@
...
+from typing_extensions import Self
+
+from .architecture_model import ArchitectureModel
+
+
+class LVAEModel(ArchitectureModel):
+    """LVAE model."""
+
+    model_config = ConfigDict(
+        use_enum_values=True, protected_namespaces=(), validate_assignment=True
```
jdeschamps: Some leftovers from old code probably:
- `use_enum_values` can go since we are not using `Enum` here
- `protected_namespaces` as well since no parameter is called `model_*`

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+    architecture: Literal["LVAE"]
+    input_shape: int = Field(default=64, ge=8, le=1024)
```
jdeschamps, line 17: Can you add documentation for each parameter. A simple for now that we can review with Ashesh.

See the recent changes in the rest of the Pydantic models.

e.g.:
```   python
input_shape: int = Field(default=64, ge=8, le=1024)
"""Dimension of the input tensor along X and Y."""
```

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,171 @@
...
+    multiscale_count: int = Field(default=3)  # TODO clarify
+    # 0 - off, len(z_dims) - 1
+    z_dims: tuple = Field(
+        default=(128, 128, 128, 128), validate_default=True
+    )  # TODO don't use tuple
```
jdeschamps: That's a TODO that should be easy to have in here and not leave hanging. Or is `mypy` complaining everywhere?

src/careamics/config/architectures/lvae_model.py
jdeschamps: if all the `Field` have `validate_default=True`, then it can also be moved to the `ConfigDict` and removed from the `Field`.

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,171 @@
...
+
+    predict_logvar: Literal[None, "global", "pixelwise", "channelwise"] = (
+        None  # TODO Remove ?
+    )
+    enable_noise_model: bool = Field(default=True, validate_default=True)
+    analytical_kl: bool = Field(default=False, validate_default=True)
+
+    @field_validator("encoder_n_filters")
+    @classmethod
+    def validate_encoder_n_filters(cls, encoder_n_filters: int) -> int:
```
jdeschamps: Can you rename the method to `validate_even`? It makes more sense to me that the validator name is about the validation constraints rather than the variable it validate.

At some points (not now), we will have to go through all our validators and see whether we can refactor some outside. We may already be using a `validate_even` somewhere. And maybe check (not now), this type constraints libraries Talley presented.

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,171 @@
...
+            raise ValueError(
+                f"Number of channels for the bottom layer must be even"
+                f" (got {encoder_n_filters})."
+            )
+
+        return encoder_n_filters
+
+    @field_validator("decoder_n_filters")
+    @classmethod
+    def validate_decoder_n_filters(cls, decoder_n_filters: int) -> int:
```
jdeschamps: This can use the previous validator:

``` python
@field_validator("encoder_n_filters", "decoder_n_filters")
```

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,171 @@
...
+        if decoder_n_filters % 2 != 0:
+            raise ValueError(
+                f"Number of channels for the bottom layer must be even"
+                f" (got {decoder_n_filters})."
+            )
+
+        return decoder_n_filters
+
+    @field_validator("z_dims")
+    def validate_z_dims(cls, z_dims: tuple) -> tuple:
```
jdeschamps, line 103: This can be done simply in the `Field` with the parameter `min_length` as far as I remember.

src/careamics/config/likelihood_model.py
```diff
@@ -0,0 +1,45 @@
...
+
+class GaussianLikelihoodModel(BaseModel):
+    """Gaussion likelihood model.
+
+    Parameters
+    ----------
+    BaseModel
+    """
+
+    model_config = ConfigDict(validate_assignment=True, arbitrary_types_allowed=True)
```
jdeschamps: I don't see any arbitrary type, so that parameters can be removed from the `ConfigDict`.

src/careamics/config/likelihood_model.py
```diff
@@ -0,0 +1,45 @@
...
+
+    type: Literal["NMLikelihoodModel"]
+
+    ch_in: int
+    color_channels: int
+    data_mean: Union[dict[str, torch.Tensor], torch.Tensor]
```
jdeschamps, line 47: This will not be serializable and deserializable in this state.

src/careamics/config/algorithm_model.py
```diff
@@ -92,21 +94,23 @@ class AlgorithmConfig(BaseModel):
...
+    )
 
     # Optional fields
     optimizer: OptimizerModel = OptimizerModel()
     """Optimizer to use, defined in SupportedOptimizer."""
 
     lr_scheduler: LrSchedulerModel = LrSchedulerModel()
-    """Learning rate scheduler to use, defined in SupportedScheduler."""
+    # TODO spit algorithmConfig into separate classes because of this ?
+    likelihood: Optional[Union[GaussianLikelihoodModel, NMLikelihoodModel]] = Field(
```
jdeschamps: If we are not using them, then we should not put them yet.

For instance I see that the `Likelihood` have tensors declared, that means we will probably not be able to write the configuration to disk or read it.

src/careamics/config/nm_model.py
```diff
@@ -0,0 +1,20 @@
...
+
+from typing import Any, Literal
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class GaussianMixtureNoiseModel(BaseModel):
+    """Gaussian mixture noise model."""
+
+    model_config = ConfigDict(validate_assignment=True, arbitrary_types_allowed=True)
```
jdeschamps: Here probably `arbitrary_types_allowed` is not necessary either? Although I don't know how Pydantic deals with `Any`.

src/careamics/config/support/supported_algorithms.py
```diff
@@ -15,6 +15,9 @@ class SupportedAlgorithm(str, BaseEnum):
     CARE = "care"
     N2N = "n2n"
     CUSTOM = "custom"
+    HDN = "hdn"
```
jdeschamps: Too early!

src/careamics/lightning/lightning_module.py
```diff
@@ -202,6 +205,200 @@ def configure_optimizers(self) -> Any:
...
+        -------
+        Any
+            Loss value.
+        """
+        x, *aux = batch
+        out = self.model(x)
+        target = aux[0]
+        self.loss_parameters.current_epoch = self.current_epoch
+        self.loss_parameters.inputs = x
+        self.loss_parameters.mask = ~((target == 0).reshape(len(target), -1).all(dim=1))
```
jdeschamps, line 303: We would need some explanations here, because it is not clear what this is doing! Maybe add a TODO for extended comment....

src/careamics/lightning/lightning_module.py
```diff
@@ -202,6 +205,200 @@ def configure_optimizers(self) -> Any:
...
+
+        Returns
+        -------
+        Any
+            Loss value.
+        """
+        x, *aux = batch
+        out = self.model(x)
+        target = aux[0]
+        self.loss_parameters.current_epoch = self.current_epoch
```
jdeschamps, line 301: Could the `loss_parameters` be an object and have a `self.loss_parameters.update(...)` to refactor these steps somewhere else?

If that makes sense, you could leave a TODO here.

src/careamics/losses/loss_factory.py
```diff
@@ -4,20 +4,86 @@
...
 
-# TODO add tests
-# TODO add custom?
 def loss_factory(loss: Union[SupportedLoss, str]) -> Callable:
     """Return loss function.
 
     Parameters
     ----------
-    loss : Union[SupportedLoss, str]
+    loss : Union[SupportedLoss: str]
```
jdeschamps: i don't understand the `:`?

src/careamics/losses/loss_factory.py
```diff
@@ -4,20 +4,86 @@
...
+from torch import Tensor as tensor
+
 from ..config.support import SupportedLoss
-from .losses import mae_loss, mse_loss, n2v_loss
+from .fcn.losses import mae_loss, mse_loss, n2v_loss
+from .lvae.losses import denoisplit_loss, musplit_loss
+
+
+@dataclass
+class FCNLossParameters:
```
jdeschamps, line 18: We would need description for each of these parameters.

src/careamics/losses/lvae/loss_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    Takes in the KL with shape (batch size, layers), returns the KL with
+    free bits (for optimization) with shape (layers,), which is the average
+    free-bits KL per layer in the current batch.
+    If batch_average is False (default), the free bits are per layer and
+    per batch element. Otherwise, the free bits are still per layer, but
+    are assigned on average to the whole batch. In both cases, the batch
+    average is returned, so it's simply a matter of doing mean(clamp(KL))
+    or clamp(mean(KL)).
+
+    Args:
```
jdeschamps, line 24: This does not abide to the same docstring style as the rest of the code base!

src/careamics/losses/lvae/loss_utils.py
```diff
@@ -0,0 +1,61 @@
+import torch
+
```
jdeschamps, line 2: Add TODOs for tests

src/careamics/losses/lvae/losses.py
```diff
@@ -0,0 +1,377 @@
+##### REQUIRED Methods for Loss Computation #####
```
jdeschamps, line 1: Module docstring?

src/careamics/losses/lvae/losses.py
```diff
@@ -0,0 +1,377 @@
...
+from typing import TYPE_CHECKING
+
+import numpy as np
+import torch
+
+if TYPE_CHECKING:
+    print("sdfdsfds")
```
jdeschamps: That's an interesting if statement. 😆

src/careamics/losses/lvae/losses.py
```diff
@@ -0,0 +1,377 @@
...
+    likelihood_obj: LikelihoodModule,
+    splitting_mask: torch.Tensor = None,
+    return_predicted_img: bool = False,
+) -> dict[str, torch.Tensor]:
+    """
+    Function to compute the reconstruction loss.
+
+    Parameters
+    ----------
+    reconstruction: torch.Tensor,
```
jdeschamps, line 23: Missing parameters descriptions and return section!

src/careamics/losses/lvae/losses.py
```diff
@@ -0,0 +1,377 @@
...
+    target: torch.Tensor
+    input: torch.Tensor
+    splitting_mask: torch.Tensor = None
+        A boolean tensor that indicates which items to keep for reconstruction loss computation.
+        If `None`, all the elements of the items are considered (i.e., the mask is all `True`).
+    return_predicted_img: bool = False
+    likelihood_obj: LikelihoodModule = None
+    """
+    output = _get_reconstruction_loss_vector(
```
jdeschamps, line 32: Is all this worth commenting on or should it be subject to later refactoring?

src/careamics/models/lvae/likelihoods.py
```diff
@@ -9,6 +9,42 @@
...
+    ----------
+    config: Union[GaussianLikelihoodModel, NMLikelihoodModel]
+        The configuration object for the likelihood module.
+
+    Returns
+    -------
+    nn.Module
+        The likelihood module.
+    """
+    if config.type == "GaussianLikelihoodModel":
```
jdeschamps: it looks much better like this 😆

src/careamics/models/lvae/lvae.py
```diff
@@ -4,74 +4,71 @@
 The current implementation is based on "Interpretable Unsupervised Diversity Denoising and Artefact Removal, Prakash et al."
 """
 
-from typing import Dict, Iterable, List, Tuple, Union
+from collections.abc import Iterable
+from typing import Dict, List, Tuple
 
-import ml_collections
+import ml_collections  # TODO: refactor this out
```
jdeschamps: I don't see it used except in the declaration of `config: ml_collections` and then this `config` is not used (as far as I can tell). So kick it out as we discussed to allow the tests from passing (hopefully)!

src/careamics/models/lvae/noise_models.py
```diff
@@ -1,12 +1,45 @@
...
+    """
+    if model_config:
+        if model_config.type == "GaussianMixtureNoiseModel":
+            return GaussianMixtureNoiseModel(**model_config.model_dump())
+        else:
+            raise NotImplementedError(f"Model {model_config.type} is not implemented")
+    return None
+
 
 class DisentNoiseModel(nn.Module):
```
jdeschamps, line 46: Do we know what this is?

jdeschamps (changes requested):
I think we should stick with the original plan of only using the new algorithms (HDN, denoiSplit) with the `LightningAPI` and not make them available through the `CAREamist AP`.

We are very close to a `0.1.0` release, which means that we might get an influx of users and incorporating code that does not work or might break can create a lot of issues.

If the modules are in the code base, they can be instantiated independently with the `Ligthning API`, that will be more than enough for Ashesh's paper. Once we are confident the quality of the code (integration, testing, documentation) is high enough, then we can make it available through the `CAREamist`.

So in the meantime, I would suggest keeping the configuration clean (removing LVAE, NoiseModels, Likelihood, and the new entries in the supported enums) and adding placeholder `NotImplementedErrors` in `CAREamist`.

For now, `NoiseModel` and `Likelihood` could be passed directly as parameters along the algorithm configuration. We can also discuss alternatives (e.g. allowing additional parameters in the Pydantic model via `ConfigDict`, but that would require a lot of `asserts` in the LVAE related code).

src/careamics/careamist.py
CatEek: But mypy isn't complaining about type ?

src/careamics/careamist.py
CatEek: Absolutely. I put `notimplemented` there for now

src/careamics/careamist.py
CatEek: Let's discuss

src/careamics/config/algorithm_model.py
CatEek: But then how would you expect tests to pass ?

src/careamics/config/algorithm_model.py
CatEek: Not clear to me neither, NMs still need lot's of work

src/careamics/config/architectures/lvae_model.py
CatEek: That's what I asked @ashesh-0 and @federico-carrara to do

src/careamics/config/architectures/lvae_model.py
CatEek: There will be some other constrains, that's why i left it there for now

src/careamics/config/likelihood_model.py
CatEek: Don't know yet

src/careamics/losses/loss_factory.py
CatEek: Same as with other descriptions

src/careamics/losses/lvae/loss_utils.py
CatEek: Same

src/careamics/losses/lvae/losses.py
CatEek: ahahahah

src/careamics/models/lvae/noise_models.py
CatEek: No

melisande-c:
> So in the meantime, I would suggest keeping the configuration clean (removing LVAE, NoiseModels, Likelihood, and the new entries in the supported enums) and adding placeholder `NotImplementedErrors` in `CAREamist`.

I will agree and say let's make the minimum number of modifications possible to the rest of the code until the LVAE pipeline is more complete, maybe not even adding not-implemented-errors to `CAREamist`. It might become clearer how we add the new components into the rest of `CAREamist` and the configuration once we understand how they work. I think it will also make the changes to the code easier to follow.

src/careamics/careamist.py
jdeschamps: One needs to try, I haven't. 😄

src/careamics/config/algorithm_model.py
jdeschamps: The tests should then not be implemented to test those until they are officially supported!

jdeschamps:
It seems there is still a `hdn` hanging around in one of the tests!

src/careamics/careamist.py
```diff
@@ -166,9 +184,15 @@ def __init__(
...
-                )
+                if (
+                    self.cfg.algorithm_config.model.architecture
+                    == SupportedArchitecture.UNET
+                ):
+                    self.model = FCNModule(
+                        algorithm_config=self.cfg.algorithm_config,
+                    )
+                else:
+                    raise NotImplementedError("Architecture not supported.")
```
melisande-c, line 185: Repetition of inner if-else statement can be removed through separation of config initialisation and model initialisation. i.e.

```python
if isinstance(source, Path) and source.is_file():
    if source.suffix == ".yaml" or source.suffix == ".yml":
        self.cfg = load_configuration(source)
        self.model = None
    else:
        self.model, self.cfg = load_pretrained(source)
elif isinstance(source, Configuration):
    self.cfg = source
    self.model = None
else:
    raise ValueError("appropriate message")

# instantiate model
if self.model is None:
    if (
        self.cfg.algorithm_config.model.architecture
        == SupportedArchitecture.UNET
    ):
        self.model = FCNModule(
            algorithm_config=self.cfg.algorithm_config,
        )
    elif (
        self.cfg.algorithm_config.model.architecture
        == SupportedArchitecture.LVAE
    ):  
        raise NotImplementedError(
            "LVAE is currently only supported through lightning api"
        )
    else:
        raise NotImplementedError("Architecture not supported.")
```

^ just an example can probably be improved further. Extracting to smaller testable sub functions might be good too.

src/careamics/config/algorithm_model.py
```diff
@@ -89,24 +89,20 @@ class AlgorithmConfig(BaseModel):
...
-    model: Union[UNetModel, VAEModel, CustomModel] = Field(discriminator="architecture")
-    """Model architecture to use, defined in SupportedArchitecture."""
+    model: Union[UNetModel, CustomModel] = Field(discriminator="architecture")
 
     # Optional fields
     optimizer: OptimizerModel = OptimizerModel()
     """Optimizer to use, defined in SupportedOptimizer."""
 
     lr_scheduler: LrSchedulerModel = LrSchedulerModel()
-    """Learning rate scheduler to use, defined in SupportedScheduler."""
```
melisande-c: Some docstrings have been deleted, probably by mistake.

src/careamics/lightning/lightning_module.py
```diff
@@ -202,6 +205,201 @@ def configure_optimizers(self) -> Any:
         }
 
 
+class VAEModule(L.LightningModule):
```
melisande-c, line 208: Are we assuming LVAE and VAE will be able to share a lightning module? (Just a question for my understanding)

src/careamics/config/likelihood_model.py
```diff
@@ -0,0 +1,45 @@
...
+    data_std: Union[dict[str, torch.Tensor], torch.Tensor]
+    noiseModel: nn.Module
```
melisande-c, line 49: Camel casing of `noiseModel` doesn't adhere to python style.

src/careamics/config/support/supported_architectures.py
```diff
@@ -14,7 +14,7 @@ class SupportedArchitecture(str, BaseEnum):
     """
 
     UNET = "UNet"
-    VAE = "VAE"
+    LVAE = "LVAE"
     CUSTOM = (
-        "Custom"  # TODO all the others tags are small letters, except the architect
+        "custom"  # TODO all the others tags are small letters, except the architect
```
melisande-c: can remove TODO

src/careamics/lightning/lightning_module.py
```diff
@@ -202,6 +205,201 @@ def configure_optimizers(self) -> Any:
...
+        self.loss_parameters.inputs = x
+        self.loss_parameters.mask = ~((target == 0).reshape(len(target), -1).all(dim=1))
+        self.loss_parameters.likelihood = self.likelihood
+        self.loss_parameters.noise_model = self.noise_model
+        loss = self.loss_func(out, target, self.loss_parameters)  # TODO ugly ?
+
+        self.log_dict(loss, on_step=True, on_epoch=True, prog_bar=True, logger=True)
+        return loss
+
+    def validation_step(self, batch: Tensor, batch_idx: Any) -> None:
```
melisande-c, line 311: Seems like a lot has been removed from the validation step in src/careamics/lvae_training/lightning_module.py is everything in there definitely useless?

src/careamics/models/lvae/layers.py
```diff
@@ -117,7 +117,7 @@ def __init__(
                     bias=conv2d_bias,
                 )
                 modules.append(conv)
-                modules.append(nonlin())
+                modules.append(nonlin)
```
melisande-c, line 120: if `nonlin=nn.ReLU` as docstring suggests then this should be `module.append(nonlin())`

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,173 @@
...
+    decoder_dropout: float = Field(default=0.1, ge=0.0, le=0.9)
+    nonlinearity: Literal[
+        "None", "Sigmoid", "Softmax", "Tanh", "ReLU", "LeakyReLU", "ELU"
+    ] = Field(
+        default="ELU",
+    )
+
+    predict_logvar: Literal[None, "global", "pixelwise", "channelwise"] = (
+        None  # TODO Remove ?
+    )
```
melisande-c, line 34: If creation of likelihood models has been extracted to `likelihood_factory` then yes it seems this parameter can be removed from LVAE.

src/careamics/models/lvae/layers.py
melisande-c: Ok nvm, I see the `get_activation` function returns for example `nn.ReLU()`. In which case the docstring should be updated.

src/careamics/models/lvae/lvae.py
```diff
@@ -494,29 +447,28 @@ def create_likelihood_module(self):
...
             )
-        if self.loss_type == LossType.DenoiSplitMuSplit or self.likelihood_NM is None:
+        else:
+            self.likelihood_gm = GaussianLikelihood(
+                self.decoder_n_filters,
+                self.target_ch,
+                predict_logvar=self.predict_logvar,
+                logvar_lowerbound=self.logvar_lowerbound,
+                conv2d_bias=self.topdown_conv2d_bias,
+            )
```
melisande-c, line 466: When is `self.likelihood_gm` used? as far as I can see it doesn't seem to be used by any of the`LVAE` class' methods. If it is only used by external functions maybe it can be removed from the `LVAE` class and passed to those functions explicitly.

tests/config/architectures/test_lvae_model.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def test_instantiation():
+    """Test that LVAEModel can be instantiated."""
+    model_params = {
+        "architecture": "LVAE",
+    }
+
+    # instantiate model
+    LVAEModel(**model_params)
```
melisande-c, line 14: maybe assert attributes are set as expected.

tests/config/architectures/test_lvae_model.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Algorithm json schema
+    schema = LVAEModel.model_json_schema()
+
+    # check that all activations are supported
+    for act in schema["properties"]["nonlinearity"]["enum"]:
+        assert act in activations
+
+
+def test_activation_wrong_values():
```
melisande-c, line 84: maybe rename `test_activation_unknown_value`

tests/config/test_algorithm_model.py
```diff
@@ -56,7 +49,7 @@ def test_model_discriminator(minimum_algorithm_n2v):
     "algorithm, loss, model",
     [
         ("n2v", "n2v", {"architecture": "UNet", "n2v2": False}),
-        ("custom", "mae", {"architecture": "UNet", "n2v2": True}),
+        ("n2n", "mae", {"architecture": "UNet", "n2v2": False}),
```
melisande-c, line 68: why this change in parametrisation?

tests/conftest.py
```diff
@@ -22,7 +22,7 @@ def create_tiff(path: Path, n_files: int):
 
 
 @pytest.fixture
-def minimum_algorithm_custom() -> dict:
+def minimum_algorithm_n2v() -> dict:
```
melisande-c, line 70: The git comparison has got a bit messed up, but has custom algorithm been removed?

melisande-c (commented):
So I still think there are some changes here that are out of scope for the PR (also the current PR title is not descriptive of the changes). I see though that it's pretty hard to make the pydantic model when the `LVAE` class does need some refactoring to extract components that don't necessarily need to be contained within the class. Maybe the splitting of the configuration you mentioned will help.

jdeschamps:
I pushed a fix for the doctests as promised.

tests/conftest.py
jdeschamps: Looking into the code, it seems that the rational for `custom` algorithm was to make sure to escape any assumption on the model parameters: https://github.com/CAREamics/careamics/blob/aed4d82387590159a005f3e8d6e08fc8efad168f/src/careamics/config/configuration_model.py#L219

That means that it should not be removed!

jdeschamps: I added back the `custom` algorithm, and added tests that were missing to test what it was actually used for.

jdeschamps:
Comment https://github.com/CAREamics/careamics/pull/180/files#r1675427899 still needs to be fixed regarding the `custom` algorithms and models.

src/careamics/careamist.py
jdeschamps: Now that the configuration has only `custom` and `UNet` models, these statements should go.

Once the `LVAE` is in the configuration, we will need to reassess how to deal with `custom` models.

src/careamics/lightning/lightning_module.py
jdeschamps: I guess so, VAE will just be a LVAE with one level.

src/careamics/lightning/lightning_module.py
jdeschamps: Please add comments about what is happening in these steps.

jdeschamps (changes requested):

src/careamics/config/algorithm_model.py
```diff
@@ -89,24 +89,29 @@ class AlgorithmConfig(BaseModel):
...
-    algorithm: Literal["n2v", "care", "n2n", "custom"]  # defined in SupportedAlgorithm
-    """Name of the algorithm, as defined in SupportedAlgorithm."""
+    # defined in SupportedAlgorithm
+    algorithm: Literal["n2v", "care", "n2n", "custom"]
+    """Name of the algorithm, as defined in SupportedAlgorithm. Use `custom` for custom
+    model architecture."""
 
     loss: Literal["n2v", "mae", "mse"]
-    """Loss function to use, as defined in SupportedLoss."""
+    """Name of the algorithm, as defined in SupportedAlgorithm."""
```
melisande-c, line 102: Algorithm docs instead of loss docs.

src/careamics/config/algorithm_model.py
```diff
@@ -146,8 +151,10 @@ def algorithm_cross_validation(self: Self) -> Self:
             if self.loss == "n2v":
                 raise ValueError("Supervised algorithms do not support loss `n2v`.")
 
-        if isinstance(self.model, VAEModel):
-            raise ValueError("VAE are currently not implemented.")
+        if (self.algorithm == "custom") != (self.model.architecture == "custom"):
+            raise ValueError(
+                "Algorithm and model architecture must be both `custom` or not."
```
melisande-c, line 156: grammar could be clearer. not what?

src/careamics/careamist.py
CatEek: Those things are gone in the config refac pr

src/careamics/lightning/lightning_module.py
CatEek: I wish I knew, it's taken from the original code as is

src/careamics/lightning/lightning_module.py
CatEek: It wasn't added, not "was removed"

tests/config/architectures/test_lvae_model.py
CatEek: I don't know what that function even doing

src/careamics/lightning/lightning_module.py
melisande-c: What I mean is in the lightning module `LadderVAELight` in this [file](https://github.com/CAREamics/careamics/blob/aed4d82387590159a005f3e8d6e08fc8efad168f/src/careamics/lvae_training/lightning_module.py), there is a lot more going on in the validation step.

tests/config/architectures/test_lvae_model.py
melisande-c: Just testing an error is raised if "nonlinearity" is not a supported activation, no?

src/careamics/config/algorithm_model.py
jdeschamps: I guess it was meant to be "None"!

PR #181, opened 2024-07-12
Feature: Prediction writer callback

melisande-c:
### Description

Complete `PredictionWriterCallback` for writing predictions to disk during lightning predict stage.

- **What**: 
  - `prediction_writer_callback` package in `careamics.lightning.callbacks` to contain all code related to `PredictionWriterCallback`. 
  - Write strategies that follow the `WriteStrategy` protocol in `prediction_writer_callback.write_strategy`.
  - `WriteStrategy` subclasses for concrete implementations: `CacheTiles`, `WriteImages` & placeholder for `WriteTilesZarr`.
  - Tests for the various components.
- **Why**: To be able to save predictions to disk in a way can take advantage of any capabilities of pytorch-lightning, (e.g. parallel computation).
- **How**: Following the strategy pattern. `PredictionWriterCallback` is not responsible for any of the save logic itself; instead, it is injected with a `WriteStrategy` class, each subclass of which handles a different method for saving predictions.

### Changes Made

- **Added**: 
  - `prediction_writer_callback` package in `careamics.lightning.callbacks`
  - `prediction_writer_callback` module in the `prediction_writer_callback` package – for containing the `PredictionWriterCallback` class.
  - `write_strategy` module in the `prediction_writer_callback` package – for containing all the different write strategies.
  - `write_strategy_factory` module in the `prediction_writer_callback` package – for containing functions to create the correct write strategy from convenient parameters.
  - `file_path_utils` module in the `prediction_writer_callback` package
  - Tests for all new objects. Including smoke tests for saving to tiff, tiled and not tiled.
- **Removed**: old `PredictionWriterCallback` and old tests.

Following a strict strategy pattern the `PredictionWriterCallback` class should not be responsible for creating the strategies. However I added a `PredictionWriterCallback.from_write_func_params` classmethod that does create a write strategy; this is for lightning API users. I decided to keep it within the `PredictionWriterCallback` class for clarity.

In the tests I tried to isolate individual functionality of every function, but let me know if I went overboard on the mocks 😅.

EDIT: Also one thing to note; predictions are saved with no tifffile metadata. Maybe they should be saved with `metadata={"axes": "C(Z)YX"}`.

#### Future plans

Now all that remains is to add a `predict_to_disk` function in the `CAREamics` class.

EDIT: using `np.testing.assert_array_equal` instead of `assert np.array_equal` seems to have fixed the randomly failing tests. No idea why.

---

- [x] Pre-commit passes


src/careamics/file_io/write/get_func.py
```diff
@@ -1,14 +1,16 @@
...
-from typing import Protocol, Union
+from typing import Literal, Protocol
 
 from numpy.typing import NDArray
 
 from careamics.config.support import SupportedData
 
 from .tiff import write_tiff
 
+SupportedWriteType = Literal["tiff", "custom"]
```
jdeschamps, line 12: So far, we have used throughout the code a pattern of `enum.Enum` for the `Supported*`. 

I am not far enough in the review to have a comment yet, but do you see a reason to prefer a `Literal` to the `Enum`?

I'd argue that since literal are just annotations, Enum are the way to go to enforce discrete values in the code. They are unfortunately a bit more verbose...

src/careamics/lightning/callbacks/prediction_writer_callback/file_path_utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+from pathlib import Path
+from typing import Union
+
+from careamics.dataset import IterablePredDataset, IterableTiledPredDataset
+
+
+# TODO: move to datasets package ?
+def get_sample_file_path(
+    ds: Union[IterableTiledPredDataset, IterablePredDataset], sample_id: int
```
jdeschamps: A there-is-a-single-speck-of-dust-on-the-book-shelf type of comment, in general I prefer to avoid very short variable names! (which probably I am guilty of often...).

Here we know what `ds` is obviously, but it would still be nicer to read to have `dataset`. 

You can ignore that comment!

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
```diff
@@ -0,0 +1,231 @@
...
+        tiled: bool,
+        write_func: Optional[WriteFunc] = None,
+        write_extension: Optional[str] = None,
+        write_func_kwargs: Optional[dict[str, Any]] = None,
+        dirpath: Union[Path, str] = "predictions",
+    ) -> PredictionWriterCallback:  # TODO: change type hint to self (find out how)
+        """
+        Initialize a `PredictionWriterCallback` from write function parameters.
+
+        This will automatically create a `WriteStrategy` to by passed to the
```
jdeschamps: small typo `by` <-> `be`.

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+    A PyTorch Lightning callback to save predictions.
+
+    Parameters
+    ----------
+    write_strategy : WriteStrategy
+        A strategy for writing predictions.
+    dirpath : Path or str, default="predictions"
+        The path to the directory where prediction outputs will be saved. If
+        `dirpath is not absolute it is assumed to  be relative to current working
```
jdeschamps: missing closing "`" and one too many space in "to  be"

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
```diff
@@ -0,0 +1,231 @@
...
+            If a known `write_type` is selected this argument is ignored. For a custom
+            `write_type` a function to save the data must be passed. See notes below.
+        write_extension : str, optional
+            If a known `write_type` is selected this argument is ignored. For a custom
+            `write_type` an extension to save the data with must be passed.
+        write_func_kwargs : dict of {{str: any}}, optional
+            Additional keyword arguments to be passed to the save function.
+        dirpath : pathlib.Path or str, default="predictions"
+            The path to the directory where prediction outputs will be saved. If
+            `dirpath is not absolute it is assumed to  be relative to current working
```
jdeschamps: missing closing "`"

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
```diff
@@ -0,0 +1,231 @@
...
+            PyTorch Lightning module.
+        stage : str
+            Stage of training e.g. 'predict', 'fit', 'validate'.
+        """
+        super().setup(trainer, pl_module, stage)
+        if stage == "predict":
+            # make prediction output directory
+            if not self.dirpath.is_dir():
+                logger.info("Making prediction output directory.")
+                self.dirpath.mkdir()
```
jdeschamps: Here, if it exists, it will raise an error. 

We should probably use `self.dirpath.mdkir(parents=True, exist_ok=True)`.

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
```diff
@@ -0,0 +1,231 @@
...
+        dataloader_idx : int
+            Dataloader index.
+        """
+        # if writing prediction is turned off
+        if not self.writing_predictions:
+            return
+
+        dls: Union[DataLoader, list[DataLoader]] = trainer.predict_dataloaders
+        dl: DataLoader = dls[dataloader_idx] if isinstance(dls, list) else dls
+        ds: Union[IterablePredDataset, IterableTiledPredDataset] = dl.dataset
```
jdeschamps: Here I would be in favor of using longer variable name (e.g. `dataloader_list`, `dataloader` and `dataset`) for clarity.

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
```diff
@@ -0,0 +1,231 @@
...
+        if not (
+            isinstance(ds, IterablePredDataset)
+            or isinstance(ds, IterableTiledPredDataset)
+        ):
+            # Note: Error will be raised before here from the source type
+            # This is for extra redundancy of errors.
+            raise TypeError(
+                "Prediction dataset has to be `IterableTiledPredDataset` or "
+                "`IterablePredDataset. Cannot be `InMemoryPredDataset because filenames"
```
jdeschamps: Two missing closing "`" and a space after "filenames".

src/careamics/lightning/callbacks/prediction_writer_callback/write_strategy.py
```diff
@@ -0,0 +1,394 @@
...
+        batch_idx : int
+            Batch index.
+        dataloader_idx : int
+            Dataloader index.
+        dirpath : Path
+            Path to directory to save predictions to.
+        """
+        dls: Union[DataLoader, list[DataLoader]] = trainer.predict_dataloaders
+        dl: DataLoader = dls[dataloader_idx] if isinstance(dls, list) else dls
+        ds: IterableTiledPredDataset = dl.dataset
```
jdeschamps: Same opinion about the variables!

src/careamics/lightning/callbacks/prediction_writer_callback/write_strategy.py
```diff
@@ -0,0 +1,394 @@
...
+            file_path = create_write_file_path(
+                dirpath=dirpath,
+                file_path=input_file_path,
+                write_extension=self.write_extension,
+            )
+            self.write_func(
+                file_path=file_path, img=prediction_image[0], **self.write_func_kwargs
+            )
+
+    def _have_last_tile(self) -> bool:
```
jdeschamps: My (non-native) brain would name the method `_has_last_tile`. Not super important, maybe just a matter of taste!

src/careamics/file_io/write/get_func.py
jdeschamps: and if the `Literal` was used for mypy and its value never tested, then maybe an `Enum` is an overkill.

jdeschamps (approved):
Very elegant! I like the writing strategy and the cached writer.

Coincidentally, this feature (with the `CAREamist` function in the next PR) was requested by our colleagues today. ^^

I have not had the bravery to look at the tests, I guess we can discuss those when we will make an audit of the tests.

src/careamics/file_io/write/get_func.py
melisande-c: So in my previous PR we spoke about using `Literal` for user facing functions. It is converted to `SupportedData` at some point before selecting the write function, although supported write types are actually a subset of `SupportedData` because `"array"` is not included.

src/careamics/lightning/callbacks/prediction_writer_callback/file_path_utils.py
melisande-c: Ok sure I will update, I agree definitely in function signatures variables should be more verbose.

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
melisande-c: There is the `if not self.dirpath.is_dir()` condition above, but this is more elegant so I will update.

src/careamics/lightning/callbacks/prediction_writer_callback/prediction_writer_callback.py
melisande-c: So I defined `dls` and `dl` mostly for type hinting, otherwise I would have indexed `trainer.predict_dataloaders` directly; I tend to go for short variable names to indicate that they are temporary or not important, but I suppose I can update the variable names for extra clarity. `ds` is actually used later on so it really should be more verbose.

src/careamics/lightning/callbacks/prediction_writer_callback/write_strategy.py
melisande-c: Yeah "has last tile" probably makes more sense I will update.

src/careamics/file_io/write/get_func.py
melisande-c: I will leave this as a `Literal` for now

jdeschamps: I don't have that good a memory. 😄 

Literals for user facing and Enum in the back end sounds good!

PR #183, opened 2024-07-15
Refactor: Stitch predictions – enforce singleton C axis in `TileInformation`

### LINKED ISSUE #175: Allow singleton channel axis in `TileInformation` [Feature]
**Description**

Allowing a channel axis in `TileInformation`, even if it is singleton, will simplify the conversion of prediction outputs. This is mentioned in #170, where a dirtier solution to converting outputs was implemented.

`TileInformation` currently disallows it's `array_shape` attribute to have singleton dimensions (see code ref below), and this was due to microscopy data almost never having singleton dimensions, as also discussed in https://github.com/CAREamics/careamics/issues/159#issuecomment-2182530827. However, since the `TileInformation` is used internally, after axis reshaping – in which singleton channels will be added, it makes sense to allow singleton channels in the `array_shape`.

https://github.com/CAREamics/careamics/blob/57ceab23dadf568175d424235c21aa8eb4dc4c6a/src/careamics/config/tile_information.py#L37-L41 

**Describe the solution you'd like**

Allow singleton channels in `TileInformation.array_shape`, possibly mandating them to prevent any ambiguity between Z and C axes.

Refactor `stitch_prediction_single` function, so expanding the prediction output axes based on the tile axes is not necessary. (see below).

https://github.com/CAREamics/careamics/blob/57ceab23dadf568175d424235c21aa8eb4dc4c6a/src/careamics/prediction_utils/stitch_prediction.py#L83-L86

melisande-c:
### Description

Allow, and in fact enforce, singleton channel dimensions in `TileInformation.array_shape` attribute, to allow stitching predictions to make more sense.

- **What**: Have array shape stored in `TileInformation` match what the prediction output shape will be (has singleton dimensions). Updated tests accordingly.
- **Why**: `TileInformation` is not a user facing class so enforcing no singleton channels wasn't necessary, especially since it is used after the data has been reshaped to have SC(Z)YX dimensions. This simplifies the stitching logic by removing the need to reshape the output array dimensions based on the tile dimensions. Reduces ambiguity; we know that the dimension are CYX or CZYX if there are 3 or 4 channels respectively.
- **How**: Enforce `TileInformation.array_shape` has 3 or 4 dimensions. Don't reshape prediction output.

### Changes Made

- **Added**: 
  - `n_dims` Pydantic validator in `TileInformation` class, to enforce number of dimensions is 3 or 4.
  - Test to make sure error is raised when the dimension is not 3 or 4.
- **Modified**:
  - `extract_tiles` function: does not squeeze `array_shape` in `TileInformation` initialisation.
  - `stitch_prediction_single` function does not reshape prediction shape.
  - Tests for stitching prediction have to index the input array so that they keep the S dim. (This might have been a bug before).
- **Removed**: 
  - `TileInformation.no_singleton_dimension` Pydantic validator.
  - Test `test_error_no_singleton_dimension`

### Related Issues

- Resolves #175 

### Breaking changes

Initialising `TileInformation` with only YX dimensions.

We might still want to enforce an error if axes Z, Y or X are not singleton.

---

- [x] Pre-commit passes


src/careamics/config/tile_information.py
```diff
@@ -34,11 +33,15 @@ class TileInformation(BaseModel):
     sample_id: int
     """Sample ID of the tile."""
 
+    # TODO: Test that ZYX axes are not singleton ?
+
     @field_validator("array_shape")
     @classmethod
-    def no_singleton_dimensions(cls, v: tuple[int, ...]):
+    def n_dims(cls, v: tuple[int, ...]):
```
jdeschamps: In general, I find it better to make the name of the Pydantic validators very explicit about what they do `ndims_equal_3_or_4` rather than `ndims`, but it does not matter much here since this is an internal Pydantic validator (as in it is unlikely that anyone will have to look in there).

One alternative to this validator is actually to use [annotated-types](https://github.com/annotated-types/annotated-types?tab=readme-ov-file#minlen-maxlen-len) (which Talley showcased last time):

```python
from pydantic import BaseModel
from typing import Annotated
from annotated_types import MinLen, MaxLen

DimTuple = Annotated[tuple, MinLen(3), MaxLen(4)]

class TupleHolder(BaseModel):

    my_tuple: DimTuple

TupleHolder(my_tuple=(1, 2, 3, 4, 5))
```

We might want to generalize using these annotated types. We could (when useful) move some annotated types to a submodule if they are reused throughout the configuration.

jdeschamps (approved):
Looks good! Just made one comment about the pydantic field validator.

src/careamics/config/tile_information.py
melisande-c: Thanks for the tips! I will update it to the annotated type.

PR #184, opened 2024-07-16
Feat: Allow passing folder in BMZ

### LINKED ISSUE #182: export_to_bmz [BUG]
**Describe the bug**
I tried the simple n2v example "Basic CAREamics usage" from [https://careamics.github.io/0.1/guides/careamist_api/]. This worked just as expected. Then, I wanted to save the generated model using the "export_to_bmz" function and ended up with lots of error messages.

**To Reproduce**
Simply run the code from "Basic CAREamics usage" and add 

`
careamist.export_to_bmz(
    path="n2v_models",
    name="n2v_model_example.bmz",
    input_array=train_data,
    authors=[{
        "name": "nobody",
        "email": "nobody@nobody",
    }]
)
`

**Expected behavior**
I was expecting to get a zip (or bmz) file containing model for further use.

**Environment:**
 - OS: Windows10
 - careamics                 0.1.0rc8
 - careamics-portfolio       0.0.12
 
Following instructions from [https://careamics.github.io/0.1/installation/]:
mamba create -n careamics python=3.10
mamba activate careamics
mamba install pytorch torchvision pytorch-cuda=11.8 -c pytorch -c nvidia

jdeschamps:
### Description

The parameters of `CAREamist.export_bmz` are counter-intuitive: the `path` parameter is supposed to be the file path to the exported archive, and `name` is the name of the model in the Bioimage Model Zoo. If the parents of `path` do not exist, an error will be raised.

Two ways from there:
- Make `name` the name of the archive and enforce `path` as a folder.
- Keep consistency with the BMZ specs, and cover cases where users might mix up the meaning of both parameters.

I decided for the second one, a bit more tortuous.

Following https://github.com/CAREamics/careamics/issues/182, this PR allow passing a path to a folder to the `CAREamist.export_bmz` method.

Three new cases are now covered:
- If `path` has parent folder that do not exist, create the parents
- If `path` points to a folder, use `name` to create a file name with the correct extension, and no parentheses or spaces.
- If `path` points to a file with the wrong suffix, replace it by `.zip`

This should cover common use cases of users.

- **What**: Create parent folders of the `path` in `CAREamist.export_bmz`, if the `path` is a folder, use `name` as name of the archive.
- **Why**: The previous use of `path` and `name` could be confusing.
- **How**: Added path and string manipulation to `careamics.model_io.bmz_io.export_bmz`.

### Changes Made

- **Modified**: `careamics.model_io.bmz_io.export_bmz`.

### Related Issues

- Fixes #182 

All the path and string manipulation could be refactored outside of this function. In general, these functions could benefit from some refactoring.

---

- [x] Pre-commit passes


melisande-c:
Could have this been solved with better documentation or naming of the arguments? I find variables changing behaviour and definitions maybe more confusing? Especially since, the condition for the change in behaviour is basically if `path` has a suffix or not, and I think the big block of additional docs to explain this shows that it's maybe not very intuitive.

Maybe if the argument `path` is changed to `file_path` or `archive_path` it would be clearer what the argument is. If someone really wants to point to a directory maybe there can be an additional optional `dir_path` argument? so the final save path would be `dir_path / file_path`.

Edit: Also the docs for the `name` variable could be more explicit in explaining that it is BMZ metadata.

jdeschamps:
I guess it comes down to the question of whether we privilege clearer (more maintainable) code or defensive code to preemptively cover user errors (because in most cases people will not read the doc and will just go with the intuition they have).

Could the parameter naming be better? Sure! I like `path_to_archive`. I am more worried that `model_name` for instance still rings like it is the name of the "exported model".

Here is a new commit, I tried to go the cleaner code way with better parameter names:
- `path_to_archive`
- `friendly_model_name` which is a bit long...

We could go with `model_name`. What do you think?

melisande-c:
> I guess it comes down to the question of whether we privilege clearer (more maintainable) code or defensive code to preemptively cover user errors (because in most cases people will not read the doc and will just go with the intuition they have).
> 
> Could the parameter naming be better? Sure! I like `path_to_archive`. I am more worried that `model_name` for instance still rings like it is the name of the "exported model".
> 
> Here is a new commit, I tried to go the cleaner code way with better parameter names:
> 
> * `path_to_archive`
> * `friendly_model_name` which is a bit long...
> 
> We could go with `model_name`. What do you think?

I think this is good as it is, I agree people will go with their intuition but hopefully misinterpretation can be reduced with clear parameter names. Plus the previous implementation could also lead to user confusion if they expected the ".zip" extension to be added for them. 

I think my main issue with the previous implementation was the multiple cases. Pick one behaviour or the other. I prefer this current implementation but I think something like the following would also be fine.

```python
    def export_to_bmz(
        self,
        dir_path: Union[Path, str],
        model_name: str,
        input_array: NDArray,
        authors: list[dict],
        general_description: str = "",
        channel_names: Optional[list[str]] = None,
        data_description: Optional[str] = None,
    ) -> None:
        """Export the model to the BioImage Model Zoo format."""
```
 where the archive path becomes:
`dir_path/formatted_model_name.zip`
and the bmz name is the also the model name.

melisande-c (approved):

PR #187, opened 2024-07-19
Algorithm split

CatEek:
Split the algorithm pydantic model into FCN and VAE

Why: N2V family configs aren't affected by the process of adding new algos

Problems:
- some tests don't pass, all related to VAE losses, nm etc.
- VAE config is basic, needs validators


src/careamics/lightning/lightning_module.py
```diff
@@ -245,16 +245,20 @@ def __init__(self, algorithm_config: Union[FCNAlgorithmConfig, dict]) -> None:
...
             if isinstance(algorithm_config, dict)
             else algorithm_config
         )
 
         # create model and loss function
         self.model: nn.Module = model_factory(self.algorithm_config.model)
-        self.likelihood = likelihood_factory(self.algorithm_config.likelihood)
         self.noise_model = noise_model_factory(self.algorithm_config.noise_model)
-        self.loss_parameters = LVAELossParameters
+        self.noise_model_likelihood = likelihood_factory(
```
CatEek, line 257: @jdeschamps @veegalinova @melisande-c Take a look at this logic. Likelihoods aren't mutually exclusive and NMLL will take a noisemodel as input. 

Should NM be inside Pydantic then ?

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,173 @@
...
+    decoder_dropout: float = Field(default=0.1, ge=0.0, le=0.9)
+    nonlinearity: Literal[
+        "None", "Sigmoid", "Softmax", "Tanh", "ReLU", "LeakyReLU", "ELU"
+    ] = Field(
+        default="ELU",
+    )
+
+    predict_logvar: Literal[None, "global", "pixelwise", "channelwise"] = (
+        None  # TODO can only be none or pixelwise
+    )
```
melisande-c: Can update, since meeting with Ashesh we know for MuSplit we need "pixel_wise" and for DenoiSplit & HDN it is None.

src/careamics/config/nm_model.py
```diff
@@ -0,0 +1,22 @@
+"""Noise models config."""
+
+from typing import Any, Literal
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class GMNMModel(BaseModel):
+    """Gaussian mixture noise model."""
```
melisande-c: This name could be more explicit. Maybe `GausianMixtureNMModel`, since I think I've seen the abbreviation NM elsewhere in the code for noise model.

src/careamics/careamist.py
```diff
@@ -150,9 +150,27 @@ def __init__(
...
+                # self.model: VAEModule = VAEModule(
+                #     algorithm_config=self.cfg.algorithm_config,
+                # )
+                raise NotImplementedError(
+                    "LVAE is currently only supported through lightning api"
+                )
+            else:
+                raise NotImplementedError(
+                    "Architecture not supported. Please use UNet or LVAE."
+                )
```
melisande-c: Since there is a now split between FCN & VAE at the algorithm level, I think there is a way to refactor the lightning module switching somewhere else.

src/careamics/config/support/supported_architectures.py
```diff
@@ -14,7 +14,7 @@ class SupportedArchitecture(str, BaseEnum):
     """
 
     UNET = "UNet"
-    VAE = "VAE"
+    LVAE = "LVAE"
     CUSTOM = (
-        "Custom"  # TODO all the others tags are small letters, except the architect
+        "custom"  # TODO all the others tags are small letters, except the architect
```
melisande-c: Remove TODO

src/careamics/config/vae_algorithm_model.py
```diff
@@ -0,0 +1,147 @@
...
+    Minimum example:
+    >>> from careamics.config import AlgorithmConfig
+    >>> config_dict = {
+    ...     "algorithm": "n2v",
+    ...     "loss": "n2v",
+    ...     "model": {
+    ...         "architecture": "UNet",
+    ...     }
+    ... }
+    >>> config = AlgorithmConfig(**config_dict)
```
melisande-c: Remove FCNAlgorithm examples

src/careamics/config/fcn_algorithm_model.py
```diff
@@ -1,18 +1,14 @@
...
-from .optimizer_models import LrSchedulerModel, OptimizerModel
+from careamics.config.architectures import CustomModel, UNetModel
+from careamics.config.optimizer_models import LrSchedulerModel, OptimizerModel
 
 
-class AlgorithmConfig(BaseModel):
+class FCNAlgorithmConfig(BaseModel):
     """Algorithm configuration.
 
     This Pydantic model validates the parameters governing the components of the
```
melisande-c, line 16: Examples in docstring below need updating due to class name change.

src/careamics/config/vae_algorithm_model.py
```diff
@@ -0,0 +1,147 @@
...
+        protected_namespaces=(),  # allows to use model_* as a field name
+        validate_assignment=True,
+        extra="allow",
+    )
+
+    # Mandatory fields
+    # defined in SupportedAlgorithm
+    algorithm_type: Literal["vae"]
+    algorithm: Literal["musplit", "denoisplit", "custom"]
+    loss: Literal["musplit_loss", "denoisplit_loss"]
```
melisande-c: Other losses (e.g. `n2v`) do not have `_loss` suffix. We should stay consistent.

src/careamics/config/vae_algorithm_model.py
```diff
@@ -0,0 +1,147 @@
...
+                raise ValueError(
+                    f"Algorithm {self.algorithm} only supports loss `musplit`."
+                )
+        # TODO add more checks
+
+        if self.algorithm == "denoisplit":
+            if self.loss == "denoisplit_loss":
+                raise ValueError(
+                    f"Algorithm {self.algorithm} only supports loss `denoisplit_loss`."
+                )
```
melisande-c: If `_loss` suffix is removed from loss names, can do
```python
if self.algorithm != self.loss
```

src/careamics/lightning/lightning_module.py
melisande-c: Can you clarify? It already is in the `VAEAlgorithmConfig` & `NMLikelihoodModel` Pydantic models. Do you mean remove the attribute from here and only access it through `noise_model_likelihood` by e.g. `noise_model_likelihood.noise_model`?

src/careamics/lightning/lightning_module.py
```diff
@@ -202,7 +204,207 @@ def configure_optimizers(self) -> Any:
...
+        x, *aux = batch
+        out = self.model(x)
+        target = aux[0]
+        # TODO rethink loss parameters
+        self.loss_parameters.current_epoch = self.current_epoch
+        self.loss_parameters.inputs = x
+        self.loss_parameters.mask = ~((target == 0).reshape(len(target), -1).all(dim=1))
+        self.loss_parameters.likelihood = self.likelihood  # TODO refac ?
+        self.loss_parameters.noise_model = self.noise_model
+        loss = self.loss_func(out, target, self.loss_parameters)  # TODO ugly ?
```
melisande-c: No attribute `loss_parameters` defined in `__init__`.

src/careamics/lightning/lightning_module.py
```diff
@@ -202,7 +204,207 @@ def configure_optimizers(self) -> Any:
...
+        val_loss = self.loss_func(out, *aux)
+
+        # log validation loss
+        self.log_dict(
+            val_loss,
+            on_step=False,
+            on_epoch=True,
+            prog_bar=True,
+            logger=True,
+        )
```
melisande-c, line 338: raise `NotImplemented` if this is not the actual validation step logic.

src/careamics/lightning/lightning_module.py
```diff
@@ -202,7 +204,207 @@ def configure_optimizers(self) -> Any:
...
+        Returns
+        -------
+        Any
+            Model output.
+        """
+        if self._trainer.datamodule.tiled:
+            x, *aux = batch
+        else:
+            x = batch
+            aux = []
```
melisande-c, line 359: raise `NotImplemented`

src/careamics/losses/lvae/losses.py
```diff
@@ -0,0 +1,374 @@
...
+    return kl_loss
+
+
+# Loss Computations
+# mask = torch.isnan(target.reshape(len(x), -1)).all(dim=1)
+
+
+def musplit_loss(model_outputs, targets, loss_parameters) -> dict:
+    """Loss function for MuSplit.
+
```
melisande-c, line 220: So I do think it would be better if all the parameters are passed explicitly here. Can you help me understand what problem the loss parameters data class is solving? When is it initialised and when are the parameters initially set? Are they in the configuration?

src/careamics/models/lvae/noise_models.py
```diff
@@ -1,71 +1,52 @@
...
+    ------
+    NotImplementedError
+        _description_
+    """
+    if model_config and paths:
+        if model_config.model_type == "GaussianMixtureNoiseModel":
+            noise_models = []
+            for path in paths:
+                model_config.path = path
+                noise_models.append(GaussianMixtureNoiseModel(model_config))
```
melisande-c, line 43: This is confusing. There is a chance that each instance of `GaussianMixtureNoiseModel` will have a reference to `model_config.path` so when it updates during the loop it will update in all the instances. Even if python does handle this by copying at assignment, it still does not seem like the best practice.

src/careamics/models/lvae/noise_models.py
```diff
@@ -134,9 +116,69 @@ def get_noise_model(
     return None
 
 
+class DisentNoiseModel(nn.Module):
```
melisande-c, line 120: Would be good to get some docs/explanation for this class @ashesh-0

src/careamics/models/lvae/noise_models.py
```diff
@@ -168,19 +210,20 @@ class GaussianMixtureNoiseModel(nn.Module):
...
         super().__init__()
         self._learnable = False
 
-        if kwargs.get("params") is None:
-            weight = kwargs.get("weight")
-            n_gaussian = kwargs.get("n_gaussian")
-            n_coeff = kwargs.get("n_coeff")
-            min_signal = kwargs.get("min_signal")
-            max_signal = kwargs.get("max_signal")
+        if config.path is None:
```
melisande-c, line 219: Not a fan of this pattern. Maybe there should be an option to initialise the class with a path or a config object. If initialised with a config then just store as an instance attribute. If initialised with a path then a load the saved parameters and create the config object from these parameters to store as an attribute. (And path is removed as an attribute from the config). This would also solve my previous comment since the class can be initialised directly from the path.

src/careamics/models/lvae/noise_models.py
```diff
@@ -307,7 +348,7 @@ def likelihood(self, observations, signals):
         return p + self.tol
 
     def getGaussianParameters(self, signals):
```
melisande-c, line 351: Update camel casing to snake casing for functions.

melisande-c (commented):
I haven't looked through the tests yet. Added some comments, I tried not to duplicate comments from the previous PRs (LVAE losses tests).

src/careamics/models/lvae/noise_models.py
CatEek: Let's do it in a separate PR with other formatting. There's too much of that

src/careamics/models/lvae/noise_models.py
CatEek: @ashesh-0 Can NMs for different channels be configured differently ?

src/careamics/careamist.py
CatEek: Good point

src/careamics/models/lvae/noise_models.py
CatEek: Especially likelihood method

src/careamics/models/lvae/noise_models.py
ashesh-0: Currently, they are not. The only change in the config is the path.

ashesh-0: But, certainly, one could handle it in a better way.

src/careamics/models/lvae/noise_models.py
ashesh-0: The purpose of Noise model is to give the likelihood tensor of same dimensions as the prediction. 
For splitting task, we have multiple target channels,and therefore we need multiple noise models. This class is responsible for managing all noise model related stuff. 
self.nmodel_0 is the noise model object for the first class, self.nmodel_1 for the second class and so on. 
likelihood function in this case should return a list of likelihood tensors, one for each channel and using the specific noise_model. 

For HDN, there will be just one noise model and so is returns, for historical reasons, not a list but a single likelihood tensor. `obs.shape[1] ==1`

ashesh-0: As for the dimensions of the input arguments of the likelihood function, the only constraint is that they should be of the same shape.  For 2D, the shape is `BxCxHxW`, (batch, channel count, height, width). For 3D data, the dimensions will be `BxCxZxHxW`.

src/careamics/config/__init__.py
```diff
@@ -1,8 +1,8 @@
 """Configuration module."""
 
 __all__ = [
-    "AlgorithmConfig",
-    "DataConfig",
+    "FCNAlgorithmConfig",
+    "VAEAlgorithmConfig" "DataConfig",
```
melisande-c: Missing comma and newline

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class LVAEModel(ArchitectureModel):
+    """LVAE model."""
+
+    model_config = ConfigDict(validate_assignment=True, validate_default=True)
+
+    architecture: Literal["LVAE"]
+    input_shape: int = Field(default=64, ge=8, le=1024)
+    multiscale_count: int = Field(default=5)  # TODO clarify
```
melisande-c, line 18: `multiscale_count` could be,`n_lateral_context_inputs`? Just a suggestion, not that important so feel free to ignore

src/careamics/config/architectures/lvae_model.py
```diff
@@ -0,0 +1,172 @@
...
+    decoder_n_filters: int = Field(default=64, ge=8, le=1024)
+    encoder_dropout: float = Field(default=0.1, ge=0.0, le=0.9)
+    decoder_dropout: float = Field(default=0.1, ge=0.0, le=0.9)
+    nonlinearity: Literal[
+        "None", "Sigmoid", "Softmax", "Tanh", "ReLU", "LeakyReLU", "ELU"
+    ] = Field(
+        default="ELU",
+    )
+
+    predict_logvar: Literal[None, "pixelwise"] = None
```
melisande-c, line 32: From our meeting with Ashesh, `predict_logvar=None` for DenoiSplit and `predict_logvar="pixelwise"` for musplit. Maybe we need a pydantic validator at the algorithm config level?

src/careamics/losses/loss_factory.py
```diff
@@ -4,14 +4,81 @@
...
+    prediction: tensor
+    targets: tensor
+    mask: tensor
+    current_epoch: int
+    loss_weight: float
+
+
+@dataclass
+class LVAELossParameters:
+    """Dataclass for LVAE loss."""
```
melisande-c, line 31: So what is the point of the loss parameter dataclass? why can the functions not just have these arguments, with defaults where applicable? Sorry if I'm missing something

src/careamics/losses/lvae/losses.py
```diff
@@ -0,0 +1,328 @@
...
+    ----------
+    reconstruction: torch.Tensor,
+    target: torch.Tensor
+    input: torch.Tensor
+    splitting_mask: torch.Tensor = None
+        A boolean tensor that indicates which items to keep for reconstruction loss computation.
+        If `None`, all the elements of the items are considered (i.e., the mask is all `True`).
+    return_predicted_img: bool = False
+    likelihood_obj: LikelihoodModule = None
+    """
```
melisande-c: Can we add a Returns section to the doc. Ideally with the dictionary keys and what they mean.

src/careamics/losses/loss_factory.py
CatEek: I guess we'll figure this out in a new pr with a functioning training step

src/careamics/losses/lvae/losses.py
CatEek: @ashesh-0 Could you please take a look ?

src/careamics/config/vae_algorithm_model.py
melisande-c: @CatEek

src/careamics/losses/lvae/losses.py
ashesh-0: Returns a dictionary with the following keys:
'loss': it is the reconstruction loss, ie, likelihood loss
'ch1_loss', 'ch2_loss',... 'ch`i`_loss': likelihood loss for the `i`th channel. 

'loss' is what is needed. Other things are not required.

ashesh-0: Also, `loss` is scalar.

ashesh-0: splitting_mask is always None.

jdeschamps:
@CatEek the docstring tests are not passing, could you fix the docstrings?

src/careamics/config/support/supported_losses.py
```diff
@@ -17,11 +17,12 @@ class SupportedLoss(str, BaseEnum):
...
-    MSE = "mse"
-    MAE = "mae"
-    N2V = "n2v"
+    MSE = "mse_loss"
+    MAE = "mae_loss"
+    N2V = "n2v_loss"
     # PN2V = "pn2v"
     # HDN = "hdn"
+    MUSPLIT = "musplit_loss"
+    DENOISPLIT = "denoisplit_loss"
```
melisande-c: tbh I meant remove the `"_loss"` suffix from `musplit` and `denoisplit` but maybe I should have been more explicit. It is better to conform with the existing implementation in cases like this because it can break user code. Additionally, we know that these are losses because they are in the `SupportedLoss` class, so we do not need additional context here...

src/careamics/config/vae_algorithm_model.py
```diff
@@ -0,0 +1,118 @@
...
+                raise ValueError(
+                    "Algorithm `musplit` only supports `predict_logvar` as `pixelwise`."
+                )
+        # TODO add more checks
+
+        if self.algorithm == "denoisplit":
+            if self.loss == "denoisplit_loss":
+                raise ValueError(
+                    f"Algorithm {self.algorithm} only supports loss `denoisplit_loss`."
+                )
```
melisande-c: Also we should probably be comparing with the `Enum` class values here rather than strings. It might have saved some trouble. `SupportedLoss.MUSPLIT=="musplit"` will return `True`, there is no need to use `.value`, I think because `SupportedLoss` also inherits from string

src/careamics/config/vae_algorithm_model.py
```diff
@@ -0,0 +1,118 @@
...
+        protected_namespaces=(),  # allows to use model_* as a field name
+        validate_assignment=True,
+        extra="allow",
+    )
+
+    # Mandatory fields
+    # defined in SupportedAlgorithm
+    algorithm_type: Literal["vae"]
+    algorithm: Literal["musplit", "denoisplit", "custom"]
+    loss: Literal["musplit_loss", "denoisplit_loss"]
```
melisande-c: In fact, maybe we should just type hint with the `Enum` classes instead of `Literal`. I've had some discussion with @jdeschamps about type hinting with `Literal` on user facing code and  using `Enum` internally, but maybe not using `Literal` will save us some trouble. 

Also, I just checked and the following code works, so users would be able to initialise the class with strings anyway.
```python
from pydantic import BaseModel, ConfigDict
from careamics.utils import BaseEnum

class Colour(str, BaseEnum):
    RED: str = "red"
    GREEN: str = "green"
    BLUE: str = "blue"

class Paint(BaseModel):
    # Pydantic class configuration
    model_config = ConfigDict(
        protected_namespaces=(),  # allows to use model_* as a field name
        validate_assignment=True,
        extra="allow",
    )

    colour: Colour

paint_instance = Paint(colour="green")
```

In this case though, there would have to be additional validation that the loss is only `musplit`, `denoisplit` or `custom` and not `mae` etc.

melisande-c:
There are still some issues, numpy-doc validation fails for a lot of the LVAE code but we will address this in the future.

Currently the most major issue is that the Pydantic model `VAEAlgorithmConfig` is not serializable. See newly added but skipped test.

https://github.com/CAREamics/careamics/blob/a378828e7dc3aa601151e1b323a2362a9e5ae587/tests/config/test_vae_algorithm_model.py#L7-L15

This must also be addressed in the very near future.

melisande-c (approved):

PR #189, opened 2024-07-23
Feature: add CAREamist predict to disk method

melisande-c:
### Description

- **What**: Add a `predict_to_disk` method to the `CAREamist` class. It uses the `PredictionWriterCallback`; an instance of which has been added as an attribute of `CAREamist`.
- **Why**: Allows users of the `CAREamist` class to write predictions to disk.
- **How**: 
  - Add an instance of `PredictionWriterCallback` as an attribute of `CAREamist` and append it to the list of callbacks.
  - The bulk of `CAREamist.predict` has been moved to `CAREamist._predict_implementation`; for reuse between `predict` and `predict_to_disk` but allow the parameter `return_predictions` to be set to `True` or `False` respectively.
  - `PredictionWriterCallback` has a context manager for toggling writing predictions on and off (added for aesthetic reasons), but it might also be useful for the lightning API users.
  - The file names for the predictions are taken from the dataset classes so they match the input files.

#### Getting the file names

This is what has caused the most trouble. I have come up with a solution that works, but I dislike it because it introduces a lot of coupling between the `WriteStrategy` classes and the dataset classes. Additionally, I can envision situations where it breaks down.

Currently, I have modified `IterablePredDataset` and `IterableTiledPredDataset` so that they store a list to keep track of the file to which a sample belongs. Then, in the `WriteStrategy`, if the index of the sample is known the file name can be retrieved from the dataset. The concrete implementations are as follows:
- In the `WriteImage` strategy the index of the sample can be calculated from the batch index and the batch size. 
- In the `CacheTiles` strategy the sample index is kept track of with an internal attribute `sample_count`; this is bad because there is no reset mechanism and if the same instance of the `CacheTiles` object is used again it will fail.

*Notes:* 
- In the datasets I tried having just a `current_file_index` attribute, but the way lightning pre-fetches batches meant it didn't work.

*Alternatives:*
- The datasets return the filename that the sample/tile belongs to. This will take some extra work because there are some places where a batch length of 2 is used to infer whether a prediction is tiled or not, which would obviously not hold anymore.

### Changes Made

- **Added**: 
  - `CAREamist.predict_to_disk` method.
  - `CAREamist._predict_implementation` method.
  - `CAREamist.prediction_writer` attribute.
  - `validate_unet_tile_size` function (in `prediction_utils` package).
  - `IterablePredDataset.sample_file_indices` & `IterableTiledPredDataset.sample_file_indices` attributes.
  - Tests
- **Modified**: 
  - `PredictionWriterCallback` class.
  - `CacheTiles` class.
  - `WriteImage` class.
  - `get_sample_file_path` function (in `lightning.callbacks.prediction_writer_callback.file_utils` module).
  - `IterablePredDataset.__iter__` & `IterableTiledPredDataset.__iter__` methods.

### Breaking changes

Shouldn't alter the behaviour of the rest of the code base.

Currently if there is more than one sample in a file, say 2, so it has the shape (2, 32, 32) with the axes being `"SYX"` and the file name "image.tiff". Then, there will be a separate file saved for each sample prediction called "image_0.tiff" and "image_1.tiff". It is implemented this way because it was the path of least resistance, but if you think this will be too inconvenient or confusing to users then I can change it with a little more work by altering the `WriteStrategy` classes.

Also, I realised I am missing a couple tests, I will add them shortly.

###

*Need opinions on:*
- A better way to get/generate the file names for the predictions. Should they be returned with the batches by the dataset classes? Or just leave it as it is for now even though it introduces a lot of coupling.
- Should I update the implementation so the saved file structure mirrors the input data (instead of 1 file per sample).

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -24,6 +24,7 @@ from careamics.config.support import (
     SupportedLogger,
 )
 from careamics.dataset.dataset_utils import reshape_array
+from careamics.file_io import WriteFunc
 from careamics.lightning import (
     CAREamicsModule,
     HyperParametersCallback,
@@ -674,6 +675,65 @@ class CAREamist:
             model=self.model, datamodule=self.pred_datamodule
         )
         return convert_outputs(predictions, self.pred_datamodule.tiled)
+
+    def predict_to_disk(
+        self,
+        source: Union[PredictDataModule, Path, str],
+        *,
+        batch_size: int = 1,
+        tile_size: Optional[tuple[int, ...]] = None,
+        tile_overlap: tuple[int, ...] = (48, 48),
+        axes: Optional[str] = None,
+        data_type: Optional[Literal["array", "tiff", "custom"]] = None,
+        tta_transforms: bool = True,
+        dataloader_params: Optional[dict] = None,
+        read_source_func: Optional[Callable] = None,
+        extension_filter: str = "",
+        write_type: Literal["tiff", "custom"] = "tiff",
+        write_extension: Optional[str] = None,
+        write_func: Optional[WriteFunc] = None,
+        write_func_kwargs: Optional[dict[str, Any]] = None,
+        predict_dir: Optional[Path] = None,
+    ) -> None:
+        # TODO: raise error for predict dataset not iterable
+
+        # assign predict dir
+        # TODO: absolute vs relative
+        if predict_dir is not None:
+            self.prediction_writer.dirpath = predict_dir
+
+        tiled = tile_size is not None
+        write_strategy = create_write_strategy(
+            write_type=write_type,
+            tiled=tiled,
+            write_func=write_func,
+            write_extension=write_extension,
+            write_func_kwargs=write_func_kwargs,
+        )
+        self.prediction_writer.write_strategy = write_strategy
+
+        # TODO: validate tile size (extract to seperate function)
+
+        # create the prediction
+        self.pred_datamodule = create_predict_datamodule(
+            pred_data=source,
+            data_type=data_type or self.cfg.data_config.data_type,
+            axes=axes or self.cfg.data_config.axes,
+            image_means=self.cfg.data_config.image_means,
+            image_stds=self.cfg.data_config.image_stds,
+            tile_size=tile_size,
+            tile_overlap=tile_overlap,
+            batch_size=batch_size or self.cfg.data_config.batch_size,
+            tta_transforms=tta_transforms,
+            read_source_func=read_source_func,
+            extension_filter=extension_filter,
+            dataloader_params=dataloader_params,
+        )
+
+        # predict (without returning predictions, saves memory)
+        self.trainer.predict(
+            model=self.model, datamodule=self.pred_datamodule, return_predictions=False
+        )
 
     def export_to_bmz(
         self,
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -531,7 +531,7 @@ class CAREamist:
         *,
         batch_size: int = 1,
         tile_size: Optional[tuple[int, ...]] = None,
-        tile_overlap: tuple[int, ...] = (48, 48),
+        tile_overlap: Optional[tuple[int, ...]] = (48, 48),
         axes: Optional[str] = None,
         data_type: Optional[Literal["tiff", "custom"]] = None,
         tta_transforms: bool = True,
@@ -547,7 +547,7 @@ class CAREamist:
         *,
         batch_size: int = 1,
         tile_size: Optional[tuple[int, ...]] = None,
-        tile_overlap: tuple[int, ...] = (48, 48),
+        tile_overlap: Optional[tuple[int, ...]] = (48, 48),
         axes: Optional[str] = None,
         data_type: Optional[Literal["array"]] = None,
         tta_transforms: bool = True,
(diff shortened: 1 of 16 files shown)


veegalinova:
I think the suggestion to return the filename from the dataset is great.
We could use namedtuple or dict as they both [work with collate_fn](https://pytorch.org/docs/stable/data.html#working-with-collate-fn).

As an example, the return structure could look like this:

```
class DataItem(NamedTuple):
    data_patch: torch.Tensor
    tile_info: TileInformation
    file_path: Path
```

But I think it is out of scope for this PR and something to discuss later.

veegalinova:
> Should I update the implementation so the saved file structure mirrors the input data (instead of 1 file per sample)

I think the saved file should indeed be the same shape as the input. Otherwise, the user would need to combine them back together manually. 
Is there a way to "cache" the samples until the S dimension is exhausted and then reshape it back to the input dimensions before writing? Or would it require too many changes in the current code?

src/careamics/careamist.py
```diff
@@ -174,9 +181,23 @@ def __init__(
...
+        # TODO: move to method
+        # prediction writer callback reference
+        # placeholder write strategy — will be changed at predict_to_disk call
+        # TODO: use a placeholder write strategy that will raise an error?
+        write_strategy = create_write_strategy("tiff", tiled=False)
+        self.prediction_writer = PredictionWriterCallback(
+            write_strategy=write_strategy, dirpath=self.work_dir / "predictions"
+        )
+        # writing predictions is turned off
+        self.prediction_writer.set_writing_predictions(False)
```
veegalinova, line 198: Could you clarify why writing_predictions are turned on and off throughout the code? What's the logic behind it?

melisande-c:
> I think the suggestion to return the filename from the dataset is great.
> We could use namedtuple or dict as they both [work with collate_fn](https://pytorch.org/docs/stable/data.html#working-with-collate-fn).

I like this, I have been thinking the returns from the datasets should probably be more explicit. There are places where the output is unpacked as `x, *aux`, but I think namedtuples can be unpacked, so this is ok. What will cause problems, however, is that I am pretty sure there are some places where an output's length is used to infer whether it is tiled or not.

I think I can work on a different PR to modify the datasets to return namedtuples. Once that is done, I can integrate it into this PR.

melisande-c:
> > Should I update the implementation so the saved file structure mirrors the input data (instead of 1 file per sample)
> 
> I think the saved file should indeed be the same shape as the input. Otherwise, the user would need to combine them back together manually. Is there a way to "cache" the samples until the S dimension is exhausted and then reshape it back to the input dimensions before writing? Or would it require too many changes in the current code?

This would be the route I would take I think. Where the `WriteStrategy` class gets the S dimension from is the problem. (Similar problem to where does it get the filename from).

src/careamics/careamist.py
```diff
@@ -596,9 +617,265 @@ def predict(
...
+        ValueError
+            If mean and std are not provided in the configuration.
+        ValueError
+            If tile size is not divisible by 2**depth for UNet models.
+        ValueError
+            If tile overlap is not specified.
+        """
+        # create and set correct write strategy
+        tiled = tile_size is not None
+        write_strategy = create_write_strategy(
```
veegalinova, line 745: Could it be a bit more concise to pack the write strategy selection into the prediction_writer itself, something like `self.prediction_writer.set_writing_strategy(write_type, ...)` and create the `write_strategy` there directly?

src/careamics/careamist.py
melisande-c: Callbacks have to be given to the lightning trainer at initialisation, which is in `CAREamist.__init__`, (lightning does not have any mechanism to add callbacks at a later time). This means without the toggle, predictions will be written every time `trainer.predict` is called, including when it is called in `CAREamist.predict`. Firstly, users may not want to write predictions every time they predict. Secondly, as it is currently implemented, it is not compatible with predicting on arrays, because the prediction writer cannot get the filenames from anywhere.

src/careamics/careamist.py
melisande-c: This was a deliberate design choice to reduce coupling between `PredictionWriterCallback` and the `WriteStrategy` classes. Removing all responsibility of `WriteStrategy` construction from `PredictionWriterCallback` means that the write strategies can be almost arbitrarily altered, with the constraint that they have a `write_batch` method, without having to alter the `PredictionWriterCallback`. This is the key aim the strategy pattern.

melisande-c: Ok I did break the pattern rules with the `from_write_func_params` method though 😅. I put this in for convenience of users of the lightning API, but I am now considering removing. The issue is that the `create_write_strategy` might have to change a bit because I don't know the parameters that the `WriteZarrTiles` strategy will need for initialisation. This means the arguments for `from_write_func_params` will also need to change and also those of a potential `set_write_strategy` method.

melisande-c:
> > > Should I update the implementation so the saved file structure mirrors the input data (instead of 1 file per sample)
> > 
> > 
> > I think the saved file should indeed be the same shape as the input. Otherwise, the user would need to combine them back together manually. Is there a way to "cache" the samples until the S dimension is exhausted and then reshape it back to the input dimensions before writing? Or would it require too many changes in the current code?
> 
> This would be the route I would take I think. Where the `WriteStrategy` class gets the S dimension from is the problem. (Similar problem to where does it get the filename from).

Actually I've thought about it a bit more and the filename could be used to solve this problem. Samples will be "cached" until the filename changes and then they can be concatenated and saved.

melisande-c:
I will close this while I make some changes. 

I need to add "sample caching" so that the prediction file structure matches that of the input file structure. 

I also need to find a solution to how the prediction writer will access the file names, a simpler solution (that I was originally trying to avoid) is simply pass the filenames to the write strategy classes. I was trying to avoid this because it will be annoying for users not using the `CAREamist` api class. I will try and find a way to make it feel convenient.

Also as a small note, I will reiterate, the main motivation for this whole set up is to be able to save tiles to Zarr when potentially an entire file doesn't fit into memory. This feature probably won't be implemented for a while but I need to make sure it won't be difficult.

PR #197, opened 2024-08-05
Fix: serialize data statistics as strings

### LINKED ISSUE #182: export_to_bmz [BUG]

### LINKED ISSUE #188: Serialization of np.float32 prevents configuration loading [BUG]
**Describe the bug**

Raised in https://github.com/CAREamics/careamics/issues/182.

When `np.float32` are passed to `DataConfig.update_means_and_stds`, the types are conserved and not cast to `float` by Pydantic. In turn, this causes `pyyaml` to export it as a class rather than a plain float.

**To Reproduce**
```python
from typing import Any, Annotated
import numpy as np
from pydantic import BaseModel, ConfigDict
import yaml
from pathlib import Path


def write_yaml(path, config):
    with open(path, "w") as f:
        # dump configuration
        yaml.dump(config.model_dump(), f, default_flow_style=False, sort_keys=False)

def read_yaml(path):
    return yaml.load(Path(path).open("r"), Loader=yaml.SafeLoader)


class StatModel(BaseModel):

    model_config = ConfigDict(
        validate_assignment=True,
    )

    means: list[float]
    stds: list[float]

    def _update(self, **kwargs: Any) -> None:
        """
        Update multiple arguments at once.

        Parameters
        ----------
        **kwargs : Any
            Keyword arguments to update.
        """
        self.__dict__.update(kwargs)
        self.__class__.model_validate(self.__dict__)

    def update_mean_std(self, means, stds):
        means= list(means)
        stds = list(stds)
        
        self._update(means=means, stds=stds)

nn = StatModel(means = [np.float32(3.2354)], stds = [np.float32(0.57)])
nn.update_mean_std(
    np.array([3.14], dtype=np.float32),
    np.array([0.57], dtype=np.float32)
)

write_yaml("test.yaml", nn)
model_dict = read_yaml("test.yaml")
```

Causes the following error:
``` python
ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/apply:numpy.core.multiarray.scalar'
```

### 1 - Use serializer

The following code would fix the issue by telling Pydantic how to serialize the values:
```python
from typing import Any, Annotated, Optional
import numpy as np
from pydantic import BaseModel, ConfigDict, PlainSerializer, model_validator
import yaml
from pathlib import Path


def write_yaml(path, config):
    with open(path, "w") as f:
        # dump configuration
        yaml.dump(config.model_dump(), f, default_flow_style=False, sort_keys=False)

def read_yaml(path):
    return yaml.load(Path(path).open("r"), Loader=yaml.SafeLoader)

def np_float_to_str(x):
    return str(x)

# use an annotated float
Float = Annotated[float, PlainSerializer(np_float_to_str, return_type=str)]

class StatModel(BaseModel):

    model_config = ConfigDict(
        validate_assignment=True,
    )

    means: Optional[list[Float]] = None
    stds: Optional[list[Float]] = None


    @model_validator(mode="after")
    def std_only_with_mean(self):
        """
        Check that mean and std are either both None, or both specified.

        Returns
        -------
        Self
            Validated data model.

        Raises
        ------
        ValueError
            If std is not None and mean is None.
        """
        # check that mean and std are either both None, or both specified
        if (self.means and not self.stds) or (
            self.stds and not self.means
        ):
            raise ValueError(
                "Mean and std must be either both None, or both specified."
            )

    def _update(self, **kwargs: Any) -> None:
        """
        Update multiple arguments at once.

        Parameters
        ----------
        **kwargs : Any
            Keyword arguments to update.
        """
        self.__dict__.update(kwargs)
        self.__class__.model_validate(self.__dict__)

    def update_mean_std(self, means, stds):
        means= list(means)
        stds = list(stds)
        
        self._update(means=means, stds=stds)

nn = StatModel()
nn.update_mean_std(
    np.array([3.14], dtype=np.float32),
    np.array([0.57], dtype=np.float32)
)

write_yaml("test.yaml", nn)
model_dict = read_yaml("test.yaml")
StatModel(**model_dict)
```

### 2 - Modify `update_means_and_stds`

We could also force the casting in the update function:
``` python
from typing import Any, Optional
import numpy as np
from pydantic import BaseModel, ConfigDict, model_validator
import yaml
from pathlib import Path


def write_yaml(path, config):
    with open(path, "w") as f:
        # dump configuration
        yaml.dump(config.model_dump(), f, default_flow_style=False, sort_keys=False)

def read_yaml(path):
    return yaml.load(Path(path).open("r"), Loader=yaml.SafeLoader)


class StatModel(BaseModel):

    model_config = ConfigDict(
        validate_assignment=True,
    )

    means: Optional[list[float]] = None
    stds: Optional[list[float]] = None


    @model_validator(mode="after")
    def std_only_with_mean(self):
        """
        Check that mean and std are either both None, or both specified.

        Returns
        -------
        Self
            Validated data model.

        Raises
        ------
        ValueError
            If std is not None and mean is None.
        """
        # check that mean and std are either both None, or both specified
        if (self.means and not self.stds) or (
            self.stds and not self.means
        ):
            raise ValueError(
                "Mean and std must be either both None, or both specified."
            )

    def _update(self, **kwargs: Any) -> None:
        """
        Update multiple arguments at once.

        Parameters
        ----------
        **kwargs : Any
            Keyword arguments to update.
        """
        self.__dict__.update(kwargs)
        self.__class__.model_validate(self.__dict__)


    def update_mean_std(self, means, stds):
        new_means = [float(x) for x in means]
        new_stds = [float(x) for x in stds]

        self._update(means=new_means, stds=new_stds)

    @model_validator(mode="after")
    def std_only_with_mean(self):
        """
        Check that mean and std are either both None, or both specified.

        Returns
        -------
        Self
            Validated data model.

        Raises
        ------
        ValueError
            If std is not None and mean is None.
        """
        # check that mean and std are either both None, or both specified
        if (self.means and not self.stds) or (
            self.stds and not self.means
        ):
            raise ValueError(
                "Mean and std must be either both None, or both specified."
            )

nn = StatModel()
nn.update_mean_std(
    np.array([3.14], dtype=np.float32),
    np.array([0.57], dtype=np.float32)
)

write_yaml("test.yaml", nn)
model_dict = read_yaml("test.yaml")
StatModel(**model_dict)
```

### 3 - Prevent numpy arrays to be passed altogether

This will not prevent this bug to reappear, but should fix the current code base.

### Which one to choose?

I favor the first one, which is more elegant. We would definitely need to add test covering this error and the various ways CAREamics might update the stats (np.ndarray, tuple, list).

@melisande-c @CatEek @veegalinova Any opinion?

jdeschamps:
### Description

Following https://github.com/CAREamics/careamics/issues/182 and https://github.com/CAREamics/careamics/issues/188, the method `set_means_and_stds` of `DataConfig` can be called with `np.float32`. However, when exporting the full configuration to yaml, `np.float32` are considered their own type and yaml cannot deserialize it automatically (https://github.com/CAREamics/careamics/issues/188).

There are many ways around it (e.g. tell pyyaml how to deal with `np.float32`, or make sure that CAREamics never sets `np.float32` as stats). One easy and elegant way is to simply annotate the types in the Pydantic config so that they are serialized as strings.

- **What**: Change the type used for statistics in `DataConfig`.
- **Why**: This forces serialization of `float` and `np.float32` as strings.
- **How**: Use an annotated type and a `PlainSerializer`.

### Changes Made

- **Modified**: `data_model.py`.

### Related Issues

- Fixes https://github.com/CAREamics/careamics/issues/188
- Resolves https://github.com/CAREamics/careamics/issues/182

Serializing `np.float32` as `str` and casting it to `float` upon reading the yaml file will probably cause some changes in the small decimals, which is unavoidable unless we make sure to cast the string back as a `np.float32`. 

---

- [x] Pre-commit passes


src/careamics/config/data_model.py
```diff
@@ -22,6 +23,30 @@
...
+    ----------
+    x : float
+        Input value.
+
+    Returns
+    -------
+    str
+        String representation of the input value.
+    """
+    return str(x)
```
melisande-c: Maybe we can use [`np.format_float_scientific(x, precision=chosen_precision)`](https://numpy.org/doc/stable/reference/generated/numpy.format_float_scientific.html), this means we can choose the precision that is saved.

jdeschamps: Very good idea!

melisande-c (approved):

CatEek (approved):

PR #205, opened 2024-08-12
Refactor: Noise Models

federico-carrara:
### Description

Refactoring of Likelihood modules (Gaussian & Noise Models) for Ladder VAE model. This PR is a first step towards the implementation of the `training_step` and `validation_step` methods for LVAE lightining module

- **What**: Continued the work started in #187 with the aim of polishing and testing the code for loss computation.
- **Why**: The existing code used to work with different configs and contained some superfluous parts, which added annoying complexity and length to the overall LVAE implementation. 
- **How**:  Replaced old configs with `pydantic` models, removed unused functions and class attributes and methods, simplified the logic in some parts of the code, and added tests to check that modules work as expected.

### Changes Made

- **Added**: More comprehensive tests for likelihoods and noise models. The rest is just the refactoring of existing code.
- **Modified**: Among many small modifications to make code cleaner and clearer, structural refactoring (compared to #187) was applied to the model used for noise models. Specifically, the noise model is now modeled by ` MultiChannelNmModel` as a list of `GaussianMixtureNmModel`'s, one for each target channel (e.g., 1 in HDN, >1 for denoiSplit). For consistency, these change were also applied to the noise model instantiation and likelihood computation accordingly. 
- **Removed**: Some deprecated tests for noise model.

---

- [x] Pre-commit passes


src/careamics/config/likelihood_model.py
```diff
@@ -3,39 +3,40 @@
...
 
-class GaussianLikelihoodModel(BaseModel):
-    """Gaussion likelihood model.
 
-    Parameters
-    ----------
-    BaseModel
-    """
+class GaussianLikelihoodModel(BaseModel):
+    """Gaussion likelihood model."""
```
melisande-c: nitpick: misspelling of Gaussian in docs. Although seems like this was here before.

src/careamics/config/nm_model.py
```diff
@@ -11,12 +12,53 @@ class GaussianMixtureNmModel(BaseModel):
     model_config = ConfigDict(
         validate_assignment=True, arbitrary_types_allowed=True, extra="allow"
     )
-    # TODO What are all these parameters?
-    model_type: Literal["GaussianMixtureNoiseModel"]
-    weight: Any = None  # TODO wtf ?
+    model_type: Literal["GaussianMixtureNoiseModel"]  # TODO: why do we need this?
+    """Model type."""
```
melisande-c, line 18: I think this has been put here as a Pydantic discriminator ([see docs](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators))

It's basically to tell pydantic that it only has to check this field to make sure a given value is valid. Not sure if this is actually used anywhere for the noise models but we can keep it for now.

@CatEek @jdeschamps correct me if I'm wrong

src/careamics/config/vae_algorithm_model.py
```diff
@@ -69,7 +69,7 @@ class VAEAlgorithmConfig(BaseModel):
     loss: Literal["musplit", "denoisplit"]
     model: Union[LVAEModel, CustomModel] = Field(discriminator="architecture")
 
-    noise_model: GaussianMixtureNmModel = Field(discriminator="model_type")
+    noise_model: Optional[NMModel] = None
```
melisande-c: This is where the discriminator field `model_type` was being used, (relating to my previous comment).

src/careamics/models/lvae/likelihoods.py
```diff
@@ -271,79 +276,73 @@ class NoiseModelLikelihood(LikelihoodModule):
...
-    def mode(params):
+    def mode(params: dict[str, torch.Tensor]) -> torch.Tensor:
         return params["mean"]
 
     @staticmethod
-    def sample(params):
-        # p = Normal(params['mean'], (params['logvar'] / 2).exp())
-        # return p.rsample()
+    def sample(params: dict[str, torch.Tensor]) -> torch.Tensor:
         return params["mean"]
```
melisande-c, line 328: I have to admit I find these methods strange (I know these were here before, @federico-carrara, so we do not have to address it now).

src/careamics/models/lvae/likelihoods.py
```diff
@@ -104,83 +114,64 @@ def forward(
...
         Given the output of the top-down pass, compute the mean and log-variance of the
         Gaussian distribution defining the likelihood.
 
         Parameters
         ----------
         x: torch.Tensor
-            The input tensor to the likelihood module, i.e., the output of the top-down pass.
+            The input tensor to the likelihood module, i.e., the output
+            the LVAE 'output_layer'. Shape is: (B, 2 * C, [Z], Y, X) in case
+            `predict_logvar` is not None, or (B, C, [Z], Y, X) otherwise.
```
melisande-c, line 158: So just to clarify, you have to know the value of the attribute `predict_logvar` to know what dimensions of `x` are valid for this method...?

(Not to be addressed now but we should make a note to refactor this, especially since, now that `parameter_net` is always the identity the mean is not altered at all by this method.)

melisande-c: So yeah, it seems if the options for `"channelwise"` and `"global"` are removed, then this function basically does nothing anymore, apart from optionally clip the logvar.

src/careamics/models/lvae/noise_models.py
```diff
@@ -35,101 +30,43 @@ def noise_model_factory(
...
-                    nmodel.make_learnable()
 
-        return DisentNoiseModel(*nmodels)
-    return None
+class MultiChannelNoiseModel(nn.Module):
+    def __init__(self, nmodels: list[GaussianMixtureNoiseModel]):
+        """Constructor.
 
+        To handle noise models and the relative likelihood computation for multiple
+        output channels (e.g., muSplit, denoiseSplit).
```
melisande-c, line 56: Question: so is the Pydantic class `NMModel` a configuration for this class?

melisande-c (commented):
Firstly, thank you so much for putting in the work to add some clearer documentation! It was much needed. The removal of some of the vestigial research code has also helped with clarity.

Looks good! Added some small comments that don't necessarily have to be addressed now. I can't comment on the noise model code because I don't understand it yet 😅 but it looks much more simplified removing all the `nmodel1`, `nmodel2` stuff so nice one! I haven't looked at the tests yet either but I'm sure they're fine.

Will let other's take a look at this before approving.

src/careamics/models/lvae/noise_models.py
federico-carrara: Good point! Yes, `NMModel` should act as a config for this. I guess I should change typing :)

src/careamics/models/lvae/likelihoods.py
federico-carrara: Yes, `predict_logvar` determines the number of channels in `x`. Moreover, `parameter_net` is not used anymore here, since it was moved into the LVAE `nn.Module` (which makes waaaaay more sense)

src/careamics/config/vae_algorithm_model.py
federico-carrara: Yes, I see. Anyway, with the new NM config, we won't "expose" the underlying channel-wise noise model `model_type` anymore (which, in addition to the currently implemented GaussianMM, can also be "Histogram-based").

src/careamics/models/lvae/noise_models.py
federico-carrara: No actually the typing should not change here. Indeed, `GaussianMixtureNoiseModel` are the noise model instances for the single channels and not the configs. Indeed, the configs are used in the `noise_model_factory()` to instantiate the noise models as `nn.Module`s.

melisande-c: Maybe the name of `NMModel` can be updated to reflect this? Following the style of other places CAREamics where Pydantic is used to instantiate other classes, e.g. network architectures (`UNetModel` for `UNet` etc.).

src/careamics/models/lvae/noise_models.py
```diff
@@ -1,135 +1,76 @@
...
+            List of noise models, one for each output channel.
         """
         super().__init__()
-        # self.nmodels = nmodels
         for i, nmodel in enumerate(nmodels):
             if nmodel is not None:
-                self.add_module(f"nmodel_{i}", nmodel)
+                self.add_module(
+                    f"nmodel_{i}", nmodel
+                )  # TODO: wouldn't be easier to use a list?
```
melisande-c, line 73: In reply to TODO: it might be easier, yes, but you have to use [`torch.nn.ModuleList`](https://pytorch.org/docs/stable/generated/torch.nn.ModuleList.html)

tests/likelihood_modules/test_likelihoods.py
```diff
@@ -1,45 +1,88 @@
...
 from careamics.config.likelihood_model import GaussianLikelihoodModel, NMLikelihoodModel
-from careamics.config.nm_model import GaussianMixtureNmModel
+from careamics.config.nm_model import GaussianMixtureNmModel, NMModel
 from careamics.models.lvae.likelihoods import likelihood_factory
 from careamics.models.lvae.noise_models import noise_model_factory
 
 
-def test_gaussian_likelihood():
-    config = GaussianLikelihoodModel(model_type="GaussianLikelihoodModel")
+def create_dummy_noise_model(
```
veegalinova, line 15: `create_dummy_noise_model` can be a fixture that is reused in both `test_likelihoods` and `test_noise_model`, could add a todo and do it later

src/careamics/models/lvae/likelihoods.py
```diff
@@ -46,41 +47,50 @@ def likelihood_factory(config: Union[GaussianLikelihoodModel, NMLikelihoodModel,
         raise ValueError(f"Invalid likelihood model type: {config.model_type}")
 
 
+# TODO: is it really worth to have this class? Or it just adds complexity?
```
veegalinova: I think it's not worth it to have this class only to reuse forward

src/careamics/models/lvae/noise_models.py
federico-carrara: Ok, we can discuss this to find a better way to do it :)

tests/likelihood_modules/test_likelihoods.py
federico-carrara: I'll be happy to sit down and see how to do that ;)

src/careamics/models/lvae/noise_models.py
federico-carrara: Update: after quick discussion with @melisande-c and @veegalinova we decided for the moment to proceed with renaming `NMModel` to a more coherent `MultiChannelNmModel`. Further refactoring will be left for later :)

melisande-c (approved):
Amazing!

PR #206, opened 2024-08-14
Feature: LVAE prediction functions

melisande-c:
### Description

- **What**: Add functions to obtain sample predictions and MMSE predictions from an LVAE model in pure PyTorch. These functions shouldn't be too hard to convert to lightning's `predict_step` once there is a more complete lightning module and the tiling functionality is added.
- **Why**: The existing prediction functions were coupled with the custom `MultiChDloader` class, and also relied on functions that took the `target` as input — which shouldn't be available at true inference time.
- **How**: Implemented prediction functions for a single input (will be batch in `predict_step`), removing dependence on `MultiChDloader` class. Extracted only relevant sections from previous prediction logic.

### Changes Made

- **Added**: 
  - `lvae_predict_single_sample` function.
  - `lvae_predict_mmse` function.
  - Basic tests for the above functions.
- **Modified**:
  - Refactored `GaussianLikelihood.get_mean_lv` (removed cases where `predict_logvar in ["channelwise", "global"]` 

Note: If #205 is merged first pre-commit will pass as documentation has been added to the likelihood classes.

---

- [x] Pre-commit passes


src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -0,0 +1,91 @@
...
+        # only keep the log variance of the first sample prediction
+        # TODO: confirm
+        if mmse_idx == 0:
+            log_var = lv
+
+        # store sample predictions
+        sample_predictions[mmse_idx, ...] = sample_prediction
+
+    # take the mean of the sample predictions
+    mmse_prediction = torch.mean(sample_prediction, dim=0, keepdim=True)
```
veegalinova: Probably meant `sample_predictions` here?

src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -0,0 +1,91 @@
...
+    return mmse_prediction, log_var
```
veegalinova, line 95: My ide swears here that log_var might be referenced before assignment, which is not the case, but setting it to None in the beggining shouldn't hurt I guess

src/careamics/models/lvae/likelihoods.py
```diff
@@ -3,7 +3,7 @@
 """
 
 import math
-from typing import Dict, Literal, Tuple, Union
+from typing import Dict, Literal, Tuple, Union, Optional
 
 import numpy as np
```
veegalinova: numpy import is not used

src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -0,0 +1,91 @@
...
+        output: torch.Tensor
+        output, _ = model(input)  # 2nd item is top-down data dict
+
+    # TODO: what if this is likelihood_NM (Not implemented at the moment)
+    # not sure how this will be fully implemented
+    #   -> potentially log_var is always None ?
+    likelihood_obj = model.likelihood_gm
+
+    sample_prediction: torch.Tensor
+    log_var: Optional[torch.Tensor]
```
veegalinova: Is this declaration made for mypy? I think it should be fine since the return types are specified in get_mean_lv

veegalinova:
Looks good! Made a few small comments

src/careamics/prediction_utils/lvae_prediction.py
melisande-c: Ah I think this was left over from when I wasn't using the `get_mean_lv` method, can remove.

src/careamics/prediction_utils/lvae_prediction.py
melisande-c: I guess if `mmse_count=0`, in which case an error should probably be raised, will fix.

src/careamics/prediction_utils/lvae_prediction.py
melisande-c: Oh whoops, nice spot, thanks

src/careamics/models/lvae/likelihoods.py
melisande-c: Not my code, but I can remove

src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -0,0 +1,91 @@
...
+    # presently, get_mean_lv just splits the output in 2 if predict_logvar=True,
+    #   optionally clips the logvavr if logvar_lowerbound is not None
+    sample_prediction, log_var = likelihood_obj.get_mean_lv(output)
+
+    # TODO: output denormalization using target stats that will be saved in data config
+
+    return sample_prediction, log_var
```
melisande-c: @ashesh-0 can you have a look to make sure these prediction functions still have the same logic as the original code?

ashesh-0: This is correct.

src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -0,0 +1,91 @@
...
+    output_shape = (input_shape[0], model.target_ch, *input_shape[2:])
+    log_var: Optional[torch.Tensor]
+    # pre-declare empty array to fill with individual sample predictions
+    sample_predictions = torch.zeros(size=(mmse_count, *output_shape))
+    for mmse_idx in range(mmse_count):
+        sample_prediction, lv = lvae_predict_single_sample(model, input)
+        # only keep the log variance of the first sample prediction
+        # TODO: confirm
+        if mmse_idx == 0:
+            log_var = lv
```
ashesh-0, line 88: Yup.should be fine.

ashesh-0 (approved):

src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -0,0 +1,92 @@
...
+        output, _ = model(input)  # 2nd item is top-down data dict
+
+    # TODO: what if this is likelihood_NM (Not implemented at the moment)
+    # not sure how this will be fully implemented
+    #   -> potentially log_var is always None ?
+    likelihood_obj = model.likelihood_gm
+
+    # presently, get_mean_lv just splits the output in 2 if predict_logvar=True,
+    #   optionally clips the logvavr if logvar_lowerbound is not None
+    sample_prediction, log_var = likelihood_obj.get_mean_lv(output)
```
federico-carrara, line 43: This is a bit hacky...
Indeed, in this function, we take the sample as the mean predicted by the LVAE model (i.e., the output of LVAE top-down is `(mean, logvar)`, we take `mean` as our sample). Therefore, the only reason why the likelihood is used is to make use of its `get_mean_lv()` method, which, given the output of the LVAE, extracts mean and logvar (if needed). 

This said, I would propose not to go through the likelihood, but simply report the code needed in this case, which would be something like:
```
if self.predict_logvar is not None:
            # Get pixel-wise mean and logvar
            mean, lv = x.chunk(2, dim=1)
else:
            mean = x
            lv = None
```
I believe code readability and the overall comprehension of what is happening could benefit from this change. Using the likelihood is highly misleading since it doesn't really play a role here.
Please let me know if you agree (especially @ashesh-0).

ashesh-0: Well, it is debatable. I would not expose how the likelihood object operates in a standalone prediction module. I would let the likelihood object to decide on how to get the log_var.
I see  that your approach has less code, but the one present is more modular.

melisande-c: I agree that using the likelihood object is potentially misleading and misrepresents the intention of the code since we are not using the true functionality of the likelihood. And I did actually start off with the implementation above (with also the optional clipping of the lowerbound of the logvar). However, I reverted back to using the `get_mean_lv` method because I still had to use two attributes from the likelihood object, namely `predict_logvar` and `logvar_lowerbound`, and I wanted to keep consistency with the training and validation steps (although I think the mean & logvar splitting happens within `reconstruction_loss`).

Additionally I think we could consider changing the output of the model itself so that the `mean` and `logvar` are already split apart. i.e. `mean, logvar, topdown_dict = model(x)`.

Therefore I reckon we make a note to refactor this when we have more of the pipeline in place, because the choice we make now might not be optimal for the final product.

federico-carrara: "_I agree that using the likelihood object is potentially misleading and misrepresents the intention of the code since we are not using the true functionality of the likelihood._"

This is exactly the point I was trying to make. My opinion is that in situations in which the code is complex and involves a lot of steps being explicit is always better than implicit. 

At the same time, I also agree to just leave a note for the moment and leave it for future discussion.

ashesh-0: well, at this point, anything and everything which keeps the code working is perfectly fine for me. :P 

However, if you think about the question: which module should decide what is logvar and what is mean of the deocder's output distribution. It should be the likelihood module because, it could, for example, say that the backbone's output is not the log of variance, but variance.  Alternatively, if you want to treat the likelihood class as simply a loss computation module, which would be a perfect valid treatment, then as @melisande-c suggests, one should do the split of mean and logvar in the main model itself.

melisande-c: > well, at this point, anything and everything which keeps the code working is perfectly fine for me. :P

Yes I didn't want to introduce too many drastic changes, especially on sections I don't fully understand, which may lead to unexpected behaviour down the road. We will leave this to be addressed at a later date.

melisande-c: Having said that, since #205, the likelihood is no longer an attribute of the LVAE model. To solve this issue I think I will simply pass the likelihood object as an argument to the prediction functions. If I want to remove dependence on the likelihood model as @federico-carrara suggests then I will have to pass both `predict_logvar` and `logvar_lowerbound` to the prediction functions.

federico-carrara (approved):

PR #207, opened 2024-08-19
Feature: Add tiling method from Disentangle repository

melisande-c:
### Description

The Disentangle repository takes a slightly different approach to tiling results than the method in CAREamics. This PR reimplements the former's approach to ensure the reproduction of the results in the microSplit paper, in a way that should be compatible with the existing CAREamics prediction pipeline with minor refactoring.

- **What**: Added a generator `extract_tiles` that yields the extracted tile and and tile information. This function is interchangeable with the existing `careamics.dataset.tiling.extract_tiles`. This means the resulting set of tiles can be passed to the existing `careamics.prediction_utils.stitch_prediction` to reconstruct the full image. The `InMemoryTiledPredDataset` can be refactored so the `extract_tiles` generator, can be chosen by the configuration.
- **Why**:  To ensure the reproduction of the results in the microSplit paper.
- **How**: Moved relevant logic contained in the Disentangle classes [`GridIndexManager`](https://github.com/ashesh-0/Disentangle/blob/3dusplit/disentangle/data_loader/patch_index_manager.py) and [`MultiChDloader`](https://github.com/ashesh-0/Disentangle/blob/3dusplit/disentangle/data_loader/vanilla_dloader.py) to a set of functions contained within the module `careamics/dataset/tiling/lvae_tiled_patching.py`. This includes the `compute_tile_info` function that computes a CAREamics compatible `TileInformation` object from a given tile's location.

### Changes Made

- **Added**:
  - `careamics/dataset/tiling/lvae_tiled_patching.py`
  - `tests/dataset/tiling/test_lvae_tiled_patching.py`
  - `src/careamics/prediction_utils/lvae_tiling_manager.py`
    - This can be ignored. I started by refactoring the `GridIndexManager` class (renamed to `TilingManager`). It is unused because I decided to create a functional implementation, more inline with what already exists in CAREamics. However, I have left it in, commented out, in case it is useful for the dataset implementation.

Bit extra, but diagrams below demonstrate the difference between the two tiling methods. The grey area in the diagram illustrates where padding has been added to array. The Disentangle method ensures the stitched portion of a tile comes from the center.
CAREamics_tiling
Asheshs_tiling

#### Where to find original logic

`extract_tiles` logic mostly comes from:
https://github.com/ashesh-0/Disentangle/blob/ed99b2614a9e52b496947ccf157d3aaa8db52872/disentangle/data_loader/vanilla_dloader.py#L435-L465

Calculation of `overlap_crop_coords` and `stitch_coords` in `compute_tile_info` comes from:
https://github.com/ashesh-0/Disentangle/blob/ed99b2614a9e52b496947ccf157d3aaa8db52872/disentangle/analysis/stitch_prediction.py#L36-L66

All other helper functions are a reimplementation of the logic in the methods of `GridIndexManager`.

#### Note
We should come with a better name for the two different approaches to avoid name clashes.

---

- [x] Pre-commit passes


ashesh-0:
This is super cool !! This covers almost all datasets. The only exception is the 3D datasets. But even for that, there should be not much change needed. 
For 3D, I found that doing reflect padding along z-dimension creates inferior performance. So, I have slightly changed my tiling strategy. But tiles are still in the center. So what you have here will work.  Essentially, overlap_crop_coords and stitch_coords in compute_tile_informations will depend upon tiling mode. But that can be done later.  

Reason for the change: In the Z dimension, you have a sequence: images go, lets say from top of an organelle to bottom. Now, when you do refect padding, this unidirectional movement is perturbed. For example, doing a refect padding of 3 pixels on [1,2,3,4,5,6] will give you `[3,2,1,2,3,4,5,6,5,4]`.    

Also, I wrote tests for stitching here https://github.com/ashesh-0/Disentangle/blob/master/disentangle/tests/analysis/test_stitch_prediction.py. But of course, they are not generic and are tied to my dataset class. So, I'm not sure if they will be helpful.

tests/dataset/tiling/test_lvae_tiled_patching.py
```diff
@@ -0,0 +1,188 @@
...
+        tile_infos.append(tile_info)
+
+        # check tile shape, ignore channel dimension
+        assert tile.shape[1:] == tile_size
+        assert len(overlap_crop_coords) == len(stitch_coords) == len(tile_size)
+
+    # stitch_prediction returns list
+    stitched_arr = stitch_prediction(tiles, tile_infos)[0]
+
+    np.testing.assert_array_equal(arr, stitched_arr)
```
ashesh-0, line 56: I see you already have both integration and unit tests in place.  If they are passing, then all is good !!

src/careamics/dataset/tiling/lvae_tiled_patching.py
```diff
@@ -0,0 +1,282 @@
...
+    # assume spatial dimension are the last dimensions so iterate backwards
+    for i in range(-1, -len(tile_size) - 1, -1):
+        result = result * n_tiles_1d(data_shape[i], tile_size[i], overlaps[i])
+
+    return result
+
+
+def compute_tile_grid_shape(
+    data_shape: NDArray[np.int_],
+    tile_size: NDArray[np.int_],
```
ashesh-0, line 257: overlap  will be a tuple in 3D. So, let's say the model takes as input 5x64x64, where 5 is the z-dimension. so, in your notation this is the tile size. However, we could have`4x48x48` as the overlap so that only the `16x16` of the central z slice in the prediction will be kept and 2 slices from the beginning and the end will be removed. But nothing to be done for now. just writing here to make you aware.

src/careamics/dataset/tiling/lvae_tiled_patching.py
```diff
@@ -0,0 +1,282 @@
...
+    -------
+    tuple of (int, int)
+        A tuple specifying the padding to add in each dimension, each element is a two
+        element tuple specifying the padding to add before and after the data. This
+        can be used as the `pad_width` argument to `numpy.pad`.
+    """
+    tile_grid_shape = np.array(compute_tile_grid_shape(data_shape, tile_size, overlaps))
+    covered_shape = (tile_size - overlaps) * tile_grid_shape + overlaps
+
+    pad_before = overlaps // 2
```
ashesh-0, line 201: This will change for the boundary tiles. it will not be a constant.  At boundaryt, we will want to keep more content because  we want to predict for the entire possibly irregular sized frame.

src/careamics/dataset/tiling/lvae_tiled_patching.py
```diff
@@ -0,0 +1,282 @@
...
+        (start, end) for start, end in zip(stitch_coords_start, stitch_coords_end)
+    )
+    overlap_crop_coords = tuple(
+        (start, end)
+        for start, end in zip(overlap_crop_coords_start, overlap_crop_coords_end)
+    )
+
+    # --- Check if last tile
+    tile_grid_shape = np.array(compute_tile_grid_shape(data_shape, tile_size, overlaps))
+    last_tile = (tile_grid_coords == (tile_grid_shape - 1)).all()
```
ashesh-0, line 158: I am not sure how last_tile is handled in CAREmics. But I think this should allow us to use more of the prediction than `overlaps` parameter would normally allow.

ashesh-0 (approved):

src/careamics/dataset/tiling/lvae_tiled_patching.py
ashesh-0: but nothing to be done now.

ashesh-0:
There could be one potential issue. I used the same grid management logic to both tile the predictions during evaluation time and get the `correct` patch for a given index `__getitem__` for my validation dataset during training. The validation dataset needs to produce all patches, which is the only constraint that needs to be satisfied to reproduce the results. We will need some logic on the validation dataset side if we do not want to use the GridIndexManager class . @veegalinova may have to take a decision I guess.

melisande-c:
> There could be one potential issue. I used the same grid management logic to both tile the predictions during evaluation time and get the `correct` patch for a given index `__getitem__` for my validation dataset during training. The validation dataset needs to produce all patches, which is the only constraint that needs to be satisfied to reproduce the results. We will need some logic on the validation dataset side if we do not want to use the GridIndexManager class . @veegalinova may have to take a decision I guess.

I think as a first step it is totally fine to have the prediction pipeline using CAREamics, and the train and validation still using the `MultiChDloader` & `GridIndexManager`. We can move iteratively towards better integration, but it's also important to get something working quickly.

Side note: the `tile_grid_coords` argument (which maybe I should have named indices) to the `compute_tile_info` function can be calculated from the index passed to `__getitem__` using `np.unravel_index`.
e.g.
```python
>>> np.unravel_index(indices=10, shape=(7, 7))
(1, 3)
>>> np.arange(49).reshape(7, 7)[1, 3]
10
```

src/careamics/dataset/tiling/lvae_tiled_patching.py
melisande-c: Not sure if this comment is in relation to the change in tiling strategy for 3D data, but this is the padding for the entire sample and not individual tiles, like in your implementation. Therefore padding is only added to the boundary. If you look in `extract_tiles`, the padding is added to the entire image before it is split into tiles. 😊

src/careamics/dataset/tiling/lvae_tiled_patching.py
melisande-c: `last_tile` doesn't affect the stitching, it is just used as as flag to know when the tiles from the next sample begin; the `stitch_prediction` function in CAREamics can be used on multiple samples at once. Only one tile is labelled as `last_tile`, and that is the one which will be yielded last by the generator, i.e. in 2D the bottom-right tile, and in 3D the equivalent for C-ordered indexing.

src/careamics/dataset/tiling/lvae_tiled_patching.py
melisande-c: So some of the boundary z-slices are discarded during prediction?

ashesh-0: If you do it naively, then yes, some of the frames will get discarded. 
But if you handle the boundary tile differently, then it is not necessary right? For the boundary tile, you could take the whole prediction and not just the central frame. I also needed to shift the location of the tile so that I did not need any padding and all the content was predicted for.

src/careamics/dataset/tiling/lvae_tiled_patching.py
ashesh-0: Okay. I misunderstood. this was also in relation to tiling to be used for the 3D case. I thought last_tile was same as boundary tile.   There will be multiple boundary tiles, different boundary tiles for different dimensions.

src/careamics/dataset/tiling/lvae_tiled_patching.py
ashesh-0: Okay. I see. again, I misunderstood since I have the padding logic for each tile. It is correct that it applies padding only for the boundary tiles . Cool. But again, this behavior may change for the 3D case.

src/careamics/dataset/tiling/lvae_tiled_patching.py
melisande-c: I had a quick look at the changes in `GridIndexManager`. It is my understanding that the tiles are shifted so that they are centred on the z-axis, and tiles are not permitted to exceed the bounds of the image (to remove the need for padding). But, in most cases (if image not equally divisible by the grid size), this necessarily leads to some of the boundary z-slices being discarded. Is this correct?

N.b. This implementation handles tiling the exact same way as yours, I just reformulated it so I would be able to integrate it with CAREamics prediction pipeline more easily. The boundary tiles are padded the same, and then the padding is removed during `stitch_prediction`. The area to stitch together is specified by `overlap_crop_coords` in `TileInformation`; the calculation of `overlap_crop_coords` in `compute_tile_info` does not allow beyond the bounds of the original image to be included.

ashesh-0: I guess we can discuss face to face. Whenever you are in the lab. It might be more efficient that way. Since you have changed the structure, it can take me sometime to fully understand which I would prefer to not give :) . Face to face would be better. 

No z-slices are discarded in my current version. And I have overlap in x,y and z.

melisande-c: I see, I will have a proper look at the changes in `GridIndexManager` because I have only skimmed it, and I'll try to get a better understanding. I think I can re-work this so that the tiling strategy can be specified per dimension, but it will basically require me to rewrite the entire thing.

I will be in the lab tomorrow, if you're available? but let's take organising to chat.

PR #208, opened 2024-08-20
General refactoring and testing of LVAE loss functions.

federico-carrara:
### Description

General refactoring and testing of LVAE loss functions.

- **What**: Refactored, polished and tested LVAE losses. 
- **Why**: Loss computation is a key point to correctly train the LVAE model.
- **How**: By implementing missing methods and fixing several bugs, adding documentation to the code (parameters, steps in computation), and removing unused/unnecessary pieces of code.

### Changes Made

- **Added**: tests for all loss functions in `tests/losses/test_lvae_losses`, loss function for `denoisplit_musplit` case (i.e., when both Gaussian and noise model likelihoods are employed).
- **Modified**: several things: 
    (1) the `LVAELossParameters` `dataclass` which provides parameters for loss computation. Specifically, unnecessary attributes were removed from here.
    (2) fixed several bugs from the functions for loss computation in the different cases.
- **Removed**: unused methods, functionalities, and parameters from the different classes and functions.

---

- [x] Pre-commit passes


src/careamics/losses/lvae/losses.py
```diff
@@ -1,152 +1,183 @@
...
-        splitting_mask = torch.ones_like(loss_dict["loss"]).bool()
 
-    # print(len(target) - (torch.isnan(loss_dict['loss'])).sum())
-
-    loss_dict["loss"] = loss_dict["loss"][splitting_mask].sum() / len(reconstruction)
+    loss_dict["loss"] = loss_dict["loss"].sum() / len(reconstruction)
     for i in range(1, 1 + target.shape[1]):
         key = f"ch{i}_loss"
-        loss_dict[key] = loss_dict[key][splitting_mask].sum() / len(reconstruction)
+        loss_dict[key] = loss_dict[key].sum() / len(reconstruction)
```
ashesh-0, line 58: Looks good

src/careamics/losses/lvae/losses.py
```diff
@@ -1,152 +1,183 @@
...
-
     for i in range(1, 1 + target.shape[1]):
         output[f"ch{i}_loss"] = None
 
-    # Log likelihood
-    ll, like_dict = likelihood_obj(reconstruction, target)
-    ll = _get_weighted_likelihood(ll)
+    # Compute Log likelihood
+    ll, _ = likelihood_obj(reconstruction, target)  # shape: (B, C, [Z], Y, X)
+    ll = _get_weighted_likelihood(ll)  # TODO: needed?
```
ashesh-0: This can be removed.

src/careamics/lightning/lightning_module.py
```diff
@@ -299,16 +315,15 @@ def training_step(self, batch: Tensor, batch_idx: Any) -> Any:
...
         target = aux[0]
+
+        # Update loss parameters
         # TODO rethink loss parameters
         self.loss_parameters.current_epoch = self.current_epoch
-        self.loss_parameters.inputs = x
-        self.loss_parameters.mask = ~((target == 0).reshape(len(target), -1).all(dim=1))
-        self.loss_parameters.noise_model_likelihood = self.noise_model_likelihood
-        self.loss_parameters.gaussian_likelihood = self.gaussian_likelihood
-        self.loss_parameters.noise_model = self.noise_model
```
melisande-c, line 311: Just to help me understand, the two likelihoods are added to `LVAELossParameters` within `__init__` and are the other fields not needed anymore, or added somewhere else?

src/careamics/losses/lvae/losses.py
```diff
@@ -1,152 +1,183 @@
...
+    recons_loss_nm = -1 * nm_likelihood(out_mean, targets)[0].mean()
+    recons_loss_gm = -1 * gaussian_likelihood(predictions, targets)[0].mean()
+    recons_loss = nm_weight * recons_loss_nm + gaussian_weight * recons_loss_gm
     return recons_loss
 
 
+# TODO: refactor this (if needed)
+# - cannot handle >2 target channels
+# - cannot handle 3D inputs
 def _get_weighted_likelihood(
```
ashesh-0: This function can be removed.

src/careamics/losses/lvae/losses.py
```diff
@@ -158,94 +189,134 @@ def _get_weighted_likelihood(
...
-        # topdown_layer_data_dict['z'][2].shape[-3:] = 128 * 32 * 32
-        norm_factor = np.prod(topdown_layer_data_dict["z"][i].shape[-3:])
-        # if _restricted_kl:
-        #     pow = np.power(2,min(i + 1, _multiscale_count-1))
-        #     norm_factor /= pow * pow
-
+        # NOTE: we want to normalize the KL-loss w.r.t. the latent space dimensions,
+        # i.e., the number of entries in the latent space tensors (C, [Z], Y, X).
+        # We assume z has shape (B, C, [Z], Y, X), where `C = z_dims[i]`.
+        norm_factor = np.prod(topdown_data["z"][i].shape[1:])
```
ashesh-0, line 182: Looks good.

src/careamics/losses/lvae/losses.py
```diff
@@ -158,94 +189,134 @@ def _get_weighted_likelihood(
...
+
+    # NOTE: at each hierarchy, the KL loss is larger by a factor of (128/i**2).
     # 128/(2*2) = 32 (bottommost layer)
     # 128/(4*4) = 8
     # 128/(8*8) = 2
     # 128/(16*16) = 0.5 (topmost layer)
 
-    # Normalize the KL-loss w.r.t. the  latent space
+    # Normalize the KL-loss w.r.t. the input image spatial dimensions (e.g., 64x64)
     kl_loss = kl_loss / np.prod(img_shape)
```
ashesh-0, line 231: At this location, we additionally need to divide by number of z dimensions for the 3D case. but nothing to be done now.

ashesh-0 (approved):

src/careamics/losses/lvae/losses.py
federico-carrara: Do you mean that we need to divide by $(Z\times Y\times X)$? If this is the case I think it is already suitable for this, as in `denoisplit_loss()` we pass `img_shape=targets[2:]` (see line 386 of the same file) (i.e., all the spatial dimensions, supposing that the shape of `target` is `(B, C, [Z], Y, X)`.

melisande-c (commented):
Firstly thank you again for all the docs! Very helpful! Looks good, tests look like they make sense.

But yeah, I really dislike this `LVAELossParameters` dataclass and how it hides the parameters of the loss functions, instead of stating them explicitly. As the code currently stands, there does not even seem to be a way to change them from the defaults anyway, (although I assume there are plans to add this in the future).

I think it would make much more sense if the losses are classes with the `__call__` method implemented. This way the loss parameters can be initialised in `__init__` by the loss factory which can be described by a pydantic class, e.g. something like `DenoisplitLossModel`. This would be consistent with the other "factory" functions in CAREamics, see [`model_factory`](https://github.com/CAREamics/careamics/blob/ae34bb81077be8c4c8d9864519ce2c76b5d1dcd2/src/careamics/models/model_factory.py#L19-L51).

Obviously this is a fair bit of work though, so we can leave it for later.

Example of what a loss class would look like below, and how you would call it.

```python
class DenoisplitLoss:

    def __init__(
        self,
        noise_model_likelihood: Optional[NoiseModelLikelihood] = None,
        gaussian_likelihood: Optional[GaussianLikelihood] = None,
        current_epoch: int = 0,
        reconstruction_weight: float = 1.0,
        musplit_weight: float = 0.0,
        denoisplit_weight: float = 1.0,
        kl_type: Literal["kl", "kl_restricted", "kl_spatial", "kl_channelwise"] = "kl",
        kl_weight: float = 1.0,
        kl_annealing: bool = False,
        kl_start: int = -1,
        kl_annealtime: int = 10,
        non_stochastic: bool = False,
    ):
        self.noise_model_likelihood = noise_model_likelihood
        self.gaussian_likelihood = gaussian_likelihood
        self.current_epoch = current_epoch
        self.reconstruction_weight = reconstruction_weight
        # etc ...

    def __call__(
        self, model_outputs: tuple[torch.Tensor, dict[str, Any]], targets: torch.Tensor
    ): 
        # calculate loss
        ...

loss_config = DenoisplitLossModel(...) # pydantic class
loss_func = DenoisplitLoss(**loss_config.model_dump())
loss_value = loss_func(model_outputs, targets) 
```

src/careamics/lightning/lightning_module.py
federico-carrara: So let me try to clarify. In the current version of the code, `LVAELossParameters` contains only the bare minimum of parameters that are needed for the loss computation. This means that all the fields that you see removed there are actually removed from the `dataclass` since not needed in any place of the loss computation.
Specifically, these are the "groups" of parameters needed in `LVAELossParameters`:
- likelihood objects
- a set of weights for the different loss terms (to give different importance to different loss terms)
- some additional parameters for KL loss annihilation (not used in ashesh's works, but I would leave it there since I think it is something pretty common to have for VAE-based algoriithms)
- some extra params (e.g., `non_stochastic`, `current_epoch`, ...)

melisande-c: hmm ok, I see thanks. Anyway looks good for now! We will take this refactoring process step by step.

melisande-c (approved):

src/careamics/losses/lvae/losses.py
ashesh-0: Makes sense.

src/careamics/lightning/lightning_module.py
federico-carrara: I'll put a big TODO so that we do not forget about this :)

melisande-c: Nice, thanks!

PR #210, opened 2024-08-21
Noise model training

CatEek:
### Description

Added GMM training 
- **What**: Starting from noise model factory in the lightning module there's an option to either load pretrained model or train new one, per channel
- **Why**: We need training functionality

### Changes Made

- **Added**: nm training func, nm_factory, pudantic configs
- **Modified**: merged current main with some conflict solving
- **Removed**: tests are currently in one file as moved by #205. Need to revisit


src/careamics/config/architectures/lvae_model.py
```diff
@@ -31,6 +31,7 @@ class LVAEModel(ArchitectureModel):
 
     predict_logvar: Literal[None, "pixelwise"] = None
 
+    # TODO this parameter is exessive ?
```
federico-carrara: You're right, this can go away since in the current version noise model and family are independent of the LVAE architecture.

src/careamics/config/nm_model.py
```diff
@@ -14,43 +16,55 @@ class GaussianMixtureNmModel(BaseModel):
...
-    - last n_gaussian rows correspond to the standard deviations
-    If `weight=None`, the weight array is initialized using the `min_signal`
-    and `max_signal` parameters."""
-
+    # TODO What are all these parameters?
+    model_type: Literal["GaussianMixtureNoiseModel"]
+    path: Union[str, Path] = None
+    signal: Union[str, Path, np.ndarray] = None
+    observation: Union[str, Path, np.ndarray] = None
+    weight: Any = None  # TODO wtf ?
```
federico-carrara: I think that by solving conflicts here you accidentally removed the description of parameters that was added in #205. This needs to be fixed

src/careamics/config/vae_algorithm_model.py
```diff
@@ -69,6 +69,7 @@ class VAEAlgorithmConfig(BaseModel):
     loss: Literal["musplit", "denoisplit"]
     model: Union[LVAEModel, CustomModel] = Field(discriminator="architecture")
 
+    # TODO different channels for nm should be where ?
```
federico-carrara: In #205 we decided to adopt a new configuration for Noise Models. Specifically:
- for each output channel (e.g., 1 in HDN, >1 in denoiSplit) we define an actual noise model (for the moment the only choice is `GaussianMixtureNoiseModel`, but in the future we will also have an `HistogramNoiseModel`).
- the previous `DisentNoiseModel`, which gathers the noise models for the single channels, has been replaced by a more clear `MultiChannelNoiseModel`, which basically gathers the different `GaussianMixtureNoiseModel`'s in a list (or something logically similar to a list). 
- therefore, in this new configuration, whenever we need the noise model (e.g., in likelihood/loss computations), we access it through the `MultiChannelNoiseModel` as it provides all the functionalities to compute log-likelihood and related stuff. In addition, this new structure hides the internal parameters of the noise models, which are set a-priori (through training or loading from `path`) and should not be touched during LVAE training or evaluation.

src/careamics/models/lvae/noise_models.py
```diff
@@ -34,18 +34,67 @@ def noise_model_factory(
         If the chosen noise model `model_type` is not implemented.
         Currently only `GaussianMixtureNoiseModel` is implemented.
     """
-    if model_config:
```
federico-carrara, line 37: We need to check if `model_config` exists. Indeed, when we instantiate for example `muSplit` stuff in the `lightining_module` we don't need the noise model, and hence we will pass a `None` `model_config`

src/careamics/models/lvae/noise_models.py
```diff
@@ -34,18 +34,67 @@ def noise_model_factory(
...
-        return MultiChannelNoiseModel(noise_models)
-    # TODO should config and paths be mutually exclusive? What you mean, Igor?
-    return None
+                raise NotImplementedError(f"Model {nm.model_type} is not implemented")
+
+        else:  # TODO this means signal/obs are provided. Controlled in pydantic model
+            # TODO train a new model. Config should always be provided?
+            if nm.model_type == "GaussianMixtureNoiseModel":
+                # TODO one model for each channel all make this choise inside the model?
+                return train_gm_noise_model(nm)
```
federico-carrara: Given the current `MultiChannelNoiseModel`, and assuming that this function returns a `GaussianMixtureNoiseModel`, I think here we'd need something like 
```
noise_models.append(train_gm_noise_model)
```

src/careamics/models/lvae/noise_models.py
```diff
@@ -34,18 +34,67 @@ def noise_model_factory(
...
+    """
+    # TODO where to put train params?
+    # TODO any training params ? Different channels ?
+    noise_model = GaussianMixtureNoiseModel(model_config)
+    # TODO revisit config unpacking
+    noise_model.train(noise_model.signal, noise_model.observation)
+    return noise_model
+
+
+# def get_noise_model(
```
federico-carrara: This function was removed in #205

tests/models/lvae/test_noise_model.py
```diff
@@ -130,3 +130,42 @@ def test_multi_channel_noise_model_likelihood(
...
+    min_sigma = np.random.rand(1)
+    filename = Path(tmp_path) / "gm_noise_model.npz"
+    np.savez(
+        filename,
+        trained_weight=trained_weight,
+        min_signal=min_signal,
+        max_signal=max_signal,
+        min_sigma=min_sigma,
+    )
+    noise_model = noise_model_factory(nm_config, [filename])
```
federico-carrara: This would not work, `noise_model_factory` only needs the `nm_config` now

federico-carrara (changes requested):
Thank you for the PR @CatEek! 
I skimmed through it and found some points in which your code is not getting along with the latest PR concerning Noise Model refactoring #205.
I am afraid that at this point would be difficult and counterproductive to refactor the NM structure (also because NM training should not interact with the outer `MultiChannelNoiseModel` structure). 
This said the only changes that I really deem necessary here are the additional `train()` method for the `GaussianMixtureNoiseModel` (provided with optional necessary attributes) and a light refactoring of `noise_model_factory()` to enable the case in which a path to a trained NM is not provided.

src/careamics/config/nm_model.py
CatEek: Will fix, thanks

src/careamics/config/vae_algorithm_model.py
CatEek: I guess this is what is happening. Sorry if I'm missing something

federico-carrara: It was just a clarification to your TODO, nothing actionable ;)

src/careamics/models/lvae/noise_models.py
```diff
@@ -410,3 +473,86 @@ def getSignalObservationPairs(self, signal, observation, lowerClip, upperClip):
     def forward(self, x, y):
         """Temporary dummy forward method."""
         return x, y
+
+    # TODO taken from pn2v. Ashesh needs to clarify this
```
ashesh-0, line 461: It looks alright. I basically compared it with https://github.com/ashesh-0/PPN2V/blob/master/src/ppn2v/pn2v/gaussianMixtureNoiseModel.py#L257 and it looks quite similar. If you did not change anything (other than jointloss removal) then it is all good.

ashesh-0 (approved):

src/careamics/config/nm_model.py
```diff
@@ -14,43 +16,75 @@ class GaussianMixtureNMConfig(BaseModel):
...
-    """A [3*n_gaussian, n_coeff] sized array containing the values of the weights
-    describing the GMM noise model, with each row corresponding to one
-    parameter of each gaussian, namely [mean, standard deviation and weight].
-    Specifically, rows are organized as follows:
-    - first n_gaussian rows correspond to the means
-    - next n_gaussian rows correspond to the weights
-    - last n_gaussian rows correspond to the standard deviations
-    If `weight=None`, the weight array is initialized using the `min_signal`
-    and `max_signal` parameters."""
+    # A [3*n_gaussian, n_coeff] sized array containing the values of the weights
```
federico-carrara: So we use this convention (# instead of """, and above the parameter declaration) for documenting `pydantic` attributes? I was told differently by @melisande-c. Anyway, I think we should find a common ground for this :)

src/careamics/config/nm_model.py
```diff
@@ -14,43 +16,75 @@ class GaussianMixtureNMConfig(BaseModel):
...
+        Parameters
+        ----------
+        self : Self
+            _description_
+        """
+        if self.path and (self.signal is not None or self.observation is not None):
+            raise ValueError(
+                "Either 'path' to pre-trained noise model should be"
+                "provided or both signal and observation in form of paths"
+                "or numpy arrays"
```
federico-carrara, line 79: I guess this error message is wrong, here you have both `path` and `signals`. Anyway, to clarify, the idea is that we only want either only `path` or only tensors at a time, isn't it?

src/careamics/models/lvae/noise_models.py
```diff
@@ -34,20 +34,76 @@ def noise_model_factory(
         If the chosen noise model `model_type` is not implemented.
         Currently only `GaussianMixtureNoiseModel` is implemented.
     """
-    if model_config:
+    if model_config.noise_models:
```
federico-carrara: Wouldn't this raise an error if `model_config` is `None` (e.g., `NoneType` does not have attribute `noise_models`)? Remember that the config is None, for instance, in the case of muSplit, where we instantiate the lghtning module without requiring the noise model.

src/careamics/models/lvae/noise_models.py
```diff
@@ -34,20 +34,76 @@ def noise_model_factory(
...
+                if nm.model_type == "GaussianMixtureNoiseModel":
+                    noise_models.append(GaussianMixtureNoiseModel(nm))
+                else:
+                    raise NotImplementedError(
+                        f"Model {nm.model_type} is not implemented"
+                    )
+
+            else:  # TODO this means signal/obs are provided. Controlled in pydantic model
+                # TODO train a new model. Config should always be provided?
+                if nm.model_type == "GaussianMixtureNoiseModel":
```
federico-carrara, line 50: Small detail: checking this conditional before checking `nm.path` will avoid code duplication :)

src/careamics/models/lvae/noise_models.py
```diff
@@ -34,20 +34,76 @@ def noise_model_factory(
...
+    """
+    # TODO where to put train params?
+    # TODO any training params ? Different channels ?
+    noise_model = GaussianMixtureNoiseModel(model_config)
+    # TODO revisit config unpacking
+    noise_model.train(noise_model.signal, noise_model.observation)
+    return noise_model
+
+
+# def get_noise_model(
```
federico-carrara: Please, remove this, it is not needed anymore.

src/careamics/models/lvae/noise_models.py
```diff
@@ -217,12 +286,13 @@ def make_learnable(self):
         self.weight.requires_grad = True
 
     def to_device(self, cuda_tensor):
+        # TODO wtf is this ?
```
federico-carrara, line 266: Remove

src/careamics/models/lvae/noise_models.py
federico-carrara: P.S. I'd rather have this instead of instantiating a `MultiChannelNMConfig` with empty `noise_models` list like you do in the test. Indeed, it is more coherent with the fact that we do not actually have a noise model in some cases (e.g., muSplit).

federico-carrara (changes requested):
Igor! There are still a few things that I'd would like to be changed, otherwise will most likely cause annoying issues with the upcoming `training_step` PR (especially see the comment about having a `None` model_config).

src/careamics/models/lvae/noise_models.py
federico-carrara: I can make an example to support my statement:
Suppose we initialize a lightning module for `muSplit`. If one wants to have a look to the internal attributes of the model, it would be way more clean and coherent to have the noise model set to `None`, since in this case it is not needed, and, hence, not defined. Having an empty noise model would make things less linear in my opinion. 

Plus I'd need to refactor the `training_step` code that is coming later, which would be painful since it is a lot of code.

tests/models/lvae/test_noise_model.py
```diff
@@ -28,6 +28,12 @@ def create_dummy_noise_model(
     np.savez(tmp_path / "dummy_noise_model.npz", **nm_dict)
 
 
+def test_factory_no_noise_model():
```
federico-carrara, line 15: @CatEek is there a particular reason why we want this behavior? In my opinion this is more logical:
- if `MultiChannelNMConfig == None`, then the noise model is `None` (see my comments in the previous review)
- if `MultiChannelNMConfig.noise_models == []`, then we simply get an "empty" noise model.

Wouldn't this be more clean and logic?

src/careamics/config/nm_model.py
melisande-c: This is how it is done in the rest of the code base, maybe we can revise at some point in the future but for now let's stay consistent. This way of writing class attribute docstrings is briefly mentioned in [PEP 257](https://peps.python.org/pep-0257/): 

  _"String literals occurring elsewhere in Python code may also act as documentation. They are not recognized by the Python bytecode compiler and are not accessible as runtime object attributes (i.e. not assigned to \_\_doc\_\_), but two types of extra docstrings may be extracted by software tools:_
    
  1. _String literals occurring immediately after a simple assignment at the top level of a module, class, or \_\_init\_\_ method are called “attribute docstrings”."_

@jdeschamps, opinions?

jdeschamps:
Can you fix the pre-commit doc errors while you are at it (they are super small)?

```
+-------------------------------------+----------------------------------------------------+---------+-------------------------------------------------+
| file                                | item                                               | check   | description                                     |
+=====================================+====================================================+=========+=================================================+
| src/careamics/config/nm_model.py:66 | nm_model.GaussianMixtureNMConfig.validate_path_to_ | PR02    | Unknown parameters {'self'}                     |
|                                     | pretrained_vs_data_to_train                        |         |                                                 |
+-------------------------------------+----------------------------------------------------+---------+-------------------------------------------------+
| src/careamics/config/nm_model.py:66 | nm_model.GaussianMixtureNMConfig.validate_path_to_ | PR09    | Parameter "self" description should finish with |
|                                     | pretrained_vs_data_to_train                        |         | "."                                             |
+-------------------------------------+----------------------------------------------------+---------+-------------------------------------------------+
| src/careamics/config/nm_model.py:66 | nm_model.GaussianMixtureNMConfig.validate_path_to_ | RT01    | No Returns section found                        |
|                                     | pretrained_vs_data_to_train                        |         |                                                 |
+-------------------------------------+----------------------------------------------------+---------+-------------------------------------------------+
```

tests/conftest.py
```diff
@@ -87,7 +87,10 @@ def minimum_algorithm_musplit() -> dict:
             "multiscale_count": 4,
             "predict_logvar": "pixelwise",
         },
-        "likelihood": {"type": "GaussianLikelihoodConfig", "color_channels": 2}, # TODO this is incorrect
+        "likelihood": {
+            "type": "GaussianLikelihoodConfig",
+            "color_channels": 2,
+        },  # TODO this is incorrect
```
jdeschamps: In case you wrote that `TODO`, could you add more details as why this is incorrect? Otherwise for anyone else it will be more complicated to figure it out.

federico-carrara: It was me, I'll update that :)

src/careamics/config/nm_model.py
```diff
@@ -16,21 +16,19 @@ class GaussianMixtureNMConfig(BaseModel):
     model_config = ConfigDict(
         validate_assignment=True, arbitrary_types_allowed=True, extra="allow"
     )
-    # TODO What are all these parameters?
     # model type
     model_type: Literal["GaussianMixtureNoiseModel"]
 
-    path: Optional[Union[Path, str]] = None
```
jdeschamps: Was moving the docstring before the attributes declaration done automatically or manually?

According to [PEP257](https://peps.python.org/pep-0257/#what-is-a-docstring) (and I'll grant you that it is really old..): 

> String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called “attribute docstrings”.

And that's (I believe) is the reason the class attributes/fields have the docstring after and not before the declaration/assignment.

Now this needs to be respected in order for Pydantic to extract it correctly and make it available to users when exporting json schema or just exploring the automatically generated description of each model and attribute (see [this](https://github.com/pydantic/pydantic/pull/6563)).

tests/models/lvae/test_noise_model.py
```diff
@@ -106,9 +101,14 @@ def test_noise_model_likelihood(
 @pytest.mark.parametrize("img_size", [64, 128])
 @pytest.mark.parametrize("target_ch", [1, 3, 5])
 def test_multi_channel_noise_model_likelihood(
-    tmp_path: Path, n_gaussians: int, n_coeffs: int, img_size: int, target_ch: int
+    tmp_path: Path,
+    n_gaussians: int,
+    n_coeffs: int,
```
federico-carrara: I like the idea of moving `create_dummy_noise_model` to `conftest.py` as a `pytest.fixture`. 
However, now it doesn't accept the parameters `n_gaussians` and `n_coeffs` anymore. 
Now, this is not a big issue, as changing these parameters is not gonna break the code. Therefore, I am ok with not using them, but I'd like to have them removed completely from the code if we decide to go this way.

federico-carrara (approved):

PR #211, opened 2024-08-21
Feature: LVAE prediction compatible with CAREamics tiling pipeline

melisande-c:
### Description

- **What**: Make LVAE prediction functions compatible with CAREamics tiling pipeline.
- **Why**: The aim is to reimplement the current LVAE prediction functions as the Lightning `predict_step` method. This means they should be compatible with the existing CAREamics tiling.
- **How**: Allow inputs to the prediction functions to have auxiliary inputs (i.e. inputs are tuples with the first element being the input tensor, for tiled inputs the second element is a `TileInformation` object). This simulates what the `batch` argument will be in the `predict_step` method.

### Changes Made

- **Added**: Smoke test of LVAE prediction with CAREamics tiling pipeline in pure Pytorch. Following #207 and a refactor of `InMemoryTiledPredDataset`, this can be updated to allow for testing the two different tiling strategies.
- **Modified**: 
  - The LVAE prediction functions have been modified to accept tuples as input, and they have also been renamed to reflect that they are intended to predict on a tiled batch.
  - The stitch prediction function had to be modified to deal with when there is a mismatch between the number of input and target channels. Currently the fix is a bit ugly and really the `array_shape` attribute in `TileInformation` should be updated with the correct channels before being passed to stitch prediction, but that would probably take a fair bit of refactoring.

- [x] Pre-commit passes


ashesh-0 (approved):
Well, this looks good to me. We had discussed and therefore are in sync with the required functionality.

src/careamics/prediction_utils/lvae_prediction.py
```diff
@@ -90,6 +150,7 @@ def lvae_predict_mmse(
         # store sample predictions
         sample_predictions[mmse_idx, ...] = sample_prediction
 
-    # take the mean of the sample predictions
     mmse_prediction = torch.mean(sample_predictions, dim=0)
```
ashesh-0, line 154: We will also need pixelwise variance for calibration estimation. `torch.std(sample_predictions, dim=0)` is needed along with mmse_prediction.

melisande-c: Ah I see, ok, I will find a way to add it in

melisande-c: @ashesh-0 so the pixel-wise standard deviation is only needed for calibration? and will the standard deviation tiles also need to be stitched together?

ashesh-0: Currently I work with stiched std and stiched prediction. however, that is not necessary. but what is necessary is that we remove the correct overlap from the tiles. so, in short,  if it is not that much pain, then tiled std would be better but i can work with tiles. 
and yes, pixel-wise std is needed just for calibration. but then it is a property of mmse. so whenever we output mmse prediction, we should also output mmse std.

melisande-c: ok makes sense, thanks

src/careamics/prediction_utils/stitch_prediction.py
```diff
@@ -76,8 +76,22 @@ def stitch_prediction_single(
...
-    input_shape = (1, *tile_infos[0].array_shape)  # add S dim
+    # TODO: this is hacky... need a better way to deal with when input channels and
+    #   target channels do not match
+    if len(tile_infos[0].array_shape) == 4:
+        # 4 dimensions => 3 spatial dimensions so -4 is channel dimension
+        tile_channels = tiles[0].shape[-4]
+    elif len(tile_infos[0].array_shape) == 3:
+        # 3 dimensions => 2 spatial dimensions so -3 is channel dimension
+        tile_channels = tiles[0].shape[-3]
+    else:
```
CatEek, line 87: I guess this could be simplified to (at least) )something like 
` if len(tile_infos[0].array_shape) >=3:
    tile_channels = tiles[0].shape[-len(tile_infos[0].array_shape)] 
   else:
       error `
tile_channels = tiles[0].shape[

melisande-c: `TileInformation.array_shape` is a tuple so it does not support element assignment. In any case this should not happen here and the array shape should be correct before reaching this function. This should be fixed in the dataset classes, at some point later on.

PR #215, opened 2024-08-22
Masked percentage value upper limit 

CatEek:
The value was set to 1, which is not exactly correct

### DIFF
--- a/src/careamics/config/transformations/n2v_manipulate_model.py
+++ b/src/careamics/config/transformations/n2v_manipulate_model.py
@@ -33,7 +33,7 @@ class N2VManipulateModel(TransformModel):
 
     name: Literal["N2VManipulate"] = "N2VManipulate"
     roi_size: int = Field(default=11, ge=3, le=21)
-    masked_pixel_percentage: float = Field(default=0.2, ge=0.05, le=1.0)
+    masked_pixel_percentage: float = Field(default=0.2, ge=0.05, le=10.0)
     strategy: Literal["uniform", "median"] = Field(default="uniform")
     struct_mask_axis: Literal["horizontal", "vertical", "none"] = Field(default="none")
     struct_mask_span: int = Field(default=5, ge=3, le=15)


jdeschamps (approved):

src/careamics/config/transformations/n2v_manipulate_model.py
```diff
@@ -33,7 +33,7 @@ class N2VManipulateModel(TransformModel):
 
     name: Literal["N2VManipulate"] = "N2VManipulate"
     roi_size: int = Field(default=11, ge=3, le=21)
-    masked_pixel_percentage: float = Field(default=0.2, ge=0.05, le=1.0)
+    masked_pixel_percentage: float = Field(default=0.2, ge=0.05, le=10.0)
```
melisande-c, line 36: At first I thought this wrong and the value is interpreted as a ratio, even though it's called percentage, i.e. the default value is 0.2 which would be equal to 20% and the upper bound is 1.0 which would be equal to 100% (not necessarily correct either, but 10.0 would also be wrong here). But I followed the value through to the line below and from I can see it is being interpreted as being between (0, 100]... We should probably investigate this further...

https://github.com/CAREamics/careamics/blob/840867237342ea89dbe51b1feafacd63eb78d537/src/careamics/transforms/pixel_manipulation.py#L142C2-L144C6

melisande-c: Actually ignore me, it's too late

melisande-c (approved):

src/careamics/config/transformations/n2v_manipulate_model.py
jdeschamps: I was thinking the same originally (as in a veeeery long time ago) which is why I set the upper limit to `1.0`, but it is for sure a percentage.

For shape `(64, 64)`, a percentage of `0.2` (default) will mask about 8 pixels, which is correct and what we observe if we run `N2VManipulate`. If it were a ratio, then it would be by default 20% and that's waaaay too many pixels!

melisande-c: yeah I realised after I wrote the comment

PR #216, opened 2024-08-23
Making functions in `utils/metrics.py` compatible with `torch.Tensor` inputs

federico-carrara:
### Description

This PR aims to make all the functions in `utils/metric.py` `torch`-compatible.

- **What**: Made all functions for computing metrics compatible with `torch.Tensor` inputs.
- **Why**: Most of these metrics are used either during model validation or evaluation, hence they need to handle `torch.Tensor`'s. Therefore, having `torch` compatible metrics readily available avoids cumbersome type conversion within the NN models. 
- **How**: Replaced function calls explicitly requesting `numpy` (e.g., `np.max(arr)`) with more flexible method calls (e.g., `arr.max()`).

### Changes Made

- **Added**: `torch` requires tensors to be of type `double` (i.e., `torch.float64`) to compute things like mean and std. Therefore, I added a private method that casts `torch.tensor`'s to `double` to enable `mean()` and `std()` computation. Casting to `numpy` happens instead in the PSNR computation that uses `skimage` module, since the module explicitly requires that. Finally, added notebook to show current issues with the tests.
- **Modified**: replaced function calls with method calls.
- **Removed**: Nothing.

###

---

- [ ] Pre-commit passes


src/careamics/utils/metrics.py
```diff
@@ -32,84 +38,129 @@ def psnr(gt: np.ndarray, pred: np.ndarray, range: float = 255.0) -> float:
...
+    x : Array
+        Input tensor.
+
+    Returns
+    -------
+    Array
+        Float tensor.
+    """
+    if isinstance(x, torch.Tensor) and x.dtype != torch.float64:
+        warn(
```
ashesh-0, line 161: Is the warning necessary? I ask because it will never lead to any information loss. Also, we do not return this tensor; we just use it to compute the metrics.

src/careamics/utils/metrics.py
```diff
@@ -32,84 +38,129 @@ def psnr(gt: np.ndarray, pred: np.ndarray, range: float = 255.0) -> float:
...
         Predicted image.
 
     Returns
     -------
     Union[float, torch.tensor]
         Scale invariant PSNR value.
     """
-    range_parameter = (np.max(gt) - np.min(gt)) / np.std(gt)
-    gt_ = _zero_mean(gt) / np.std(gt)
+    # cast tensors to double dtype
```
ashesh-0, line 137: logic looks good.

ashesh-0 (approved):

src/careamics/utils/metrics.py
federico-carrara: I'd definitely remove it "in production", but I have temporarily put it there for debugging :)

jdeschamps:
What is the status of this PR?

federico-carrara:
I remember there was a problem with dtypes changing from `numpy` to `torch` (at least this is what I have in my notes). Unless this is a feature we want to have soon, I'll just close the PR!

jdeschamps:
Let's leave it for now!

PR #217, opened 2024-08-23
Feature: Function that computes CAREamics `TileInformation` with `GridIndexManager`

melisande-c:
### Description

- **What**: Add function that creates tile information object using the `GridIndexManager`.
- **Why**: It might be easier to use this function in the prediction pipeline if the different tiling approaches will be used, in particular the `ShiftBoundary` tile strategy.
- **How**: Almost duplicates the first `compute_tile_info` function I wrote, just uses the data stored in `GridIndexManager` and uses the `GridIndexManager.get_location_from_dataset_idx` method.

### Changes Made

- **Added**: `compute_tile_info_legacy` to `careamics/src/careamics/dataset/tiling/lvae_tiled_patching.py`
- **Modified**: Renamed variable `tile_grid_coords` to `tile_grid_indices`. This is to try to prevent misinterpretation. 

The `TilingMode` enum in the Disentangle repo is not in the current version of the dataset in CAREamics. 

Note: logic comes from here, mostly only variables have been renamed to match CAREamics naming scheme. https://github.com/ashesh-0/Disentangle/blob/ed99b2614a9e52b496947ccf157d3aaa8db52872/disentangle/analysis/stitch_prediction.py#L37-L66

---

- [x] Pre-commit passes


src/careamics/dataset/tiling/lvae_tiled_patching.py
```diff
@@ -93,19 +96,98 @@ def extract_tiles(
...
+    -------
+    TileInformation
+        Information that describes how to crop and stitch a tile to create a full image.
+
+    Raises
+    ------
+    ValueError
+        If `grid_index_manager.data_shape` does not have 4 or 5 dimensions.
+    """
+    data_shape = np.array(grid_index_manager.data_shape)
```
ashesh-0, line 122: I think this might just work. however, it is worth noting that this can be a potential issue. Let me explain. In some cases, where we just have the input channel, data_shape will have 1 as the channel dimension but we will be predicting >2 channels. So it made sense to pick the channel dimension from the prediction. In this case, where there is no prediction, one could pass the information about how many channels are getting predicted.

src/careamics/dataset/tiling/lvae_tiled_patching.py
```diff
@@ -93,19 +96,98 @@ def extract_tiles(
...
+    tile_coords_start = stitch_coords_start - grid_index_manager.patch_offset()
+
+    # --- replace out of bounds indices
+    out_of_lower_bound = stitch_coords_start < 0
+    out_of_upper_bound = stitch_coords_end > data_shape
+    stitch_coords_start[out_of_lower_bound] = 0
+    stitch_coords_end[out_of_upper_bound] = data_shape[out_of_upper_bound]
+
+    # TODO: TilingMode not in current version
+    # if grid_index_manager.tiling_mode == TilingMode.ShiftBoundary:
```
ashesh-0, line 144: I see this as the default tiling mode. this is what you have in CAREamics: shifting the boundary tile such that you get prediction for the full frame without using any padding.

ashesh-0 (approved):

src/careamics/dataset/tiling/lvae_tiled_patching.py
melisande-c: I have already encountered this issue, and I hacked a fix into the careamics stitch prediction function [here](https://github.com/CAREamics/careamics/blob/427fa268a4b6a0cd1db445834bf440e541dda1f3/src/careamics/prediction_utils/stitch_prediction.py#L79-L94), in a previous PR. It's not very elegant but careamics might have to deal with this in other applications so I will leave it like this for now.

src/careamics/dataset/tiling/lvae_tiled_patching.py
melisande-c: yes, for the default case the previous `compute_tile_info` [function](https://github.com/CAREamics/careamics/blob/a94b36aad1027110151d60fc60e5271dfd504fcf/src/careamics/dataset/tiling/lvae_tiled_patching.py#L178-L184) I wrote should be sufficient, but I have mostly added this function for the other cases. When #223 is merged the `TilingMode` enum will be included.

PR #218, opened 2024-08-23
Implemented `training_step` and `validation_step`, fixed all the configs needed to instantiate the lightning module 

federico-carrara:
### Description

This PR finally implements a working training loop for the lightining module `VAEModule`.

- **What**: we added `training_step` and `validation_step` plus all the required methods. We also fixed some bugs in the configs (`pydantic` model) describing the different model components.
- **Why**: To be able to train a LVAE model.
- **How**: see below.

### Changes Made

- **Added**: 
  - `training_step` and `validation_step` in the lightning `VAEModule` (plus other related methods). 
  - Some required attributes and validators in the `VAEAlgorithmmConfig`.
  - An example notebook to illustrate how to train the LVAE model using current lightning API. 
  - `RunningPSNR` class in `metrics.py`.
  - Tests for the added features.
- **Modified**:
  - Typing and validators in different `pydantic` models (e.g., `lvae_model.py`, `nm_model.py`) 
- **Removed**: Nothing in particular.

### Things that can be improved

- `validation_step` requires some additional methods for computing the PSNR at each epoch. These methods are currently within the `VAEModule` since require some of its attributes. If possible, it would be nice to move this methods out of the lightning module to have something cleaner.

---

- [x] Pre-commit passes


src/careamics/config/vae_algorithm_model.py
```diff
@@ -66,9 +72,10 @@ class VAEAlgorithmConfig(BaseModel):
     #   - values can still be passed as strings and they will be cast to Enum
     algorithm_type: Literal["vae"]
     algorithm: Literal["musplit", "denoisplit", "custom"]
-    loss: Literal["musplit", "denoisplit"]
+    loss: Literal["musplit", "denoisplit", "denoisplit_musplit"]
```
CatEek, line 75: Why do we need a separate loss for denoisplit_musplit ?

src/careamics/lightning/lightning_module.py
```diff
@@ -315,9 +332,10 @@ def training_step(self, batch: Tensor, batch_idx: Any) -> Any:
...
         """
-        x, *aux = batch  # TODO: check what is a `batch`
+        x, target = batch
+
+        # Forward pass
         out = self.model(x)
-        target = aux[0]
 
         # Update loss parameters
         # TODO rethink loss parameters
```
CatEek, line 341: loss parameters is probably the abomination I created and probably we need to rethink it, but later

src/careamics/lightning/lightning_module.py
```diff
@@ -420,7 +456,99 @@ def configure_optimizers(self) -> Any:
...
+        # update running psnr
+        for i in range(out_channels):
+            self.running_psnr[i].update(rec=recons_img[:, i], tar=target[:, i])
+
+        # compute psnr for each channel in the current batch
+        # TODO: this doesn't need do be a method of this class
+        # and hence can be moved to a separate module
+        return [
+            psnr_func(
+                gt=target[:, i].clone().detach().cpu().numpy(),
```
CatEek, line 522: That's a minor thing and not urgent, but if we compute metrics during train/val step maybe we should just do it on tensors ? Moving data around takes time

tests/models/test_model_factory.py
```diff
@@ -54,6 +54,8 @@ def forward(self, input):
     assert model.out_features == 5
 
 
+# TODO: add test for LVAE
+@pytest.mark.skip(reason="This test is outdated. LVAE is now supported.")
```
CatEek: Maybe we can remove the skip then ?

src/careamics/config/vae_algorithm_model.py
```diff
@@ -94,23 +101,63 @@ def algorithm_cross_validation(self: Self) -> Self:
...
+                and self.model.predict_logvar is not None
+            ):
                 raise ValueError(
-                    "Algorithm `denoisplit` only supports `predict_logvar` as `None`."
+                    "Algorithm `denoisplit` with loss `denoisplit` only supports "
+                    "`predict_logvar` as `None`."
                 )
 
-        # TODO: what if algorithm is not musplit or denoisplit
+        # TODO: what if algorithm is not musplit or denoisplit (HDN?)
```
ashesh-0, line 124: HDN has denoiSplit's loss: noise model enabled with log_var =None

src/careamics/lightning/lightning_module.py
```diff
@@ -326,31 +344,49 @@ def training_step(self, batch: Tensor, batch_idx: Any) -> Any:
         # Compute loss
         loss = self.loss_func(out, target, self.loss_parameters)  # TODO ugly ?
 
-        self.log_dict(loss, on_step=True, on_epoch=True, prog_bar=True, logger=True)
+        # Logging
+        # TODO: implement a separate logging method?
+        self.log_dict(loss, on_step=True, on_epoch=True)
+        # self.log("lr", self, on_epoch=True)
```
ashesh-0, line 350: I donot see a check for NaN. That is important.  https://github.com/ashesh-0/Disentangle/blob/master/disentangle/nets/lvae.py#L856
If it is there at some other place, then all is good.

src/careamics/utils/metrics.py
```diff
@@ -4,7 +4,7 @@
 This module contains various metrics and a metrics tracking class.
 """
 
-from typing import Union
+from typing import Optional, Union
```
ashesh-0, line 7: At some point , https://github.com/juglab/MicroSSIM?tab=readme-ov-file#usage needs to be added here. In evaluation I use MicroSSIM, MicroS3IM and range invariant PSNR. Not for now though.

ashesh-0 (commented):
All looks good. Just add the check for Nan's. I don't see that in training_step()

src/careamics/lightning/lightning_module.py
ashesh-0: @federico-carrara mentioned that bit is in the loss computation code and so, nothing to be done now.

ashesh-0 (approved):
All good.

src/careamics/config/vae_algorithm_model.py
federico-carrara: I see. Anyway, we'll live HDN for later refactoring, as it most likely require different configs and different naming

src/careamics/lightning/lightning_module.py
federico-carrara: Exactly. Specifically, it happens here: https://github.com/federico-carrara/careamics/blob/83583aa4803466fe79149f9ecb25eb4e89f961dc/src/careamics/losses/lvae/losses.py#L367 (and similarly for other losses)

src/careamics/utils/metrics.py
federico-carrara: Good point. Just a small question @ashesh-0: do you just need it at evaluation time or also for validation?

ashesh-0: It is only needed at evaluation time.

tests/models/test_model_factory.py
federico-carrara: You mean remove the test? That I agree, but I wanted to ask first.

src/careamics/lightning/lightning_module.py
federico-carrara: The current issue is that all metrics support only `np.ndarray` inputs. I tried to make them compatible to `torch` too (check #216), but there are a few things to solve there. I'll surely agree with you though.

src/careamics/config/vae_algorithm_model.py
federico-carrara: If you look at the code in `losses` you'll see that the loss in that case is pretty much different from the "usual" `denoiSplit` case. That's why we need it.

CatEek: No hard opinions, maybe we can revisit this later

tests/models/test_model_factory.py
CatEek: Either remove the test or the decorator

CatEek (approved):

PR #221, opened 2024-08-23
Add custom algorithm to `FCNAlgorithmConfig`

jdeschamps:
### Description

During the split of the `AlgorithmConfig`, the `custom` algorithm disappeared. But it is used to allow `custom` models to not implement the `is_3D` method. 

https://github.com/CAREamics/careamics/blob/29bc2bf066e73d53741609fbf444c4372728c1f7/src/careamics/config/configuration_model.py#L223

Whether this is a good thing to have or not, it is currently a bug. In the future, we might want to kick this out altogether, have a way to enforce a certain contract with custom model classes, or keep it that way.

- **What**: Enable custom algorithms in the config.
- **Why**: Mechanism is used to guard against a non implemented method in custom models.
- **How**: Added `custom` to the literal in `FCNAlgorithmConfig`

### Modifications

- Added attribute docstrings to `fcn_algorithm_model.py`, `supported_architectures.py` and `supported_algorithms.py`
- Added custom model as fixture in `conftest.py`
- Added `custom` to the literal in `FCNAlgorithmConfig`
- Added tests to `test_custom_model.py` and `test_fc_algorithm_config.py`
- Disabled a test in `test_register_model.py` as it would cause failure of subsequent custom model tests by wiping out the registry of models.

### Related Issues

This point was raised in a previous PR that was closed, while the PR's commit made it into another PR...

This will allow some of the existing examples in the doc to pass again.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/architectures/custom_model.py
+++ b/src/careamics/config/architectures/custom_model.py
@@ -76,7 +76,6 @@ class CustomModel(ArchitectureModel):
     architecture: Literal["custom"]
     """Name of the architecture."""
 
-    # name of the custom model
     name: str
     """Name of the custom model."""
 
--- a/src/careamics/config/fcn_algorithm_model.py
+++ b/src/careamics/config/fcn_algorithm_model.py
@@ -67,7 +67,7 @@ class FCNAlgorithmConfig(BaseModel):
     # Mandatory fields
     # defined in SupportedAlgorithm
     algorithm_type: Literal["fcn"]
-    algorithm: Literal["n2v", "care", "n2n"]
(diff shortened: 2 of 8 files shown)


melisande-c:
Think some tests might have got removed too. Here's your previous commit https://github.com/CAREamics/careamics/pull/180/commits/9aa978baf74b3e4ab1de408255ed4616cb70fa9c.

jdeschamps:
Oh very good catch, totally forgot about that!

federico-carrara (approved):

PR #222, opened 2024-08-24
Simplify `AlgorithmConfig` discrimination

jdeschamps:
### Description

Currently, the choice between `FCNAlgorithmConfig` and `VAEAlgorithmConfig` is done through the newly introduced `algorithm_type` attribute (`fcn` vs `vae`). This attribute has no other goal.

In the meantime, the `Literal` for the `algorithm` attribute in both algorithm config are non overlapping (currently `n2v, care, n2n` vs `musplit, denoisplit, custom`). That mean that instead of a useless additional parameter polluting the code, we could simply use the `algorithm` as discriminator.

The one thing that could prevent it is the use of `custom` on each side (there is another PR to add back `custom` https://github.com/CAREamics/careamics/pull/221). Here I decided to remove the `custom` from the `vae` side, as we are far from being able to support a custom model there. 

The idea behind the custom is to allow people to simply plug their own model into the established careamics pipeline. This is going to be hard in the vae case, as there will be many assumptions about the model itself that custom models would need to implement.

I suggest we wait until the vae is fully in careamics before allowing custom ones. If we do want the custom mechanism, then we might need to have a `custom_fcn` (if the name sticks...) and `custom_vae`.

- **What**: Simplify the discrimination of `FCNAlgorithmConfig` and `VAEAlgorithmConfig` in the main configuration.
- **Why**: The attribute `algorithm_type` was no useful beyond the discrimination, which can be achieved using the `algorithm` attribute.
- **How**: Removed `algorithm_type` throughout the code, change the discriminator to `algorithm`.

### Changes Made

Removed `algorithm_type` throughout the code, change the discriminator to `algorithm` in `Configuration`.

### Breaking changes

Any code using `algorithm_type` attribute.

It will interact with any other PR that uses the CAREamics code (as in, not the LVAE part).

Since the vae config is still on-going work (see noise models), there is no test for whether the mechanism actually works in careamics!

Here is a simple toy example to illustrate how it works:

```python
from pydantic import BaseModel, Field
from typing import Literal, Union

class Latin(BaseModel):
    greeting: Literal["ola", "ciao", "bonjour"]
    cold_weather: bool = False

class Nordic(BaseModel):
    greeting: Literal["guten tag", "hej"]
    cold_weather: bool = True


class Country(BaseModel):    
    type: Union[Latin, Nordic] = Field(..., discriminator="greeting")


a = Country(type={"greeting": "bonjour"})
assert a.type.cold_weather == False

b = Country(type={"greeting": "hej"})
assert b.type.cold_weather == True
```

---

- [x] Pre-commit passes


CatEek (approved):
No hard opinions, if I remember correctly, the presence of custom was the reason this parameter appeared.

PR #223, opened 2024-08-24
LVAE dataset: add MultiFileDset

veegalinova:
### Description

- **What**: This is a follow-up on updating the datasets from the Disentangle repo (reference commit `57cad67` from the main branch), adding the `MultiFileDset` class, `TilingMode` and  `EmptyPatchFetcher`
- **Why**: The `MultiFileDset` is necessary for the datasets in the paper
- **How**: Copy/paste the code from the Disentangle repo with minimal changes

### Changes Made

- **Added**: 
  - from the Disentangle repo:
    - copied `MultiFileDset` class and related classes
    - added `TilingMode` and updated related code in `MultiChDloader`
    - copied `EmptyPatchFetcher`
    - copied several new parameters for `MultiChDloader`
    - copied code for flips (using _Albumentations_)
    - added "load_data_fn" parameter to dataset classes to pass the data loading logic from different experiments
  - added lists of the dataset types needed for the paper in the `data_utils.py`
  - added a simple test to test the `MultiFileDset` object initialization
- **Modified**: 
  - moved dataset configs into `lvae_training/dataset/configs` 
  - moved `GridIndexManager` and `IndexSwitcher` to separate files in `lvae_training/dataset/utils` 
- **Removed**: 
  - removed code related to the BioSR dataset from `data_utils.py` and tests

---

- [ ] Pre-commit passes


src/careamics/lvae_training/dataset/utils/data_utils.py
```diff
@@ -0,0 +1,197 @@
...
+import numpy as np
+from skimage.io import imread, imsave
+
+from careamics.lvae_training.dataset.configs.vae_data_config import (
+    DataSplitType,
+    DataType,
+)
+
+Multifile_Datasets = [
+    "Pavia3SeqData",
```
ashesh-0: Other things are okay. but this cannot work.  For example Pavia3SeqData has 9 different dataset. I do not see them anywhere. Similar things could be said about Nicola dataset. 
This classification between singlefile and multifile is also not correct. `OptiMEM100_014`, for example is a single file dataset. 

I would copy the rawdata loader codes for the different dataset. If this makes sense, you can have a submodule in ceramics which handles the data loading specifications for this work. this would allow you to not mix that much code into caremics. so, in careamics, you will have the basic code needed to run the different models. but the raw data loader part, which is tied to the paper, can live elsewhere.

ashesh-0 (changes requested):

src/careamics/lvae_training/dataset/vae_dataset.py
```diff
@@ -177,9 +173,9 @@ def __init__(
 
         self._rotation_transform = None
         if self._enable_rotation:
-            raise NotImplementedError(
-                "Augmentation by means of rotation is not supported yet."
-            )
+            # TODO: fix this import
+            import albumentations as A
+
             self._rotation_transform = A.Compose([A.Flip(), A.RandomRotate90()])
```
CatEek, line 177: I guess flip/rotate can be safely replaced by existing funcs

veegalinova: There is custom logic for flips and rotations in 3D, I'm not sure yet how compatible it is with the existing transforms

jdeschamps: Are flips and rotations applied in 3D?

I never saw a verification, but PSFs being usually not symmetrical in 3D, this is potentially bad. If data is limiting, that might help, but otherwise that might introduce issues.

We could easily extend the Flip and Rot90 to Z, with the right parameters to turn them off. This would be better than having Albumentations back in.

jdeschamps: Heard from Ashesh that he is not using z-flips or rotations (which harm the training), so we can just use the CAREamics transformations directly instead of the Albumentations. (at least it should work right away hopefully)

src/careamics/lvae_training/dataset/utils/data_utils.py
veegalinova: I added the parameter `load_data_fn`, where `get_train_val_data` functions can be passed from different experiments. 
I also manually went through all the python files for the dataset list that you gave, things we are missing now for the experiments to run:
1) `multifile_raw_dloader.py` file, used by 3 datasets, which I think should exist together with the code of the experiments
2)  dependencies: `czifile`, `nis2pyr`, `nd2`,

src/careamics/lvae_training/dataset/vae_dataset.py
veegalinova: Decided to fix it in a separate PR

src/careamics/lvae_training/dataset/multifile_dataset.py
```diff
@@ -0,0 +1,274 @@
...
+    def __init__(
+        self,
+        data_config: Union[MultiChDatasetConfig, LCDatasetConfig],
+        fpath: str,
+        load_data_fn: Callable[..., NDArray],
+        val_fraction=None,
+        test_fraction=None,
+    ):
+        self._fpath = fpath
+        data = load_data_fn(
```
ashesh-0: Want to point out the inconsistency here. this file (multifile_dataset.py) is included in CAREamics. however, load_data_fn, in the current code state, will return a class object, and the class is defined outside the CAREamics (multifile_raw_dloader.py). I have trust in you that you can easily find a suitable way to resolve this.

ashesh-0 (approved):

src/careamics/lvae_training/dataset/vae_dataset.py
melisande-c: PR here #238

PR #224, opened 2024-08-25
Bug fix NoiseModel and LVAE training

CatEek:
### Description

Fixing bugs

Training launches using examples/example_training_LVAE_split.ipynb notebook.

@ashesh-0 and @federico-carrara Could you please take a look into device placement 

### Related Issues

Some fixes are ugly af. In the likelihoods module device placement doesn't seem to be used. @ashesh-0 and @federico-carrara Could you please take a look into device placement?

bug_fixing_towards_vae_training/src/careamics/models/lvae/likelihoods.py#L311
bug_fixing_towards_vae_training/src/careamics/models/lvae/likelihoods.py#L360-L361

WIP

- [ ] Pre-commit passes

### DIFF
--- a/src/careamics/config/likelihood_model.py
+++ b/src/careamics/config/likelihood_model.py
@@ -10,8 +10,6 @@ from careamics.models.lvae.noise_models import (
     MultiChannelNoiseModel,
 )
 
-NoiseModel = Union[GaussianMixtureNoiseModel, MultiChannelNoiseModel]
-
 
 class GaussianLikelihoodConfig(BaseModel):
     """Gaussian likelihood configuration."""
@@ -39,5 +37,7 @@ class NMLikelihoodConfig(BaseModel):
     """The standard deviation of the data, used to unnormalize data for noise
     model evaluation. Shape is (target_ch,) (or (1, target_ch, [1], 1, 1))."""
 
-    noise_model: Union[NoiseModel, None] = None
+    noise_model: Union[
+        Union[GaussianMixtureNoiseModel, MultiChannelNoiseModel], None
+    ] = None
     """The noise model instance used to compute the likelihood."""
--- a/src/careamics/models/lvae/likelihoods.py
+++ b/src/careamics/models/lvae/likelihoods.py
@@ -357,7 +357,7 @@ class NoiseModelLikelihood(LikelihoodModule):
             The log-likelihood tensor. Shape is (B, C, [Z], Y, X).
         """
         self.data_mean = self.data_mean.to(x.device)
-        self.data_std = self.data_std.to(x.device) # TODO this is ugly fix !!
+        self.data_std = self.data_std.to(x.device)  # TODO this is ugly fix !!
         predicted_s_denormalized = params["mean"] * self.data_std + self.data_mean
         x_denormalized = x * self.data_std + self.data_mean
         likelihoods = self.noiseModel.likelihood(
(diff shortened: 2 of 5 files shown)


src/careamics/config/likelihood_model.py
```diff
@@ -10,8 +10,6 @@
     MultiChannelNoiseModel,
 )
 
-NoiseModel = Union[GaussianMixtureNoiseModel, MultiChannelNoiseModel]
```
federico-carrara, line 13: I'd keep this honestly unless you have a strong argument against it ;)

src/careamics/models/lvae/likelihoods.py
```diff
@@ -355,6 +356,8 @@ def log_likelihood(self, x: torch.Tensor, params: dict[str, torch.Tensor]):
         torch.Tensor
             The log-likelihood tensor. Shape is (B, C, [Z], Y, X).
         """
+        self.data_mean = self.data_mean.to(x.device)
```
federico-carrara, line 359: I guess this should be done once for all inside the `lightning_module` `__init__`. Indeed, here it would be instead called at every `training step`.

src/careamics/models/lvae/noise_models.py
```diff
@@ -76,7 +76,7 @@ def train_gm_noise_model(
     # TODO any training params ? Different channels ?
     noise_model = GaussianMixtureNoiseModel(model_config)
     # TODO revisit config unpacking
-    noise_model.train_noise_model(noise_model.signal, noise_model.observation)
+    noise_model.train_noise_model(model_config.signal, model_config.observation)
```
federico-carrara, line 79: seems fine

federico-carrara (commented):
Thanks @CatEek for spotting some tricky bugs! 
I think the "moving all tensors to same device" stuff needs our attention. Hence, I agree to sit down and discuss how to handle it in the cleanest way possible.
In general, for the moment, I'd prefer not to merge any changes regarding the "example notebook" since I am heavily changing it too. I'd instead agree to integrate the changes in `noise_models.py` and, temporarily, the one in `likelihoods.py`.
Thanks, 

Federico

jdeschamps:
> Thanks @CatEek for spotting some tricky bugs! I think the "moving all tensors to same device" stuff needs our attention. Hence, I agree to sit down and discuss how to handle it in the cleanest way possible. In general, for the moment, I'd prefer not to merge any changes regarding the "example notebook" since I am heavily changing it too. I'd instead agree to integrate the changes in `noise_models.py` and, temporarily, the one in `likelihoods.py`. Thanks,
> 
> Federico

I kinda agree, maybe we should only let @federico-carrara update the notebook to avoid the nightmare of fixing conflicts there.

If there is a PR that breaks the notebook, then either:
1. Put a code snippet with the cell numbers of what needs to be corrected in the PR body
2. Create a branch with the notebook modification alone, push it and tell @federico-carrara 
3. Make the notebook a python script and only push the script (then you can convert it back and forth from notebook if that's what you need to use)

src/careamics/config/likelihood_model.py
CatEek: No strong opinions, just wasn't working properly

CatEek:
@jdeschamps Yes, the NB here is just for reference and as a storage of (kinda) working code. @federico-carrara Uses his own internal one to actually run tests

CatEek:
But otherwise agreed, this notebook shouldn't be here at all in the end. Just a temp storage

jdeschamps:
This PR now has conflicts that should be resolved and one pre-commit error @CatEek

PR #230, opened 2024-08-28
feat: added more metrics for evaluation of Splitting results

federico-carrara:
### Description

This PR adds metrics for the evaluation of LVAE-based splitting algorithms.

- **What**: Added SSIM-like metrics and others to collect average metrics over batches of elements.
- **Why**: For a proper evaluation of their performance, splitting algorithms require more metrics than the one previously existing in `careamics`. 
- **How**: Added functions, extended some auxiliary functions to accept `torch.Tensor`s.

### Changes Made

- **Added**:  
  - Added `multiscale_ssim` and `_range_invariante_multiscale_ssim` functions.
  - Added functions to compute average metrics over batches of images.
- **Modified**:
  - Made auxiliary functions compatible with `torch.Tensor`s since SSIM-based functions accept them.
  - Added warning to `psnr()` function for the case in which `data_range` is not provided.
- **Removed**: Nothing.

Note about `psnr()` for further discussion and changes: the current PSNR computation is carried out by calling `skimage.metrics.peak_signal_to_noise_ratio`. The problem with this function is that when `data_range` is not specified (i.e., `None`), then it is automatically inferred from the data type (`np.dtype`). This can cause unexpected behavior for some data types (e.g., `float` or `int64`). This is for example the case for the PSNR computed during validation, where the employed tensors are of type `float`.

Therefore, I suggest to include here a brief code snippet that automatically computes `data_range` as `gt.max() - gt.min()` whenever a `data_range` is not provided. 

---

- [x] Pre-commit passes


src/careamics/utils/metrics.py
```diff
@@ -186,3 +204,173 @@ def get(self) -> Optional[torch.Tensor]:
             return None
         rmse = torch.sqrt(self.mse_sum / self.N)
         return 20 * torch.log10((self.max - self.min) / rmse)
+
+
+def _range_invariant_multiscale_ssim(
```
ashesh-0, line 201: I think this needs to change. this is CARE-SSIM. (I know this existed in the Disentangle repository and so you copied it..).  However, MicroSSIM, the SSIM variant I report the numbers of is https://github.com/juglab/MicroSSIM

```
from microssim import MicroSSIM, MicroMS3IM
gt: N x H x W
pred: N x H x W

ssim = MicroSSIM() # or MicroMS3IM()
ssim.fit(gt, pred)

for i in range(N):
    score = ssim.score(gt[i], pred[i])
    print('SSIM score for', i, 'th image:', score)

```
Specifically, I use the following function to compute the numbers https://github.com/ashesh-0/Disentangle/blob/master/disentangle/scripts/evaluate.py#L127

jdeschamps: Can we hold off on that until we make MicroSSIM a dependency of CAREamics? Should happen next week.

federico-carrara: Sure, not an issue as the main metric to evaluate performance is PSNR!

jdeschamps (approved):

PR #232, opened 2024-08-29
refac: refactored serialization of arrays for noise model and likelihood configs

federico-carrara:
### Description

- **What**: Implemented logic to serialize arrays defined within pydantic models (specifically with reference to `nm_model.py` and `likelihood_model.py`).
- **Why**:  Because during training we need to save info in the configs. Specifically, since arrays cannot be deserialized by default we need to decide whether to keep them or not.
- **How**: Excluded large arrays from serialization, wrote custom serializers for others.

### Changes Made

- **Added**: 
  - Custom serializer `array_to_json()`.
  - 2 different custom deserializers: `list_to_numpy()`, `list_to_torch()`.
- **Modified**: Excluded some arrays from serialization.
- **Removed**: None.

### For further discussion

NOTE1: why deserializer takes in a list? Because in our use case we need deserializers to move config files usually stored as dicts into pydantic models. But, such dicts are often the result of loading config file mainly stored as `json`, `pkl`, or `yml`. The loader for these file types automatically deserialize strings into lists. That's why what is left to do for us is to move lists into arrays or tensors.

NOTE2: why 2 different deserializers? Because in some cases we want to deserialize list to torch tensor and in some other cases we prefer numpy arrays. 

After discussing with @jdeschamps we realized that some large arrays should not be part of configs (e.g., `signal`, `observation` in `nm_model.py`). Therefore, I left TODOs to remind about discussing this for future refactoring.

---

- [x] Pre-commit passes


src/careamics/config/nm_model.py
```diff
@@ -4,8 +4,17 @@
...
+import torch
+from pydantic import BaseModel, ConfigDict, Field, PlainSerializer, model_validator
+from typing_extensions import Annotated, Self
+
+from careamics.utils.serializers import array_to_json
+
+Array = Annotated[
+    Union[np.ndarray, torch.Tensor], PlainSerializer(array_to_json, return_type=str)
+]
+"""Annotated float type, used to serialize arrays or tensors to JSON strings."""
```
federico-carrara: Typo, fix

src/careamics/utils/serializers.py
federico-carrara, line 1: Add deserializer (from `str` to `Array`)

jdeschamps:
Ping me when you have the deserializer in the PR, so we can review it!

CatEek (approved):

tests/utils/test_serializers.py
```diff
@@ -0,0 +1,63 @@
...
+
+import numpy as np
+import pytest
+import torch
+from pydantic import BaseModel, ConfigDict, PlainSerializer, PlainValidator
+from typing_extensions import Annotated
+
+from careamics.utils.serializers import array_to_json, list_to_numpy, list_to_torch
+
+Array = Annotated[
```
jdeschamps: It would be better to use the actual type that is in the code base (that way we are testing it!).

tests/utils/test_serializers.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.mark.parametrize("arr", [np.array([1, 2]), torch.tensor([1, 2])])
+def test_serialize_tensor(arr: Union[np.ndarray, torch.Tensor]):
+    """Test array_to_json function."""
+    arr_model = MyTensor(arr=arr)
+    assert arr_model.model_dump() == {"arr": "[1, 2]"}
+
+
+def test_deserialize_array():
```
jdeschamps: In order to test the deserialization, it would make more sense to me to do something like this:

``` python
arr_model = MyArray(arr=arr)
serialize_dict = arr_model.model_dump()
arr_model_deserialized = MyArray(**serialize_dict)
```

jdeschamps (changes requested):

tests/utils/test_serializers.py
federico-carrara: Mmmm but this would fail with the current deserializer... Indeed, I explicitly designed it to deal with the configs loaded from `json`, `pkl` or `yml` files. In that case, when you load the config, all the list-like strings (e.g., `'[1, 2]'`) gets automatically loaded as lists. Therefore, the deserialization we need in practice is from list to array/tensor and not from str to array/tensor.

federico-carrara: One possible solution is to add in the deserializer a conditional that checks whether the input is str (and so it first converts it in a list) or directly a list.

jdeschamps (approved):

PR #236, opened 2024-09-02
fix: making range parameter mandatory in `psnr()`

federico-carrara:
### Description

- **What**: We force user to provide a `data_range` as input to `psnr()` function.
- **Why**: out `psnr()` function is based on `peak_signal_noise_ratio` function from `skimage`. This function infers the data range from the data type (e.g., 255 if data in `np.uint8`), producing weird results in some cases.
- **How**: If `data_range` is not provided raise a `ValueError`.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/utils/metrics.py
+++ b/src/careamics/utils/metrics.py
@@ -11,28 +11,33 @@ import torch
 from skimage.metrics import peak_signal_noise_ratio
 
 
-def psnr(gt: np.ndarray, pred: np.ndarray, range: float = 255.0) -> float:
-    """
-    Peak Signal to Noise Ratio.
-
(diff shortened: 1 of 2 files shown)


src/careamics/utils/metrics.py
```diff
@@ -11,28 +11,32 @@
...
+
     Parameters
     ----------
-    gt : NumPy array
-        Ground truth image.
-    pred : NumPy array
-        Predicted image.
-    range : float, optional
-        The images pixel range, by default 255.0.
+    gt : np.ndarray
```
jdeschamps, line 25: Very small stuff, it would be better to have the type be `numpy.ndarray` to be more explicit.

jdeschamps (approved):

PR #238, opened 2024-09-05
Feature: Transform multiple arrays identically to `patch` and `target`

melisande-c:
### Description

- **What**: Add option in `XYFilp` and `XYRandomRotate90` to transform additional arrays identically to the patch and target. Include the ability to compose these transforms. Also fixed some type hint issues, but some remain.
- **Why**: The LVAE training requires the identical transformation of multiple arrays, including images and noise arrays. Previously it was implemented with the `Albumentations` library. Original motivation came from this comment chain https://github.com/CAREamics/careamics/pull/223#discussion_r1733656634.
- **How**: A few options were considered, but ultimately I went with what I thought would require the least amount of the code base to change. Additional arrays can passed to the transform `__call__` function as kwargs, and they will be returned as a dict after the augmented patch and target. Composition of transforms with additional arrays has been kept as a separate method, `Compose.transform_with_additional_arrays`, mostly to not have to change too much of the code base, this can be revised and integrated with `Compose.__call__` later if desired.

### Changes Made

- **Added**: 
  - Tests for the new functionality.
  - `_chain_transforms_additional_arrays` and `transform_with_additional_arrays` methods to `Compose`.
- **Modified**: 
  -  add `**additional_arrays: NDArray` argument to the `__call__` method of some transforms.
  - A few existing tests that needed to modify the unpacking of transform outputs, now that they output a tuple of 3 rather than 2.

### Breaking changes

Code calling transforms directly, the output is now a tuple of 3 rather than 2.

#### Note

One of the biggest pain points that hasn't been fully resolved is that the `N2VManipulate` transform doesn't have the same function signature in the `__call__` method as the other transforms, it returns 3 arrays rather than 2. We know it gets applied last in the algorithm but we could make it more explicit somehow, this would resolve some annoying type hinting.

#### Converting LVAE transform code

https://github.com/CAREamics/careamics/blob/b008187292f269455272c92c9009d45a330e7ca0/src/careamics/lvae_training/dataset/vae_dataset.py#L867-L881

To convert the code snippet above, I think (not tested) the following can be done, assuming `self.rotation_transform` is created using the CAREamics `Compose` instead of that from Albumentations:

```python
img_kwargs = {}
for i, img in enumerate(img_tuples):
    for k in range(len(img)):
        img_kwargs[f"img{i}_{k}"] = img[k]

noise_kwargs = {}
for i, nimg in enumerate(noise_tuples):
    for k in range(len(nimg)):
        noise_kwargs[f"noise{i}_{k}"] = nimg[k]

img, _, rot_dic = self._rotation_transform.transform_with_additional_arrays(
    img_tuples[0][0], **img_kwargs, **noise_kwargs
)
```

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/transforms/n2v_manipulate.py
+++ b/src/careamics/transforms/n2v_manipulate.py
@@ -3,6 +3,7 @@
 from typing import Any, Literal, Optional, Tuple
 
 import numpy as np
+from numpy.typing import NDArray
 
 from careamics.config.support import SupportedPixelManipulation, SupportedStructAxis
 from careamics.transforms.transform import Transform
@@ -98,8 +99,8 @@ class N2VManipulate(Transform):
         self.rng = np.random.default_rng(seed=seed)
 
     def __call__(
-        self, patch: np.ndarray, *args: Any, **kwargs: Any
-    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+        self, patch: NDArray, *args: Any, **kwargs: Any
+    ) -> Tuple[NDArray, NDArray, NDArray]:
         """Apply the transform to the image.
 
         Parameters
@@ -142,5 +143,8 @@ class N2VManipulate(Transform):
         else:
             raise ValueError(f"Unknown masking strategy ({self.strategy}).")
 
+        # TODO: Output does not match other transforms, how to resolve?
+        #     - Don't include in Compose and apply after if algorithm is N2V?
+        #     - or just don't return patch? but then mask is in the target position
         # TODO why return patch?
         return masked, patch, mask
--- a/src/careamics/transforms/xy_random_rotate90.py
+++ b/src/careamics/transforms/xy_random_rotate90.py
@@ -3,6 +3,7 @@
 from typing import Optional, Tuple
 
 import numpy as np
+from numpy.typing import NDArray
 
 from careamics.transforms.transform import Transform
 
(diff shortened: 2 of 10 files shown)


jdeschamps:
Yes, the N2VManipulate is annoying. We could pack some of the return arrays into one... Dirty and not clear...

Another way would be to inspect again the loss. If I remember it there is a reason to return the three arguments, but maybe we can rewrite it to only need `mask * original_patch`. (Unfortunately it probably needs the original patch for something).

src/careamics/transforms/xy_random_rotate90.py
```diff
@@ -49,8 +50,11 @@ def __init__(self, p: float = 0.5, seed: Optional[int] = None):
         self.rng = np.random.default_rng(seed=seed)
 
     def __call__(
-        self, patch: np.ndarray, target: Optional[np.ndarray] = None
-    ) -> Tuple[np.ndarray, Optional[np.ndarray]]:
+        self,
+        patch: NDArray,
+        target: Optional[NDArray] = None,
```
jdeschamps, line 55: Would it make sense then to fold the target in the keyword argument as well?

src/careamics/transforms/n2v_manipulate.py
```diff
@@ -142,5 +143,8 @@ def __call__(
         else:
             raise ValueError(f"Unknown masking strategy ({self.strategy}).")
 
+        # TODO: Output does not match other transforms, how to resolve?
+        #     - Don't include in Compose and apply after if algorithm is N2V?
+        #     - or just don't return patch? but then mask is in the target position
```
jdeschamps, line 148: The patch needs to be returned, what I always wondered was whether we could return `patch * mask` rather than `patch, mask`.

However, there was a reason in the loss. So we might want to have a closer look again and see if we can reimplement the loss in a way where we don't need the both patch and mask individually.

jdeschamps (approved):

jdeschamps:
For the sake of the LVAE submission, let's merge it as it. But we should open issues about the following:

- One can just be discussed briefly in person: should `target` be folded in the additional arrays? Would require an additional PR.
- One issue to discuss what to do with the N2VManipulate signature, and whether n2v loss can be modified to not require `patch` and `mask`, but just `patch * mask`.

veegalinova (approved):
Great! 
Thank you very much for the LVAE code example, I'll test it soon.

PR #239, opened 2024-09-05
Add custom transforms to convenience functions, and other API improvement

### LINKED ISSUE #231: Specify transformations in config convenience functions
Currently, users can only interact with the transformations by enabling/disabling them in the convenience functions:

```python
    experiment_name="convallaria_structn2v",
    data_type="array",
    axes="YX",
    patch_size=(64, 64),
    batch_size=16,
    num_epochs=50,
    use_augmentations=False,
)
```

To have a single x-axis flip augmentation (rather than x and y-flips, and 90 degrees rotations), users would have to do the following:

```python
from careamics.config.transformations import XYFlip

config.data_config.transforms.insert(
    0,
    XYFlip(flip_x=True, flip_y=False),
)
```

This is not very convenient and would be better to be able do the following:

```python
from careamics.config.transformations import XYFlip

    experiment_name="convallaria_structn2v",
    data_type="array",
    axes="YX",
    patch_size=(64, 64),
    batch_size=16,
    num_epochs=50,
    
    # new parameters
    augmentations = [XYFlip(flip_x=True, flip_y=False)]

    # default
    use_augmentations=True,
)
```

jdeschamps:
### Description

Following https://github.com/CAREamics/careamics/issues/231, there is a lack of flexibility in the way users can specify the transforms in the config convenience functions. The choice to users comes down to using only the default augmentations or no augmentation at all.

In passing, I also added the `dataloader_params` parameter by request of students during the MBL course.

This PR does a few things:
- Refactor `TRANSFORMS_UNION` type in its own module in order to find it more easily.
- Add an `augmentations` parameter that allows user to pass non-default augmentations. 
- Remove `use_augmentations`.
- Type explicitely the Pydantic trnasfroms
- Add `dataloader_params` to allow setting PyTorch dataloader parameters (as expected by the configuration).
- Rename `model_kwargs` to `model_params` for clarity.
- Refactor the convenience functions and tests in order to minimize duplicated code.

Summary:

- **What**: Allow users to pass non-default augmentations and dataloader parameters.
- **Why**: Enable cases where augmentations are useful but need to be different from the default ones (e.g. no rotations), and cases where users might want to add some dataloader parameters.
- **How**: Added two new parameters to the convenience functions.

### Changes Made

- **Addded**: `transform_union.py` to hold `TRANSFORMS_UNION`.
- **Modified**: `configuration_factory.py`, corresponding tests and the imports of `TRANSFORMS_UNION`.

### Related Issues

- Fixes https://github.com/CAREamics/careamics/issues/231

### Breaking changes

Any call that uses the convenience functions together with the parameter `model_kwargs`.

This PR will require an update to the doc and the examples.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/configuration_factory.py
+++ b/src/careamics/config/configuration_factory.py
@@ -577,7 +577,7 @@ def create_n2n_configuration(
     ...     patch_size=[64, 64],
     ...     batch_size=32,
     ...     num_epochs=100,
-    ...     use_augmentations=[]
+    ...     augmentations=[]
     ... )
 
     A list of transforms can be passed to the `augmentations` parameter to replace the
--- a/src/careamics/config/configuration_factory.py
+++ b/src/careamics/config/configuration_factory.py
@@ -13,7 +13,6 @@ from .support import (
 )
 from .training_model import TrainingConfig
 from .transformations import (
-    TRANSFORMS_UNION,
     N2VManipulateModel,
     TransformModel,
     XYFlipModel,
@@ -75,7 +74,7 @@ def _create_configuration(
     patch_size: list[int],
     batch_size: int,
     num_epochs: int,
-    augmentations: list[TRANSFORMS_UNION],
+    augmentations: list[TransformModel],
     independent_channels: bool,
     loss: Literal["n2v", "mae", "mse"],
     n_channels_in: int,
@@ -183,7 +182,7 @@ def _create_supervised_configuration(
     batch_size: int,
     num_epochs: int,
     use_augmentations: bool = True,
-    augmentations: Optional[list[TRANSFORMS_UNION]] = None,
+    augmentations: Optional[list[TransformModel]] = None,
     independent_channels: bool = True,
     loss: Literal["mae", "mse"] = "mae",
     n_channels_in: int = 1,
@@ -251,7 +250,7 @@ def _create_supervised_configuration(
     # augmentations
     if use_augmentations:
         if augmentations is None:
-            transform_list: list[TRANSFORMS_UNION] = [
+            transform_list: list[TransformModel] = [
                 XYFlipModel(),
                 XYRandomRotate90Model(),
             ]
@@ -287,7 +286,7 @@ def create_care_configuration(
     batch_size: int,
     num_epochs: int,
     use_augmentations: bool = True,
-    augmentations: Optional[list[TRANSFORMS_UNION]] = None,
+    augmentations: Optional[list[TransformModel]] = None,
     independent_channels: bool = True,
     loss: Literal["mae", "mse"] = "mae",
     n_channels_in: int = 1,
@@ -454,7 +453,7 @@ def create_n2n_configuration(
     batch_size: int,
     num_epochs: int,
     use_augmentations: bool = True,
-    augmentations: Optional[list[TRANSFORMS_UNION]] = None,
+    augmentations: Optional[list[TransformModel]] = None,
     independent_channels: bool = True,
     loss: Literal["mae", "mse"] = "mae",
     n_channels_in: int = 1,
@@ -621,7 +620,7 @@ def create_n2v_configuration(
     batch_size: int,
     num_epochs: int,
     use_augmentations: bool = True,
-    augmentations: Optional[list[TRANSFORMS_UNION]] = None,
+    augmentations: Optional[list[TransformModel]] = None,
     independent_channels: bool = True,
     use_n2v2: bool = False,
     n_channels: int = 1,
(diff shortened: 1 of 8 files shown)


src/careamics/config/transformations/transform_union.py
```diff
@@ -0,0 +1,20 @@
...
+from .xy_random_rotate90_model import XYRandomRotate90Model
+
+TRANSFORMS_UNION = Annotated[
+    Union[
+        XYFlipModel,
+        XYRandomRotate90Model,
+        N2VManipulateModel,
+    ],
+    Discriminator("name"),  # used to tell the different transform models apart
+]
```
melisande-c, line 19: Small question: if all the transform models inherit from `TransformModel`, why can't they simply be type hinted as this base class?

melisande-c: Although I notice that `Normalize` is not included in the type union

src/careamics/config/configuration_factory.py
```diff
@@ -1,46 +1,202 @@
...
+    TRANSFORMS_UNION,
+    N2VManipulateModel,
+    TransformModel,
+    XYFlipModel,
+    XYRandomRotate90Model,
+)
+
+
+def _create_unet(
+    axes: str,
```
melisande-c: maybe `_create_unet_configuration`?

src/careamics/config/configuration_factory.py
```diff
@@ -57,6 +213,9 @@ def _create_supervised_configuration(
         Number of epochs.
     use_augmentations : bool, optional
         Whether to use augmentations, by default True.
+    augmentations : list of TransformModel, optional
+        List of transforms (augmentations) to apply instead of the default ones, by
+        default None.
```
melisande-c: Maybe mention defaults are xyfilp and rotate90

src/careamics/config/configuration_factory.py
```diff
@@ -1,46 +1,202 @@
...
+        experiment_name=experiment_name,
+        algorithm_config=algorithm_config,
+        data_config=data,
+        training_config=training,
+    )
+
+    return configuration
 
 
-# TODO rename ?
```
melisande-c, line 19: I guess TODO was here because something like muSplit etc. are also supervised

src/careamics/config/configuration_factory.py
```diff
@@ -319,23 +602,26 @@ def create_n2n_configuration(
...
 def create_n2v_configuration(
     experiment_name: str,
     data_type: Literal["array", "tiff", "custom"],
     axes: str,
-    patch_size: List[int],
+    patch_size: list[int],
     batch_size: int,
     num_epochs: int,
     use_augmentations: bool = True,
+    augmentations: Optional[list[TRANSFORMS_UNION]] = None,
```
melisande-c: `use_augmentations` argument is maybe redundant? If a user doesn't want augmentations they can pass an empty list to augmentations?

melisande-c (approved):
Added some small, not important comments

src/careamics/config/transformations/transform_union.py
jdeschamps: This annotation is actually used by Pydantic to select which transform to instantiate (based on the `Discriminator`). 

Now, whether this can be replaced for typing elsewhere in the code, I need to have another look!

src/careamics/config/configuration_factory.py
jdeschamps: Could be, just a very undescriptive todo... (which I may have written ^^).

src/careamics/config/transformations/transform_union.py
melisande-c: I see, ok! I need to look at the Pydantic discriminator stuff again, I always forget how it works

src/careamics/config/configuration_factory.py
jdeschamps: True, I just felt that `use_augmentations=False` is a much more readable way of doing it for users (especially the less technically-minded).

Now, it does come at the cost of a few lines of code and a potential problem:

```python
use_augmentations = True,
augmentations = [ ],
```

but I feel that's worth it. 

Happy to change my mind, I don't have a super strong opinion.

melisande-c: fair enough, I don't have a strong opinion either

src/careamics/config/transformations/transform_union.py
jdeschamps: So indeed for all the uses outside the Pydantic model, we can use the parent class.

But this brought me to then specify exactly what transforms were allowed (since there are only two...), then to write a lot of guarding clauses that became too complex.

So indeed went along with your suggestion to only use a single parameter to simplify things!

src/careamics/config/configuration_factory.py
jdeschamps: After writing a lot of guarding conditions and raising errors right and left, I went along with your suggestion!

jdeschamps:
I typed the transforms differently in the end, and added further refactoring and some error raising!

veegalinova (approved):

PR #233, opened 2024-08-31
refac: removed noise model parameters from `LadderVAE` module

federico-carrara:
### Description

- **What**: Remove superfluous noise model-related parameters (e.g., `enable_noise_model`) from `LadderVAE` `torch.nn.Module` class.
- **Why**: These parameters are not used anymore since noise model has been moved out of the main model class.
- **How**: Deleted parameters, tested resulting code.

---

**Please ensure your PR meets the following requirements:**

- [x] Code builds and passes tests locally, including doctests
- [ ] New tests have been added (for bug fixes/features)
- [x] Pre-commit passes
- [ ] PR to the documentation exists (for bug fixes / features)

PR #240, opened 2024-09-05
3D model

CatEek:
### Description

Adding 3D/2.5D to LVAE model

- **What**: Merge current implementation in the original repo with refactoring done previously.

### Changes Made
- Relevant pydantic config
- Added encoder_conv_strides and decoder_conv_strides params. They control the strides in conv layers, and can have len 2 or 3. They're meant to make a choice between 2D and 3D convs and control the shape of encoder/decoder  (subject to change)
- Input shape is now a tuple containing shapes of 2/3 dimensions
- Stochastic layer are in the separate module (subject to change)
- NonStochastic is removed alongside with relevant parameters
- Some docs update
- Basic tests 

### Notes/Problems
- Dosctings are a mess, should be fixed in a separate PR later
- Lot's of mypy etc issues
- Some tests don't pass because we need to clarify multiscale count param 

---

- [x ] Code builds and passes tests locally, including doctests
- [x ] New tests have been added (for bug fixes/features)
- [ ] Pre-commit passes


src/careamics/config/architectures/lvae_model.py
```diff
@@ -15,9 +15,14 @@ class LVAEModel(ArchitectureModel):
...
-    multiscale_count: int = Field(default=5)  # TODO clarify
-    # 0 - off, len(z_dims) + 1 # TODO can/should be le to z_dims len + 1
+    input_shape: list[int] = Field(default=(64, 64), validate_default=True)
+    """Shape of the input patch (C, Z, Y, X) or (C, Y, X) if the data is 2D."""
+    encoder_conv_strides: list = Field(default=[2, 2], validate_default=True)
+    # TODO make this per hierarchy step ?
+    decoder_conv_strides: list = Field(default=[2, 2], validate_default=True)
+    """Dimensions (2D or 3D) of the convolutional layers."""
+    multiscale_count: int = Field(default=1)  # TODO clarify
+    # 1 - off, len(z_dims) + 1 # TODO can/should be le to z_dims len + 1 or ?
```
federico-carrara: FYI, I plan to refactor the way this parameter works pretty soon.

src/careamics/config/architectures/lvae_model.py
```diff
@@ -40,6 +45,88 @@ class LVAEModel(ArchitectureModel):
...
+        # adding 1 to encoder strides for the number of input channels
+        if len(self.input_shape) != len(self.encoder_conv_strides):
+            raise ValueError(
+                f"Number of input dimensions must be equal to the number of encoder conv strides"
+                f" (got {len(self.input_shape)} and {len(self.encoder_conv_strides)})."
+            )
+
+        if len(self.encoder_conv_strides) < len(self.decoder_conv_strides):
+            raise ValueError(
+                f"Decoder can be 3D when encoder is 2D (got"
```
federico-carrara: I think you wanted to say "Decoder CAN'T be ..."

src/careamics/config/architectures/lvae_model.py
```diff
@@ -40,6 +45,88 @@ class LVAEModel(ArchitectureModel):
...
+                f" {len(self.encoder_conv_strides)} and {len(self.decoder_conv_strides)})."
+            )
+
+        if any(s < 1 for s in self.encoder_conv_strides) or any(
+            s < 1 for s in self.decoder_conv_strides
```
federico-carrara, line 84: Most precisely, I guess we should check
`any(s < 1 for s in self.encoder_conv_strides[1:])`
since in this implementation stride along Z is always 1, and we want the strides in the other spatial dimensions to be grater than 1.

src/careamics/config/architectures/lvae_model.py
```diff
@@ -40,6 +45,88 @@ class LVAEModel(ArchitectureModel):
...
+        ):
+            raise ValueError(
+                f"Strides must be greater than 1 (got {self.encoder_conv_strides} and {self.decoder_conv_strides})."
```
federico-carrara: "At least one stride value" (?)

src/careamics/config/architectures/lvae_model.py
```diff
@@ -40,6 +45,88 @@ class LVAEModel(ArchitectureModel):
...
+        ValueError
+            If the number of dimensions is not 3 or 4.
+        """
+        if len(input_shape) < 2 or len(input_shape) > 3:
+            raise ValueError(
+                f"Number of input dimensions must be 2 for 2D data 3 for 3D (got {len(input_shape)})."
+            )
+
+        if any(s < 1 for s in input_shape):
+            raise ValueError(f"Input shape must be greater than 1 (got {input_shape}).")
```
federico-carrara: "greater than 1 IN AT LEAST ONE DIMENSION"

src/careamics/config/architectures/lvae_model.py
```diff
@@ -142,12 +229,16 @@ def validate_multiscale_count(cls, self: Self) -> Self:
...
+        if self.multiscale_count > 1:
+            if self.multiscale_count != len(self.z_dims) - 1:
+                raise ValueError(
+                    f"Multiscale count must be 0 or equal to the number of Z "
+                    f"dims - 1 (got {self.multiscale_count} and {len(self.z_dims)})."
+                )
+        elif self.multiscale_count == 1:
+            pass # TODO add multicsale/z dims validation
+        else:
+            raise ValueError(f"Multiscale count must be greater or equal to 1 (got {self.multiscale_count}).")
```
federico-carrara: This is wrong, `multiscale_count` should be 1 when LC is disabled or any number between 2 and `len(z_dims) + 1` when LC is enabled.

src/careamics/config/architectures/lvae_model.py
```diff
@@ -160,7 +251,10 @@ def set_3D(self, is_3D: bool) -> None:
         is_3D : bool
             Whether the algorithm is 3D or not.
         """
-        raise NotImplementedError("VAE is not implemented yet.")
+        if is_3D:
+            self.conv_dims = 3
+        else:
+            self.conv_dims = 2
```
federico-carrara, line 244: I think this is not defined. But at the same time, I'd like this to be defined, so that we can use it instead of less intuitive (and clean) `len(encoder_conv_strides)` (we already discussed this).

src/careamics/models/lvae/non_stochastic.py
federico-carrara, line 1: Please, delete this and, instead, restore the module within `layers.py` to the version we currently have in `main`.

src/careamics/models/lvae/stochastic.py
federico-carrara, line 1: Same as above.

tests/conftest.py
```diff
@@ -395,3 +395,27 @@ def create_dummy_noise_model(
...
+        "multiscale_count": None,
+        "conv_strides": [2, 2],
+        "z_dims": [128, 128, 128, 128],
+        "encoder_n_filters": 64,
+        "decoder_n_filters": 64,
+        "encoder_dropout": 0.1,
+        "decoder_dropout": 0.1,
+        "nonlinearity": "ELU",
+        "predict_logvar": "pixelwise",
+        "enable_noise_model": False,
```
federico-carrara: `enable_noise_model` was removed in one of the latest PRs. so it should not be here (test might fail too!)

tests/models/lvae/test_lvae_architecture.py
```diff
@@ -0,0 +1,526 @@
...
+    gmm = GaussianMixtureNMConfig(
+        model_type="GaussianMixtureNoiseModel",
+        path=tmp_path / "dummy_noise_model.npz",
+        # all other params are default
+    )
+    # TODO is this correct?
+    multi_channel_nm = MultiChannelNMConfig(noise_models=[gmm] * output_channels)
+    # algorithm, loss and noise_model are not important here since
+    # we are only interested in testing the model architecture
+    config = VAEAlgorithmConfig(
```
federico-carrara, line 50: I guess we actually don't need the noise model here, since we are only testing the model itself. 
You might simply leave it to the default value  in `VAEAlgorithmConfig` (which should be `None`).

tests/models/lvae/test_lvae_architecture.py
federico-carrara, line 1: General comment: it is really a lot of code, so I will need more time to go over it (I see some parts where we could have issues). For the moment I agree to skip all tests

src/careamics/models/lvae/lvae.py
```diff
@@ -113,14 +154,6 @@ def __init__(
```
federico-carrara, line 139: Please define `self.encoder_conv_dims = len(encoder_conv_strides)` for readability ;) (same for decoder obv)

src/careamics/models/lvae/lvae.py
```diff
@@ -445,54 +509,63 @@ def _init_multires(
...
-        NOTE 2: `_multiscale_count` attribute defines the total number of inputs to the bottom-up pass.
-        In other terms if we have the input patch and n_LC additional lateral inputs, we will have a total of (n_LC + 1) inputs.
+        Method defines the input block/branch to encode/compress low-res lateral inputs.
+
+        at different hierarchical levels
+        in the multiresolution approach (LC). The role of the input branches is similar
+        to the one of the first bottom-up layer in the primary flow of the Encoder,
+        namely to compress the lateral input image to a degree that is compatible with
+        the one of the primary flow.
+
```
federico-carrara, line 500: Linting fixes destroyed the content of the comment... We should fix this

src/careamics/models/lvae/lvae.py
```diff
@@ -445,54 +509,63 @@ def _init_multires(
...
-        ), msg
+        ), msg  # TODO how ?
 
         msg = (
-            "if multiscale is enabled, then we are just working with monocrome images."
+            "Multiscale approach only supports monocrome images. "
+            f"Found instead color_ch={self.color_ch}."
         )
-        assert self._multiscale_count == 1 or self.color_ch == 1, msg
+        # assert self._multiscale_count == 1 or self.color_ch == 1, msg
```
federico-carrara, line 528: That is still true, so I'd not comment it out.

src/careamics/models/lvae/layers.py
```diff
@@ -51,13 +50,13 @@ def __init__(
         self,
         channels: int,
         nonlin: Callable,
-        kernel: Union[int, Iterable[int]] = None,
+        conv_strides: tuple[int] = (2, 2),
```
federico-carrara, line 50: replace with `conv_dims` once defined :)

src/careamics/models/lvae/layers.py
```diff
@@ -178,17 +185,24 @@ def __init__(self, *args, **kwargs):
...
     """
     Double the number of channels through a convolutional layer, then use
     half the channels as gate for the other half.
     """
 
-    def __init__(self, channels, kernel_size, nonlin=nn.LeakyReLU):
+    def __init__(
+        self,
+        channels: int,
+        conv_strides: tuple[int] = (2, 2),
```
federico-carrara, line 208: conv_dims

src/careamics/models/lvae/layers.py
```diff
@@ -1333,666 +1332,4 @@ def forward(
...
-
-        # Compute log p(z)# NOTE: disabling its computation.
-        # if mode_pred is False:
-        #     logprob_p =  p.log_prob(z).sum((1, 2, 3))
-        # else:
-        #     logprob_p = None
-
-        if q_params is not None:
-            # Compute log q(z)
-            logprob_q = q.log_prob(z).sum((1, 2, 3))
```
federico-carrara, line 1755: Also this needs to be changed for 3D.

src/careamics/models/lvae/layers.py
```diff
@@ -1333,666 +1332,4 @@ def forward(
...
-
-            # KL term only associated to the portion of the latent tensor that is used for prediction and
-            # summed over channel and spatial dimensions. [Shape: (batch, )]
-            # NOTE: vanilla_latent_hw is the shape of the latent tensor used for prediction, hence
-            # the restriction has shape [Shape: (batch, ch, vanilla_latent_hw[0], vanilla_latent_hw[1])]
-            if self._restricted_kl:
-                pad = (kl_elementwise.shape[-1] - self._vanilla_latent_hw) // 2
-                assert pad > 0, "Disable restricted kl since there is no restriction."
-                tmp = kl_elementwise[..., pad:-pad, pad:-pad]
-                kl_samplewise_restricted = tmp.sum((1, 2, 3))
```
federico-carrara, line 1564: Talking about what could not work in the 3D case if we restore this (as we should): all the aggregations of KL clearly need to be changed!

federico-carrara: Please change also the comments accordingly, so that they reflect the actual dimensions of tensors (otherwise it is a mess to understand).

src/careamics/models/lvae/layers.py
```diff
@@ -1333,666 +1332,4 @@ def forward(
...
-        """
-        q_params = self.conv_in_q(q_params)
-
-        q_mu, q_lv = q_params.chunk(2, dim=1)
-        if var_clip_max is not None:
-            q_lv = torch.clip(q_lv, max=var_clip_max)
-
-        if q_mu.shape[-1] % 2 == 1 and allow_oddsizes is False:
-            q_mu = F.center_crop(q_mu, q_mu.shape[-1] - 1)
-            q_lv = F.center_crop(q_lv, q_lv.shape[-1] - 1)
```
federico-carrara, line 1650: This will fail for 3D since `F.center_crop` only works in 3D. A solution is to replace it with `crop_to_img` defined somewhere in the codebase (likely in `lvae/utils`).

src/careamics/models/lvae/layers.py
```diff
@@ -1333,666 +1332,4 @@ def forward(
...
-            logprob_q = None
-
-        # Store meaningful quantities to use them in following layers
-        data = kl_dict
-        data["z"] = z  # sampled variable at this layer (batch, ch, h, w)
-        data["p_params"] = p_params  # (b, ch, h, w) where b is 1 or batch size
-        data["q_params"] = q_params  # (batch, ch, h, w)
-        # data['logprob_p'] = logprob_p  # (batch, )
-        data["logprob_q"] = logprob_q  # (batch, )
-        data["qvar_max"] = debug_qvar_max
```
federico-carrara, line 1771: Please update shapes in the comments for 3D!

src/careamics/models/lvae/layers.py
```diff
@@ -1333,666 +1332,4 @@ def forward(
...
-
-        return out, data
-
-
-class NonStochasticBlock2d(nn.Module):
```
federico-carrara, line 1776: This is more or less the same to the `NormalStochasticBlock2d`. Hence it will need the same changes!

federico-carrara: Clearly also the name needs to be changed!!!

federico-carrara (changes requested):
Left some comments, although I couldn't review everything in extreme detail.
The "content" is more or less there, but there are a lot of style-related changes to be done if we want to keep the code clean.
Happy to help if something is not clearl :)

jdeschamps:
Can you update the description? Impossible to review otherwise. :)

src/careamics/config/architectures/lvae_model.py
CatEek: Which parameter ?

src/careamics/config/architectures/lvae_model.py
CatEek: I might be not 1 in the future

src/careamics/models/lvae/layers.py
CatEek: Good point!

src/careamics/config/architectures/lvae_model.py
CatEek: What's wrong then ?

src/careamics/config/architectures/lvae_model.py
CatEek: We have conv strides because we can also have 2.5D which is 3D encoder 2D decoder. And we'd have to define strides anyway, otherwise we'd have to hardcode them which is not good

src/careamics/models/lvae/non_stochastic.py
CatEek: Why?

tests/models/lvae/test_lvae_architecture.py
CatEek: We need couple more test cases that we'll make with @ashesh-0 and after performance matches we can continue refactoring

src/careamics/models/lvae/lvae.py
CatEek: Not sure what you mean and why do we need that param ?

src/careamics/models/lvae/layers.py
CatEek: We need to discuss this in more detail to figure out the way we handle 2D/3D

src/careamics/models/lvae/layers.py
CatEek: I guess that is fixed 
`q.log_prob(z).sum(tuple(range(1, z.dim())))`

src/careamics/models/lvae/layers.py
CatEek: Also updated in the newer version

CatEek:
@federico-carrara Thanks for the review, updates some things, some were already fixed after you reviewed. I have a couple of questions though. Could you please take a look again

src/careamics/models/lvae/layers.py
CatEek: So NormalStochasticBlock2d isn't used in current algorithms

src/careamics/config/architectures/lvae_model.py
federico-carrara: `multiscale_count`

src/careamics/config/architectures/lvae_model.py
federico-carrara: Ye, I know. But I would anyway create also a `conv_dims` attributes within the model class derived from the strides

src/careamics/models/lvae/non_stochastic.py
federico-carrara: Because this is not refactored (i.e., no comments, code's not polished) and moreover I think it is more consistent to have all layers modules in the same place (otherwise we'd need to have a different file for each layer module). This is how it used to be in Ashesh repo. 
Finally, the module we have in the main should work just fine, and only slight modifications would be needed to extend to 3D.

src/careamics/models/lvae/lvae.py
federico-carrara: Not needed at all, but would greatly help readability and conciseness.

src/careamics/models/lvae/layers.py
CatEek: But it's not a stride it's a kernel size if I'm not mistaken

src/careamics/config/architectures/lvae_model.py
```diff
@@ -124,27 +212,20 @@ def validate_z_dims(cls, z_dims: tuple) -> tuple:
...
-        #     if self.multiscale_count != len(self.z_dims) - 1:
-        #         raise ValueError(
-        #             f"Multiscale count must be 0 or equal to the number of Z "
-        #             f"dims - 1 (got {self.multiscale_count} and {len(self.z_dims)})."
-        #         )
-
+        if self.multiscale_count < 1 or self.multiscale_count > len(self.z_dims) + 1:
+            raise ValueError(
+                f"Multiscale count must be 1 for LC off or equal to the number of Z"
+                f"dims + 1 (got {self.multiscale_count} and {len(self.z_dims)})."
```
federico-carrara: This error message is not entirely true @CatEek. 
Indeed, it can attain **any** value between 1 and z_dims + 1. 
For instance multi_count=3 means LC on the first 2 hierarchies.

tests/models/lvae/test_lvae_architecture.py
```diff
@@ -0,0 +1,547 @@
...
+    lvae_model_config = LVAEModel(
+        architecture="LVAE",
+        input_shape=input_shape,
+        z_dims=z_dims,
+        encoder_conv_strides=encoder_conv_strides,
+        decoder_conv_strides=decoder_conv_strides,
+        multiscale_count=multiscale_count,
+        output_channels=output_channels,
+        predict_logvar=predict_logvar,
+        enable_noise_model=False,
```
federico-carrara: Since #233 this is not anymore among the model parameters. So I assume this would raise an error (?)

tests/models/lvae/test_lvae_architecture.py
```diff
@@ -0,0 +1,547 @@
...
+
+@pytest.mark.skip(reason="Needs to be updated")
+@pytest.mark.parametrize(
+    "img_size, encoder_conv_strides, decoder_conv_strides",
+    [
+        ([1, 64, 64], [2, 2], [2, 2]),
+        ([1, 128, 128], [2, 2], [2, 2]),
+        ([1, 16, 64, 64], [1, 2, 2], [1, 2, 2]),
+        ([1, 16, 128, 128], [1, 2, 2], [1, 2, 2]),
+        ([1, 16, 64, 64], [1, 2, 2], [2, 2]),
```
federico-carrara, line 68: Not entirely sure about the new meaning of `img_size`. What does the first axis stand for? Is it the LC dimension?

src/careamics/config/architectures/lvae_model.py
CatEek: Yeah true, misleading

tests/models/lvae/test_lvae_architecture.py
CatEek: Probably not cuz we have extras allowed in pydantic model. But will remove

tests/models/lvae/test_lvae_architecture.py
CatEek: It's a skipped test. Needs update

src/careamics/lightning/lightning_module.py
```diff
@@ -265,12 +266,16 @@ def __init__(self, algorithm_config: Union[VAEAlgorithmConfig, dict]) -> None:
...
         # self.save_hyperparameters(self.algorithm_config.model_dump())
-
+        # define likelihood configurations
+        self.algorithm_config.noise_model_likelihood_model = NMLikelihoodConfig()
         # create model and loss function
         self.model: nn.Module = model_factory(self.algorithm_config.model)
         self.noise_model: NoiseModel = noise_model_factory(
             self.algorithm_config.noise_model
         )
+        self.algorithm_config.noise_model_likelihood_model.noise_model = (
```
federico-carrara, line 276: Can you explain why we need this here? In my understanding these attributes can be directly defined in the `VAEAlgorithmConfig`

federico-carrara: Same holds for the noise model ",model" above

src/careamics/losses/lvae/losses.py
```diff
@@ -168,18 +170,27 @@ def get_kl_divergence_loss_usplit(
         The key for the KL-loss values in the top-down layer data dictionary.
         To choose among ["kl", "kl_restricted", "kl_spatial", "kl_channelwise"]
         Default is "kl".
+
+    z_size: int
+        The size of the Z dimension in the 3D model. 0 if 2D model.
```
federico-carrara, line 175: Outdated param description (?)

src/careamics/losses/lvae/losses.py
```diff
@@ -210,12 +221,21 @@ def get_kl_divergence_loss_denoisplit(
         Default is "kl"
 
     kl[i] for each i has length batch_size resulting kl shape: (bs, layers).
+
+    z_size: int
+        The size of the Z dimension in the 3D model. 0 if 2D model.
```
federico-carrara, line 226: Oudated param description (?)

src/careamics/lvae_training/eval_utils.py
```diff
@@ -596,51 +606,49 @@ def get_dset_predictions(
...
-        np.concatenate(predictions_std, axis=0),
-        np.concatenate(logvar_arr),
-        np.array(losses),
-        psnr,
-    )
+            # predictions_std.append(mmse_std.cpu().numpy())
+
+    # psnr = [x.get() for x in patch_psnr_channels]
+    return np.concatenate(predictions, axis=0)
+    # np.concatenate(predictions_std, axis=0),
```
federico-carrara, line 648: If I remember correctly, this output is needed for the sake of calibration... So we might need to fix the code (as I guess it is commented out because there's some bug)

federico-carrara: But it is neither urgent nor necessary rn, so leaving a TODO is enough. And I also think all of this will be replaced by something @melisande-c has implemented.

src/careamics/models/lvae/layers.py
```diff
@@ -727,33 +814,37 @@ def __init__(self, *args, **kwargs):
...
-            w_pad = (lw - w) // 2
-            lowres = lowres[:, :, h_pad:-h_pad, w_pad:-w_pad]
+            lz, ly, lx = lowres.shape[2:]
+            z = lz // self.multiscale_lowres_size_factor
+            y = ly // self.multiscale_lowres_size_factor
+            x = lx // self.multiscale_lowres_size_factor
+            z_pad = (lz - z) // 2
+            y_pad = (ly - y) // 2
+            x_pad = (lx - x) // 2
+            lowres = lowres[:, :, z_pad:-z_pad, y_pad:-y_pad, x_pad:-x_pad]
```
federico-carrara, line 841: This might be wrong, since we don't want to touch the z dimension. 
At the same time this `else` will never be executed in standard cases (and, anyway, I solved this in another PR I am working on)

src/careamics/models/lvae/layers.py
```diff
@@ -990,75 +1063,68 @@ def __init__(
...
+        self.stochastic = NormalStochasticBlock(
+            c_in=n_filters,
+            c_vars=z_dim,
+            c_out=n_filters,
+            conv_dims=len(conv_strides),
+            transform_p_params=(not is_top_layer),
+            vanilla_latent_hw=vanilla_latent_hw,
+            restricted_kl=restricted_kl,
+            use_naive_exponential=stochastic_use_naive_exponential,
+        )
```
federico-carrara, line 1120: Finally removed the non-stochastic block. Nice :)

src/careamics/models/lvae/stochastic.py
```diff
@@ -0,0 +1,398 @@
...
+        else:  # if predicting, no need to compute KL
+            kl_elementwise = kl_samplewise = kl_spatial = kl_channelwise = None
+
+        kl_dict = {
+            "kl_elementwise": kl_elementwise,  # (batch, ch, h, w)
+            "kl_samplewise": kl_samplewise,  # (batch, )
+            "kl_samplewise_restricted": kl_samplewise_restricted,  # (batch, )
+            "kl_spatial": kl_spatial,  # (batch, h, w)
+            "kl_channelwise": kl_channelwise,  # (batch, ch)
+        }
```
federico-carrara: Need to update dimension in comments. Clearly, not fundamental, but very useful !

src/careamics/models/lvae/stochastic.py
```diff
@@ -0,0 +1,398 @@
...
+        """
+        q_params = self.conv_in_q(q_params)
+
+        q_mu, q_lv = q_params.chunk(2, dim=1)
+        if var_clip_max is not None:
+            q_lv = torch.clip(q_lv, max=var_clip_max)
+
+        if q_mu.shape[-1] % 2 == 1 and allow_oddsizes is False:
+            q_mu = F.center_crop(q_mu, q_mu.shape[-1] - 1)
+            q_lv = F.center_crop(q_lv, q_lv.shape[-1] - 1)
```
federico-carrara, line 297: Remove?! It anyway won't work a `F.center_crop()` only works on 2D data

src/careamics/models/lvae/utils.py
```diff
@@ -162,29 +190,12 @@ def crop_img_tensor(x, size) -> torch.Tensor:
...
-        When log() is called, torch.log(self._tensor.exp()) is computed instead.
-
-    NOTE 2:
-        Given the output from exp(), torch.log() or the log() method of the class give identical results.
+    Here, the idea is that everything is done on the tensor which you've given in the constructor.
+    when exp() is called, what that means is that we want to compute self._tensor.exp()
+    when log() is called, we want to compute torch.log(self._tensor.exp())
+
+    What is done here is that definition of exp() has been changed. This, naturally, has changed the result of log.
+    but the log is still the mathematical log, that is, it takes the math.log() on whatever comes out of exp().
```
federico-carrara, line 198: I think some merging error happened here...

src/careamics/models/lvae/utils.py
```diff
@@ -206,60 +217,39 @@ def exp(self):
...
-        Parameters
-        ----------
-        logvar: torch.Tensor
-            The input (true) logvar vector, to be converted in the Stable version.
-        enable_stable: bool, optional
-            Whether to compute the stable version of log-variance. Default is `True`.
-        var_eps: float, optional
-            The minimum value attainable by the variance. Default is `1e-6`.
+        Args:
+            var_eps: var() has this minimum value.
```
federico-carrara: I think some merging error happened here...

src/careamics/models/lvae/lvae.py
```diff
@@ -17,42 +18,80 @@
...
-        Parameters
-        ----------
-
-        """
         super().__init__()
 
         # -------------------------------------------------------
         # Customizable attributes
         self.image_size = input_shape
+        """Input image size. (C, Z, Y, X) or (C, Y, X) if the data is 2D."""
```
federico-carrara: Need clarification here: Channel dimension is now part of `image_size`? If yes, what do we expect it to be? In my understanding, it should be `multiscale_count`. 

But if this is the case then we would have redundant info in this attribute!

federico-carrara (commented):
Overall it seems accurate and there are no evident issues. However, I pointed out some little stuff to be addressed. Try to have a look to see if you can solve them. If it seems too much work for you tell me in the comment or leave a TODO, but please don't ignore. Otherwise my reviewing work is just useless and wasted ;)

src/careamics/models/lvae/lvae.py
CatEek: Multiscale_count is different. But channel shouldn't be there

src/careamics/lvae_training/eval_utils.py
CatEek: Yeah, commented because of some inconsistencies. We'd have to revisit

federico-carrara (approved):

PR #249, opened 2024-09-26
Add stop training method

jdeschamps:
### Description

If CAREamics is used in a multithreaded application (e.g. napari plugin), then we need a way to stop the training programmatically. This PR add a `stop_training` method to the `CAREamist`.

Additionally, this PR removes the unused parameter `experiment_name` from the CAREamist.

- **What**: Add stop training method.
- **Why**: Enable stopping training in multithreaded applications (e.g. napari)
- **How**: Add `stop_training` method that raises the `should_stop` flag of Lightning.

### Changes Made

- **Modified**: `careamist.py`, by adding the `stop_training` function and cleaning up unused keywords.

---

**Please ensure your PR meets the following requirements:**

- [x] Code builds and passes tests locally, including doctests
- [x] New tests have been added (for bug fixes/features)
- [x] Pre-commit passes
- [ ] PR to the documentation exists (for bug fixes / features)

PR #253, opened 2024-10-14
Feature: Predict to disk (outerloop implementation)

melisande-c:
### Description

- **What**: Add a `predict_to_disk` function to the `CAREamist` class.
- **Why**: So users can save predictions without having to write the saving process themselves.
- **How**: This implementation loops through the files, predicts the result and saves each one in turn. N.b. this will eventually be replaced with the `PredictionWriterCallback` version.

### Changes Made

- **Added**: 
  - `predict_to_disk` function to `CAREamist` class.
  - tests
- **Modified**:
  - Some type hints.

Currently the results are saved to a directory called "predictions" but maybe it should be configurable by the user?

This function can only be called on source data from a path because the prediction files are saved with the same name as the source files.

Not the neatest as I expect this to be replaced soon.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -682,9 +682,7 @@ class CAREamist:
         tile_size: Optional[tuple[int, ...]] = None,
         tile_overlap: Optional[tuple[int, ...]] = (48, 48),
         axes: Optional[str] = None,
-        data_type: Optional[
-            Union[Literal["array", "tiff", "custom"], SupportedData]
-        ] = None,
+        data_type: Optional[Literal["array", "tiff", "custom"]] = None,
         tta_transforms: bool = True,
         dataloader_params: Optional[dict] = None,
         read_source_func: Optional[Callable] = None,
@@ -772,9 +770,9 @@ class CAREamist:
         if write_func_kwargs is None:
             write_func_kwargs = {}
 
-        data_type = SupportedData(data_type)
-        # TODO: make configurable?
+        # write_type = SupportedData(write_type)
 
+        # TODO: make configurable?
         write_dir = self.work_dir / "predictions"
 
         # guards for custom types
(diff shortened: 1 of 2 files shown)


src/careamics/careamist.py
```diff
@@ -668,6 +669,168 @@ def predict(
...
+        write_type: Literal["tiff", "custom"] = "tiff",
+        write_extension: Optional[str] = None,
+        write_func: Optional[WriteFunc] = None,
+        write_func_kwargs: Optional[dict[str, Any]] = None,
+        **kwargs,
+    ) -> None:
+        """
+        Make predictions on the provided data and save outputs to files.
+
+        The predictions will be saved in a new directory 'predictions' within the set
```
jdeschamps, line 695: From the user point of view, it might make more sense to have a mandatory `destination` folder.

I am not sure how that would fit API-wise with your future plans. If `destination` is a zarr, then it detect that in the future (together with the `write_type`?

src/careamics/careamist.py
```diff
@@ -668,6 +669,168 @@ def predict(
...
+        index appended. E.g. If the the source file name is 'images.tiff' then the first
+        sample's prediction will be saved with the file name "image_0.tiff".
+        Input can be a PredictDataModule instance, a path to a data file, or a numpy
+        array.
+
+        If `data_type`, `axes` and `tile_size` are not provided, the training
+        configuration parameters will be used, with the `patch_size` instead of
+        `tile_size`.
+
+        Test-time augmentation (TTA) can be switched off using the `tta_transforms`
```
jdeschamps: "switched on" since the default value is `False`?

jdeschamps (approved):

src/careamics/careamist.py
melisande-c: Oops, I copy and pasted the docs from the `predict` function. In your last PR #249 you changed the tta default to False. I can update the docstring for both `predict` and `predict_to_disk`

src/careamics/careamist.py
melisande-c: I think it can work if `destination` is a directory. I think it is better to not give too many options, makes the code less confusing. If enough people request different behaviour we can think about it

jdeschamps: Sounds like a good solution, `destination` as a folder.

src/careamics/careamist.py
jdeschamps: yes please!

When commenting, I totally had the impression that I was responsible for that. ^^

src/careamics/careamist.py
melisande-c: There is now `prediction_dir` parameter, default is "predictions". It can handle the directory being absolute or relative to the working directory set in CAREamist.

PR #261, opened 2024-11-04
refac: made some attributes of `VAEAlgorithmConfig` optional

federico-carrara:
### Description

- **What**: Made all attributes, except `model`, `Optional` in `VAEAlgortihmConfig`.
- **Why**: At inference time the only the `model` attribute is needed. This allows to simplify model initialization for evaluation.
- **How**: Simply marked attributes as `Optional`, added `None` default if needed.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/vae_algorithm_model.py
+++ b/src/careamics/config/vae_algorithm_model.py
@@ -40,18 +40,18 @@ class VAEAlgorithmConfig(BaseModel):
     #   - values can still be passed as strings and they will be cast to Enum
     algorithm: Literal["musplit", "denoisplit"]
 
-    # NOTE: these are all configs (pydantic models)
-    loss: LVAELossConfig
+    # NOTE: these are all configs (pydantic models)s
     model: Union[LVAEModel, CustomModel] = Field(discriminator="architecture")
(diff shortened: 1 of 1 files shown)


jdeschamps (commented):
That means that we would need to check whether they have been assigned for training.

Would it make sense to simply give them a default value?

federico-carrara:
I think I got your point, but let me be sure. The problem you are pointing out is that in this way we could instantiate an Algorithm config that will raise a RuntimrError during training since some attributes are not defined, right?

If this is the case, yeah good point. I guess a solution is to change the default values of the pydantic configs, e.g.,
```
loss: Optional[LVAELossConfig] = LVAELossConfig()
```

What do you think?

federico-carrara:
Or we could also skip this PR. However, it is really ugly to load a loss config or an optimizer config which are just useless at inference time...

jdeschamps:
Sorry I lost track of this. 

I do prefer having an unused parameter class than guarding against it being `None` in the training calls.

federico-carrara:
I completely see your point. Let's skip this PR then, not important at all!
If you agree I will close it :)

PR #272, opened 2024-11-13
refac: added possibility to pick `kl_restricted` loss & other loss-related refactoring

federico-carrara:
### Description

In some LVAE training examples (see `microSplit_reproducibility` repo) there is the need to consider the *restricted KL loss*, instead of the simple sample-wise one. This PR allows the user to pick that one.

- **What**: The KL loss type is no longer hardcoded in the loss functions. Now it is possible to pick also the `restricted_kl` KL loss type.
- **Why**: It is needed for some experiments/examples.
- **How**:  added an input parameter to the KL loss functions.

### Breaking changes

The `kl_type` parameter is added in the loss functions, so we need to be careful of correctly specifying it in the examples in the `microSplit_reproducibility` repo. 

---

- [x] Pre-commit passes


src/careamics/losses/loss_factory.py
```diff
@@ -35,92 +28,6 @@ class FCNLossParameters:
     loss_weight: float
 
 
-@dataclass  # TODO why not pydantic?
-class LVAELossParameters:
```
jdeschamps, line 39: Can you comment on where do all these parameters went and how they are now passed to the loss?

jdeschamps (approved):

src/careamics/losses/loss_factory.py
federico-carrara: It was deprecated, replaced by what we now have in `src/careamics/config/loss_model.py`.

PR #277, opened 2024-11-18
fix: fixed a bug in KL loss aggregation (LVAE)

federico-carrara:
### Description

Found a bug in the KL loss aggregation happening in the `LadderVAE` model `training_step()`.
Specifically, the application of free-bits (`free_bits_kl()`, basically clamping the values of KL entries to a certain lower threshold) was happening after KL entries were rescaled. In this way, when free-bits threshold was set to 1, all the KL entries were clamped to 1, as normally way smaller than this.

- **What**: See above.
- **Why**: Clear bug in the code.
- **How**: Inverted the order of calls in the `get_kl_divergence_loss()` function & adjusted some parts of the code to reflect the changes.
---

- [x] Pre-commit passes


src/careamics/losses/lvae/losses.py
```diff
@@ -168,30 +168,30 @@ def get_kl_divergence_loss(
         dim=1,
     )  # shape: (B, n_layers)
 
+    # Apply free bits (& batch average)
+    kl = free_bits_kl(kl, free_bits_coeff)  # shape: (n_layers,)
```
jdeschamps, line 172: There seem to be two definitions of this function in the current code base:

- https://github.com/federico-carrara/careamics/blob/b29fc6c95c65026c6169fbb35f1f9b4f31383568/src/careamics/losses/lvae/loss_utils.py#L4
- https://github.com/federico-carrara/careamics/blob/b29fc6c95c65026c6169fbb35f1f9b4f31383568/src/careamics/models/lvae/utils.py#L407

Could you take this opportunity to clean that up?

jdeschamps (changes requested):

src/careamics/losses/lvae/losses.py
federico-carrara: Thanks for noticing.  I will clear it up :)

jdeschamps (approved):

Issue #278, opened 2024-11-18
Review and update BMZ export


melisande-c:
List of TODOs for exporting to BMZ. Once these are completed we can create a new CAREamics release re-run our notebooks and re-upload the new versions of the models.

- [x] Review README creation
  - [x] Review information order, e.g. should data description come before algorithm configuration?
  - [x] Add validation description, as discussed in #276 
  - [x] Update links to docs (they are broken, `/latest` doesn't exist)
- [x] We need to add the covers, it seems there is no automatic cover generation, issue also open here #176 
- [x] We should relax the test kwargs, absolute and relative tolerance, this can be done by adding them to the rdf `config.bioimageio.test_kwargs`. This is where they are extracted during the CI tests: https://github.com/bioimage-io/collection/blob/efb7f06797574544385222de62b08f13ae99854d/bioimageio_collection_backoffice/run_dynamic_tests.py#L109-L115.

melisande-c:
Additional point not to forget:

- [x] Formatting error: "algorithm description" should be a markdown heading, the same as for example the data description.

mese79:
Just to share my experience, even though I've put a `### Validation` part into my model's `doc.md` file, still I've got the warning `No '# Validation' (sub)section found in doc.md.` .  
I haven't try to change the `### Validation` to `# Validation`, but the check shouldn't be such meticulous and limited to only the `h1` header.

melisande-c:
In the documentation you shared in #276 it says

> '#[#] Validation' (sub)section with details on how to quantitatively validate the model on unseen data.

I assume this means it can be `h1` or `h2`

jdeschamps:
So half meticulous. :)

melisande-c:
Something we should also think about is the model versioning, it seems we hardcode the model version to be 0.1.0, but if a model needs to be updated I am not sure how the bmz handles this (I think different model versions are saved in different folders in the s3 bucket, but if the version is the same maybe it gets overwritten).

Below is where we hardcode the version:
https://github.com/CAREamics/careamics/blob/cc4cc04368e5286591266f9df2489546b656e826/src/careamics/model_io/bioimage/model_description.py#L294

mese79:
This is the issue Caterina asked community partners to add some documentation:
https://github.com/bioimage-io/bioimage.io/issues/437

PR #279, opened 2024-11-18
Fix(BMZ): Relax model output validation kwargs; extract weights and config file following new `spec` and `core` release

melisande-c:
### Description

- **What**: Relaxing the model output validation kwargs, both absolute and relative tolerance, from the default, `1e-4`, to `1e-2`.
- **Why**: The defaults are pretty strict and some of our uploaded models are stuck in pending because of slightly mismatching input and outputs.
  - e.g.
  ```console
  Output and expected output disagree:
    Not equal to tolerance rtol=0.0001, atol=0.00015
  Mismatched elements: 40202 / 1048576 (3.83%)
  Max absolute difference: 0.1965332
  Max relative difference: 0.0003221
  ```
- **How**: In the model description config param, added the new test kwargs. Additionally, updated `bmz_export` so that the test kwargs in the model description are used during model testing at export time.

### Changes Made

- **Modified**: Describe existing features or files modified.
  - `create_model_description`: added test_kwargs to config param
  - `export_to_bmz`: use test_kwargs in model description for model testing at export time.

### Related Issues

- Resolves 
  - last checkbox in #278 

EDIT:
This PR also fixes loading from BMZ following an incompatible release of `bioimageio/core` (`0.7.0`) and `bioimageio/spec` (`0.5.3.5`). The problem was `load_model_description` no longer unzips the archive file but only streams the `rdf.yaml` file data. This means we have to now extract the weights and careamics config from the zip to load them, which can be done using `bioimageio.spec._internal.io.resolve_and_extract`

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/model_io/bioimage/model_description.py
+++ b/src/careamics/model_io/bioimage/model_description.py
@@ -280,6 +280,16 @@ def create_model_description(
             "https://careamics.github.io/latest/",
         ],
         license="BSD-3-Clause",
+        config={
+            "bioimageio": {
+                "test_kwargs": {
+                    "pytorch_state_dict": {
+                        "absolute_tolerance": 10e-2,
+                        "relative_tolerance": 10e-2,
+                    }
+                }
+            }
+        },
         version="0.1.0",
         weights=weights_descr,
         attachments=[FileDescr(source=config_path)],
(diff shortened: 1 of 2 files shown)


jdeschamps (approved):

melisande-c:
Think the failing tests are caused by a new `bioimageio/core` and `bioimageio/spec` release, I will look into it.

Additionally, should I set the absolute tolerance to 0? (Would mean only the relative tolerance is used). In the example above (from a model pending in the bmz) it has a largest absolute tolerance of `0.1965332` but a largest relative tolerance of `0.0003221`...

melisande-c:
See edit about load from bmz fix

PR #285, opened 2024-11-29
Noise models for paper notebooks

CatEek:
### Description

Small refac for submission notebooks. Training code works if called explicitly. Training inside careamics needs further engineering. 
NM code needs refactoring

### Changes Made

- **Added**: 
- Basic training
- plotting function

- [x ] Code builds and passes tests locally, including doctests
- [x ] Pre-commit passes


src/careamics/models/lvae/noise_models.py
```diff
@@ -12,6 +13,67 @@
 # TODO this module shouldn't be in lvae folder
 
 
+def create_histogram(bins, min_val, max_val, observation, signal):
```
jdeschamps, line 15: No return annotation

src/careamics/models/lvae/noise_models.py
```diff
@@ -2,6 +2,7 @@
 
 from typing import TYPE_CHECKING, Optional
 
+import matplotlib.pyplot as plt
```
jdeschamps: So far we have always avoided depending on matplotlib directly. This makes the CI fail on this PR. 

In addition, I am not sure it should be the responsibility of the NM to plot itself, I'd rather have this type of function in a utility class rather than buried in essential code. Details, the first comment is more important.

jdeschamps (changes requested):

src/careamics/models/lvae/noise_models.py
CatEek: Removed matplotlib as discussed

jdeschamps (approved):

PR #288, opened 2024-12-04
Fix: fix device array handling in NoiseModels

CatEek:
related to #285, missed commit


src/careamics/models/lvae/noise_models.py
```diff
@@ -564,6 +559,7 @@ def train_nm(
             )
             signals = torch.from_numpy(signals).float().to(self.device)
 
+            # TODO put all fucking tensors on gpu !!!
```
jdeschamps: can you sanitize your comments?

jdeschamps (changes requested):

jdeschamps (approved):

PR #344, opened 2025-01-09
refac: Algorithm specific configurations

### LINKED ISSUE #320: Better pydantic configuration organization
## Description

Currently, we have a single `Configuration` pydantic model, and it covers all the algorithms. There is a split only in the model configurations (`FCC` vs `VAE`). This means that we have a lot of validation in different part of the configuration to check for the specificities of N2V for instance.

Note that we are not using the pydantic mypy plugin, so we are not getting objective typing static checking for the models...

### Which part of the code?

- https://github.com/CAREamics/careamics/blob/main/src/careamics/config/configuration_model.py

Specifically:
https://github.com/CAREamics/careamics/blob/cd2eae26252b8ef0618b5b7e6fcdac3f5a58d8f9/src/careamics/config/configuration_model.py#L349

### Better type checking

Add mypy pydantic plugin.

### Split configurations 

Make algorithm-specific pydantic configurations so as to offload all validation to pydantic defaults, and not to have to constantly implement `field_validators` in every model.

This will also solve most mypy errors once the pydantic plugin is properly used.

### Add configuration handler model

The only problem is that the creation of the configuration will be more complex. We need something along the lines of:

```python

class DataModel(BaseModel):
    ...

class N2VDataModel(DataModel):
    ...

class Configuration(BaseModel):
    data_model: DataModel


class N2VConfig(Configuration):
    data_model: N2VDataModel
 

class ConfigHandler(BaseModel):
    config: Union[N2VConfiguration, Configuration]

my_config: dict = ...
config = ConfigHandler(config=my_config).config
```

Specifics to be worked out!

## References

- [mypy pydantic plugin](https://docs.pydantic.dev/latest/integrations/mypy/)

jdeschamps:
### Description

> [!IMPORTANT]  
> **TLDR**: Split Pydantic configurations to simplify validation.

Following https://github.com/CAREamics/careamics/issues/320, this PR adds algorithm-specific main configurations, as well as algorithm and data sub-configurations when necessary (e.g. N2V) in order to simplify the code base. Additionally, it fix mypy handling of Pydantic models using the Pydantic mypy plugin. Finally, the `CustomModel` capacity was removed, fixing errors in the mypy report.

#### Pydantic mypy plugin

Natively, mypy does not have a good handle of Pydantic models. Adding the [Pydantic mypy plugin](https://docs.pydantic.dev/latest/integrations/mypy/#configuring-the-plugin) yields new errors that were hidden from us (see [report here](https://results.pre-commit.ci/run/github/607105530/1729457186.StZU4dBMTrq2a55jjzwmGw)).

#### Custom models

The custom models were added without having a real use-case, and the examples were mostly very simple toy examples. It is likely that any real use case will require more than plugging in a model in the CAREamist API, and will probably start with using one's own Lightning module in the Lightning API.

In addition, the presence of unknown Pydantic models creates issues with mypy (see previous point), and adds complexity in the code for a feature we are not sure will be used. This PR removes the `CustomModel` class and all mechanism to register models.

#### Splitting configurations

Another pain point in the code base (https://github.com/CAREamics/careamics/issues/320) is that having one single `Configuration` for vastly different algorithms (and their constraints) generates complex validation code: validation across sub-configurations, many `if else` statements, and long files with tons of `field_validators`.

An elegant solution is to define a general configuration and sub-class it with algorithm-specific Pydantic models. Then, a "factory" model can leverage Pydantic model selection to choose the correct algorithm-specific configuration based on the parameters. Here is an illustration:

```python
class Configuration(BaseModel):
    # some example parameters
    version: Literal["0.1.0"] = "0.1.0"
    algorithm_config: Union[UNetBasedAlgorithm, VAEBasedAlgorithm]
    data_config: Union[DataConfig, N2VDataConfig]

class N2VConfiguration(Configuration):
    algorithm_config: UNetBasedAlgorithm
    data_config: N2VDataConfig

class HDN(Configuration):
    algorithm_config: VAEBasedAlgorithm
    data_config: DataConfig
```

In this scenario, all the validation related to the `N2VManipulate` is in `N2VDataConfig` and `N2VConfiguration` can freely assume that the correct transforms are present. Similarly, `DataConfig` does not need to care about `N2VManipulate`, and other subclasses can exclude it from the allowed types in its transforms. In the same spirit, the algorithm config is always `UNet` for N2V, and `VAE` for HDN. In this PR, we followed that idea with algorithm specific configurations (both for the main configuration and the algorithm sub configuration).

The main issue raised by this approach is that `Configuration` can no longer be used to instantiate the correct configuration, but rather `configuration_factory` should be used:

``` python
my_cfg = {
    ...
}

# old way
cfg = Configuration(**my_cfg)

# new approach
cfg = configuration_factory(my_cfg)
```

Here, the model choice (`N2VConfiguration`, `CAREConfiguration`, etc...) is delegated to Pydantic via the `configuration_factory`.

Similar factories have been created for the algorithm and data sub-configurations in order to allow correct instantiation in the Lightning API.

### Changes Made

- **Added**: 
    - `N2VConfiguration`: main configuration for N2V
    - `N2NConfiguration`: idem for N2N
    - `CAREConfiguration`: idem for CARE
    - `N2VDataConfig`: data subconfiguration for N2V (enforced `N2VManipulate`)
    - `N2VAlgorithm`: algorithm subconfiguration for N2V
    - `N2NAlgorithm`: idem for N2N
    - `CAREAlgorithm`: idem for CARE
    - `ConfigurationFactory`: a Pydantic model instantiating the correct configuration
    - `AlgorithmFactory`: idem for algorithm subconfigurations
    - `DataFactory`: idem for data subconfigurations
    - `configuration_io.py`: read/save methods for configurations, refactored from `configuration_factory.py`
- **Modified**: 
    - All calls to `Configuration` and `Configuration` itself. 
    - Renamed UNet and VAE algorithm configurations.
- **Removed**: `references` module, all calls to `has_n2v_manipulate`, anything to do with `CustomModel`.

> [!WARNING]  
> Testing could be improve further and there are probably some aspects that I am missing. But I'd rather get reviews already!

### Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/320

### Breaking changes

- Configuration, algorithm subconfig, and data subconfig should be instantiated via the factories.
- Renaming impacts the algorithm subconfigurations (`UNetAlgorithm`, `VAEAlgorithm`).
- `CustomModel` is now completely removed.
- Calls to `has_n2v_manipulate` no longer work, prefer `isinstance(cfg, N2VConfiguration)`.

> [!WARNING]  
> Any current work on adding new algorithm and configuration will be largely impacted by this PR. @CatEek 
---

- [x] Pre-commit passes


src/careamics/config/algorithms/unet_algorithm_model.py
```diff
@@ -0,0 +1,101 @@
...
+    `name` in the custom pydantic model.
+
+    Attributes
+    ----------
+    algorithm : {"n2v", "care", "n2n", "custom"}
+        Algorithm to use.
+    loss : {"n2v", "mae", "mse"}
+        Loss function to use.
+    model : UNetModel or CustomModel
+        Model architecture to use.
```
melisande-c: Docs should be updated to remove reference of custom model

src/careamics/config/configuration_factory.py
```diff
@@ -2,26 +2,68 @@
...
+
+    algorithm: Union[
+        N2VAlgorithm, N2NAlgorithm, CAREAlgorithm, UNetBasedAlgorithm, VAEBasedAlgorithm
+    ]
+
+
+class DataFactory(BaseModel):
+    """A factory to instantiate a data model."""
+
+    data: Union[DataConfig, N2VDataConfig]
```
melisande-c: What are these used for? can't find them being used in the code

src/careamics/config/configuration_io.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    # load dictionary from yaml
+    if not Path(path).exists():
+        raise FileNotFoundError(
+            f"Configuration file {path} does not exist in " f" {Path.cwd()!s}"
+        )
+
+    dictionary = yaml.load(Path(path).open("r"), Loader=yaml.SafeLoader)
+
+    return ConfigurationFactory(configuration=dictionary).configuration
```
melisande-c: This solution works, but can actually use pydantic's `TypeAdaptor` class. e.g.:

```python
from typing import Union, Literal
from pydantic import BaseModel, TypeAdapter

class A(BaseModel):
    value: Literal["a"] = "a"

class B(BaseModel):
    value: Literal["b"] = "b"

def factory(params: Union[A, B]):
    adaptor = TypeAdapter(Union[A, B])
    return adaptor.validate_python(params)

print(factory(params={"value": "b"}))
```
```
B(value='b')
```

src/careamics/config/data/data_model.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from pprint import pformat
 from typing import Annotated, Any, Literal, Optional, Union
```
melisande-c, line 7: Not totally sure if it makes a difference, but shouldn't we be using the `Sequence` class from `typing` and not `collections` for type hinting.

src/careamics/lightning/train_data_module.py
```diff
@@ -659,21 +666,35 @@ def create_train_datamodule(
...
+                # add it
+                data_dict["transforms"].append(n2v_manipulate)
         else:
-            raise ValueError(
-                "Cannot have both supervised training (target data) and "
-                "N2V manipulation in the transforms. Pass a list of transforms "
-                "that is compatible with your supervised training."
-            )
+            # create the transforms
+            data_dict["transforms"] = [n2v_manipulate]
```
melisande-c: We should probably be leveraging the pydantic config to automatically choose the N2VDataConfig, and have the Noise2Void transform added there. Might need an extra discriminator field in the DataConfigs, (with extra argument as mentioned in TODO).

src/careamics/config/data/n2v_data_model.py
```diff
@@ -0,0 +1,190 @@
...
+            elif transform_list[-1] != SupportedTransform.N2V_MANIPULATE:
+                index = transform_list.index(SupportedTransform.N2V_MANIPULATE.value)
+                transform = transforms.pop(index)
+                transforms.append(transform)
+
+        else:
+            raise ValueError(
+                f"{SupportedTransform.N2V_MANIPULATE} transform "
+                f"is required for N2V training."
+            )
```
melisande-c: Maybe `N2V_MANIPULATE` can be added automatically instead of raising an error. It will already modify the order if it's not last.

src/careamics/config/data/n2v_data_model.py
```diff
@@ -0,0 +1,190 @@
...
+        raise ValueError(
+            f"{SupportedTransform.N2V_MANIPULATE} transform "
+            f"is required for N2V training."
+        )
+
+    def set_structN2V_mask(
+        self, mask_axis: Literal["horizontal", "vertical", "none"], mask_span: int
+    ) -> None:
+        """
+        Set structN2V mask parameters.
```
melisande-c, line 146: Maybe N2VManipulate params can also be set at initialisation by having `n2v_manipulate_config` as a field of N2VDataConfig e.g.
```python
class N2VDataConfig(GeneralDataConfig):
    """N2V specific data configuration model."""

    n2v_manipulate_config: N2VManipulateModel = N2VManipulateModel()

    transforms: Sequence[N2V_TRANSFORMS_UNION] = Field(
        default_factory = lambda data: [
            XYFlipModel(), XYRandomRotate90Model(), data["n2v_manipulate_config"]
        ],
        validate_default=True,
    )
```
Then one can initialise `N2VDataConfig` in the following way and the n2vmanipulate params will be automatically populated and validated:
```python
N2VDataConfig(
    n2v_manipulate_config={
        "strategy": "median", 
        "struct_mask_axis": "horizontal", 
        "mask_span": 7
    }
)
```

melisande-c (commented):
Looks good! Will def make maintenance easier in future. Added a few small comments and suggestions. (I didn't look at tests yet)

src/careamics/config/configuration_factory.py
melisande-c: nvm see next comment though with suggestion to use pydantic's `TypeAdaptor` class

src/careamics/config/algorithms/unet_algorithm_model.py
jdeschamps: Good point, skipped those!

src/careamics/config/configuration_factory.py
jdeschamps: They are used in the Lightning API related modules to create the algorithm and data configurations (since the users do not create a global configuration).

One such example is `create_train_datamodule` (or similar).

src/careamics/config/data/data_model.py
jdeschamps: hmm this must have been done by a pre-commit or VSCode, because I definitely used the typing one when writing the code.

src/careamics/lightning/train_data_module.py
jdeschamps: I am not sure I understand! In my mind that's what is happening here, since discrimination is done using the `N2VManipulate`.

Now there is something wrong here: rather than creating the default `N2V` transforms, this removes all transforms if no target and no transforms are passed.... And that's wrong!

Maybe that's what you hinted at!

src/careamics/config/data/n2v_data_model.py
jdeschamps: I thought about it and this was the previous behaviour.

I think we should avoid as much as possible modifications done by Pydantic validators to what users are passing and errors should be thrown instead. The responsibility to maintain things coherent should not be borne by the Pydantic model, otherwise we will end up with very complex validation and silently modifying user input!

Now that means that the coherence should be ensured by any convenience function that instantiate these models. So in the end:
- expert users can create the models directly
- non-experts use convenience functions that do the heavy lifting to ensure proper instantiation

src/careamics/config/data/n2v_data_model.py
jdeschamps: I am not sure I understand what's happening there, but it is difficult to read. I need to play around with it! Need to check how `mypy` reacts to that.

What I like about this is that in the future we might get rid altogether from `N2VManipulate` as a transform, but we will still need the parameters to be stored somewhere.

But now I am wondering what happens if users pass `n2v_manipulate_config` and an instance of `N2VManipulateModel` in the transforms... If both are passed without errors, that means we need a complex model validation to decide which one has priority.

src/careamics/config/configuration_io.py
jdeschamps: Ah interesting, I have never used `TypeAdapter`. Will give it a go!

src/careamics/config/data/n2v_data_model.py
melisande-c: Fair enough, but then maybe the order of the transforms shouldn't be modified either?

src/careamics/config/data/n2v_data_model.py
melisande-c: basically the `default_factory` arg in `Field` can either be a callable that takes no arguments, or the previously validated arguments as a dictionary, this is the `lambda` func (would probably clearer to write a separate type hinted function).

However I now realise that, if the transforms argument is passed at initialisation it will overwrite the default so the n2v_config will not be copied over... so yeah the default factory solution doesn't work...

Maybe something better would be:
```python
class N2VDataConfig(GeneralDataConfig):
    """N2V specific data configuration model."""

    n2v_manipulate: N2VManipulateModel = N2VManipulateModel()

    # TODO: think of better name
    other_transforms: Sequence[TRANSFORMS_UNION] = Field(
        default = [
            XYFlipModel(), XYRandomRotate90Model()
        ],
        validate_default=True,
    )

    @computed_field
    @property
    def transforms(self) -> Sequence[N2V_TRANSFORMS_UNION]:
        return self.other_transforms + [self.n2v_manipulate]
```
This solution also removes the need to validate that the n2v manipulate transform is in the transforms list

src/careamics/config/data/data_model.py
jdeschamps: Probably pre-commit (ruff?) because `typing.sequence` is deprecated (see [official doc](https://docs.python.org/3/library/typing.html#typing.Sequence)), and `collections.abc.Sequence` should be used.

src/careamics/config/data/n2v_data_model.py
jdeschamps: Maybe yes, it is such an edge case anyway that we don't gain much by correcting that!

src/careamics/config/data/n2v_data_model.py
melisande-c: ah just been experimenting and get:
```
ValueError: you can't override a field with a computed field
```

jdeschamps: Maybe we then leave it like that for now, especially since refactoring `N2V` might impact these changes anyway.

melisande-c: Ok because of the inheritance can't overwrite transforms but could do something like:
```python
class N2VDataConfig(GeneralDataConfig):
    """N2V specific data configuration model."""

    n2v_manipulate: N2VManipulateModel = N2VManipulateModel()

    transforms: Sequence[TRANSFORMS_UNION] = Field(
        default = [
            XYFlipModel(), XYRandomRotate90Model()
        ],
        validate_default=True,
    )

    def get_transforms(self) -> Sequence[N2V_TRANSFORMS_UNION]:
        return self.other_transforms + [self.n2v_manipulate]
```

Of course then we should only access transforms by calling `get_transforms`. `DataConfig` will also need the same method. Some renaming could be considered to discourage accessing transforms.

src/careamics/lightning/train_data_module.py
melisande-c: Ok I did write this confusingly, was agreeing with the TODO. If the data configs have a discriminator field then there can be an additional argument like `data_config_type` that will allow pydantic to automatically select the correct data config.

src/careamics/config/data/n2v_data_model.py
melisande-c: But yeah can leave for now, and think about how to change in future, especially if N2VManipulate will no longer be a transform in the near future

src/careamics/lightning/train_data_module.py
melisande-c: but just a note for the future

jdeschamps: Actually there was a simple way to do things correctly: instantiate the correct config based on the presence of a target. 

We will revist later fotr HDN I guess.

src/careamics/config/configuration_io.py
jdeschamps: Reading up the [doc](https://docs.pydantic.dev/latest/concepts/type_adapter/), I am not sure I see the point of using `TypeAdapter` in this case as all types are `BaseModels`. I find it clearer the way it is now. Unless maybe I am missing something?

I do like, however, the idea of wrapping it in a function.

jdeschamps: Updated with the wrapping functions

melisande-c: Ok ngl I didn't really read the docs, I just saw these two posts on stackoverflow: [1](https://stackoverflow.com/questions/76537360/initialize-one-of-two-pydantic-models-depending-on-an-init-parameter), [2](https://stackoverflow.com/questions/78983916/implementing-discriminated-unions-in-pydantic-without-using-nested-models).

Is a stylistic choice I guess. I find defining a whole class just to instantiate other classes a little weird, but happy to leave as is, I agree `TypeAdapter` isn't the clearest.

jdeschamps: Sleeping on it, I don't really have a strong opinion. Current solution feels easier to ready but indeed slightly convoluted, `TypeAdapter` are an edge case feature of Pydantic but have the advantage of being more lightweight and have a smaller footprint...

It does not feel like we should agonize over something that makes little difference. Since the `TypeAdapter` reduces the number of lines, I went for your suggestion!

melisande-c (approved):
Still haven't looked at tests but I'm sure they're good 😄

CatEek (approved):

PR #361, opened 2025-01-17
feat: Add algorithm model validators

### LINKED ISSUE #354: Validate UNet `final_activation` in N2V, CARE and N2N algorithm configurations
## Describe the bug

Currently, users could set the `final_activation` parameter in the`UNetModel`, which is actually not compatible with denoising with N2V, CARE and N2N.

https://github.com/CAREamics/careamics/blob/8f03c885111f602252c8b1e969a0318792898c01/src/careamics/config/architectures/unet_model.py#L52

## Solution

We should validate this parameter in the algorithm configurations.

jdeschamps:
### Description

Following https://github.com/CAREamics/careamics/issues/354: `final_activation` was allowed to be set although it would break CARE, N2N and N2V. 

This PR adds a Pydantic validator preventing final activation for these algorithms. It also takes the opportunity to improve the Pydantic models by moving the validators to a different module and using annotations for cleaner code.

- **What**: CARE, N2N and N2V do not allow final activation in the UNet.
- **Why**: This would otherwise break the algorithm training.
- **How**: Add model validators in the respective Pydantic models.

### Changes Made

- **Added**: `model_validators.py`.
- **Modified**: Algorithms model.

### Related Issues

- Fixes https://github.com/CAREamics/careamics/issues/354

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/algorithms/care_algorithm_model.py
+++ b/src/careamics/config/algorithms/care_algorithm_model.py
@@ -1,10 +1,14 @@
 """CARE algorithm configuration."""
 
-from typing import Literal
+from typing import Annotated, Literal
 
-from pydantic import field_validator
+from pydantic import AfterValidator
 
 from careamics.config.architectures import UNetModel
+from careamics.config.validators import (
+    model_without_final_activation,
+    model_without_n2v2,
+)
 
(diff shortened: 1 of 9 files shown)


melisande-c (approved):
Nice! the annotated types are pretty neat

PR #365, opened 2025-01-21
refac: Move N2V masking in lightning module

CatEek:
### 

- **What**: Rewrote all N2V related transforms in torch and moved them to lightning module. It's now applied over batch dimension. Seems like it's also faster now 
- **Why**:  Them being in the dataset created unnecessary complication for the dataset classes and problem integrating LVAE based algos
- **How**: With my crabby claws, unfortunately 

### Changes Made

- **Added**: N2VMAnipulate is now in the preprocess param of the lightning module. It is a function that is applied to a batch before inputting it to the model.
- **Modified**: Many files
- **Removed**: Some old tests

- Only tested N2V performance on BSD, still need to test others. 
- There're a bunch of todos related to some ugly hack here and there. Needs to be addressed later
- Need to update docs
---

- [x] Pre-commit passes


jdeschamps:
There are missing docs in the torch N2V manipulate: https://results.pre-commit.ci/run/github/607105530/1737501613.FBC6iWDHTgWR3WhJeX4o9A

jdeschamps:
It's seems the failing of tests because of a `NaN` array is happening a lot with this PR...

src/careamics/config/algorithms/n2v_algorithm_model.py
```diff
@@ -16,14 +19,77 @@
...
 
+    model_config = ConfigDict(validate_assignment=True)
+
     algorithm: Literal["n2v"] = "n2v"
     """N2V Algorithm name."""
 
     loss: Literal["n2v"] = "n2v"
     """N2V loss function."""
 
+    n2v_masking: N2VManipulateModel = N2VManipulateModel()
```
melisande-c: I think this would be clearer if it was called n2v_config

src/careamics/config/algorithms/n2v_algorithm_model.py
```diff
@@ -16,14 +19,77 @@
...
+                    f"{SupportedPixelManipulation.MEDIAN} pixel manipulation strategy"
+                    f". Change the N2VManipulate transform strategy."
+                )
+        else:
+            if self.n2v_masking.strategy != SupportedPixelManipulation.UNIFORM.value:
+                raise ValueError(
+                    f"N2V can only be used with the "
+                    f"{SupportedPixelManipulation.UNIFORM} pixel manipulation strategy"
+                    f". Change the N2VManipulate transform strategy."
+                )
```
melisande-c: refer to the pydantic field name in the error message. (and maybe don't mention transforms)

src/careamics/config/__init__.py
```diff
@@ -12,16 +12,15 @@
     "CheckpointModel",
     "Configuration",
     "DataConfig",
+    "DataConfig",
```
jdeschamps: Is it duplicated here because of a merge maybe?

jdeschamps:
@melisande-c @veegalinova 

To help with the review, because of merge conflicts with other PRs to main, two aspects were conflated in this PR:

- Igor's work: moving N2V manipulate transform into the FCC lightning module and rewriting the transform into `torch`
- My merges and fixes: reworking on the configuration factories to have a better typing and returned types. This means:
    - Some validations moved from `N2VConfiguration` to `N2VAlgorithm`
    - Some classes disappeared: `N2VDataConfig`, `GeneralDataConfig`, which also removed the need for some checks in datasets
    - N2N and CARE configurations do not specify anymore the `DataConfig` (same for all algorithms)
    - configuration factories have improved Pydantic validation with `Discriminator` and `Field(discriminator=...)`.
    - Some return declarations had to change to fix `mypy` errors.
    - The convenience functions to create the lightning modules have had changes in parameters, where all the N2V parameters moved from `create_train_dataset` to `create_train_module`.

Unfortunately, some of the changes here are already in main, but simply appear because of the merge conflicts.

src/careamics/lightning/lightning_module.py
```diff
@@ -119,10 +133,14 @@ def training_step(self, batch: Tensor, batch_idx: Any) -> Any:
...
         """
-        # TODO can N2V be simplified by returning mask*original_patch
-        x, *aux = batch
-        out = self.model(x)
-        loss = self.loss_func(out, *aux)
+        x, *targets = batch
+        if self.algorithm == "n2v":
+            x_preprocessed, *aux = self.preprocess(x)
+        else:
+            x_preprocessed = x
```
jdeschamps, line 143: In `self.preprocess`, non-N2V algorithm returns `Identity`. Why is it not used here?

src/careamics/lightning/lightning_module.py
```diff
@@ -179,8 +202,13 @@ def predict_step(self, batch: Tensor, batch_idx: Any) -> Any:
 
         if is_tiled:
             x, *aux = batch
+            if type(x) in [list, tuple]:
```
jdeschamps, line 211: Could you comment on on what the situation here? When is `type(x) in [list, tuple]`?

src/careamics/lightning/lightning_module.py
```diff
@@ -179,8 +202,13 @@ def predict_step(self, batch: Tensor, batch_idx: Any) -> Any:
...
+                x = x[0]
         else:
-            x = batch
+            if type(batch) in [list, tuple]:
+                x = batch[0]  # TODO change, ugly way to deal with n2v refac
```
jdeschamps, line 215: Can you comment on potential solutions?

src/careamics/transforms/pixel_manipulation_torch.py
```diff
@@ -0,0 +1,406 @@
...
+    np.ndarray
+        Array of coordinates of the masked pixels.
+    """
+    if rng is None:
+        rng = torch.default_generator
+
+    # TODO explain this
+    mask_pixel_distance = round((100 / mask_pixel_perc) ** (1 / len(shape)))
+
+    pixel_coords = []
```
jdeschamps, line 139: Any chance you could add comments here to explain the various steps for future us?

src/careamics/transforms/pixel_manipulation_torch.py
```diff
@@ -0,0 +1,406 @@
...
+    Returns
+    -------
+    tuple[torch.Tensor, torch.Tensor]
+        tuple containing the manipulated patch and the corresponding mask.
+    """
+    if rng is None:
+        rng = torch.default_generator
+        # TODO do we need seed ?
+
+    transformed_patch = patch.clone()
```
jdeschamps, line 232: Same comment here about adding explanations on what is happening.

src/careamics/transforms/pixel_manipulation_torch.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Returns
+    -------
+    tuple[torch.Tensor, torch.Tensor, torch.Tensor]
+           tuple containing the manipulated patch, the original patch and the mask.
+    """
+    if rng is None:
+        rng = torch.default_generator
+
+    transformed_patch = patch.clone()
```
jdeschamps: Same comment here about adding explanations.

tests/config/test_n2v_preprocessing.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+
+from careamics.config import N2VAlgorithm
+from careamics.config.support import (
+    SupportedStructAxis,
+    SupportedTransform,
+)
+from careamics.transforms import get_all_transforms
+
+# TODO name is confusing, and the tests are probably not in the right place. They should
```
jdeschamps, line 10: I agree with my own TODO comment. These tests don't belong here and might not be needed.

tests/test_careamist.py
```diff
@@ -1137,3 +1137,19 @@ def test_read_logger(tmp_path, minimum_n2v_configuration):
     assert len(losses) == 4
     for key in losses:
         assert len(losses[key]) == config.training_config.num_epochs
+
+
+def test_all_parameters_used(tmp_path, minimum_n2v_configuration):
```
jdeschamps, line 1142: What is the point of that tests? 

If there is not test to check whether the configuration is properly saved, then wouldn't `config == careamist.cfg` work?

Unless we want to make sure that nothing in the configuration changes.

jdeschamps (changes requested):

src/careamics/transforms/compose.py
```diff
@@ -82,21 +80,13 @@ def _chain_transforms(
...
-                patch, *_ = params
-                params = t(patch=patch)
-            else:
-                *params, _ = t(*params)  # ignore additional_arrays dict
-
-        return params
+            *params, _ = t(*params)  # ignore additional_arrays dict
+
+        # avoid None values that create problems for collating
+        return tuple(p for p in params if p is not None)
```
melisande-c, line 89: What happens here downstream when there is a target vs when there isn't a target? Surely this was already handled somewhere else?

src/careamics/dataset/in_memory_pred_dataset.py
```diff
@@ -80,9 +80,7 @@ def __getitem__(self, index: int) -> NDArray:
 
         Returns
         -------
-        NDArray
+        (numpy.ndarray, numpy.ndarray or None)
```
melisande-c: This does not match type hint output which is `tuple of numpy.ndarray`

src/careamics/lightning/lightning_module.py
melisande-c: also can use `isinstance(x, (list, tuple))`

src/careamics/lightning/lightning_module.py
melisande-c: I agree, then can remove the "`if self.algorithm == "n2v"`"

src/careamics/lightning/lightning_module.py
melisande-c: but I am also confused, is this with targets vs without? Is this not already accounted for above?

melisande-c (commented):
Haven't looked at everything fully, left a few comments

src/careamics/config/__init__.py
CatEek: Probably

src/careamics/lightning/lightning_module.py
CatEek: After we finalize the dataset, we can be sure about the structure

CatEek:
@melisande-c @jdeschamps @veegalinova Thanks for the comments, I'll take a look. Another thing is there's probably some inconsistencies in the configuration factories. Like I'm not sure N2V transform should be defined there

src/careamics/transforms/compose.py
CatEek: This is to deal with collate being unable to stack None

src/careamics/config/algorithms/n2v_algorithm_model.py
CatEek: Sorry, not sure what you mean

tests/test_careamist.py
CatEek: I don't know, with all the tests we have I was able to make a bug where while batch size was passed, it wasn't actually used

src/careamics/lightning/lightning_module.py
CatEek: Because in other cases input to the Identity can be a list, which won't work. This whole construction is quite ugly, and definitely needs revisiting.

src/careamics/config/algorithms/n2v_algorithm_model.py
melisande-c: Make the error message something like: 
    "N2V2 can only be used with the {SupportedPixelManipulation.MEDIAN} pixel manipulation strategy. Change the the masking strategy in `n2v_masking` ." 

It will be clearer to users since pydantic allows nested objects to be passed as dictionaries as well as the `N2VManipulateModel` object itself.

src/careamics/transforms/compose.py
melisande-c: Ok sure I see now it should be fine, I was just confused why it had to change since it was already working before but I guess something else changed somewhere else.

melisande-c: Ok, I now see in the old code the None values were removed in the case of n2v by how the parameters were unpacked

src/careamics/lightning/lightning_module.py
melisande-c: I've just tested it and I get 

```
>>> torch.nn.Identity()([torch.tensor([1,2,3])])
[tensor([1, 2, 3])]
````

This is the case where `targets` are a list?

I've also noticed this function is typed incorrectly batch should have the type `list[torch.Tensor]` I believe.

jdeschamps:
A summary of the latest changes since the reviews:

- Merging main and fixing the conflicts
- Implementing @melisande-c's comment https://github.com/CAREamics/careamics/pull/365#discussion_r1926941571
- Fixed a problem preventing CAREamics to run on macOS intel on CPU (forcing MPS on macOS)
- Removed unused pre-processing factory, note that `mypy` forced me to add if statements...

There are some stuff that will need refactoring or splitting into multiple functions, but better open issues and deal with them later at this point.

jdeschamps (approved):

Issue #292, opened 2024-12-06
General framework for dataset, format-agnostic and compatible with NGFF


jdeschamps:
## Description

> [!IMPORTANT]
> Summary: replacing current datasets by a general framework compatible with NGFF file formats.

Currently, we have accumulated a lot of technical debt on our datasets for the sake of simplifying the code. We have ended up with two different types of datasets (InMemory and PathIterable), which have different patching strategies (sequential vs random) and are different types of datasets. Furthermore, the train and prediction datasets have been split because of the tiling in the prediction.

This creates a lot of complexity and different behaviours depending on hyperparameters. The fact that every time we discuss datasets we are confused about which one does what is a bad sign!

To allow additional file formats, we also provide parameters to pass extensions and file reader callables. While this is a nice idea, it is very limiting and will not scale well.

As we are moving towards compatibility with NGFF, we need a better and more general dataset system that can be compatible with multiple back-ends. In addition, this would help third-parties implement their own file format. 

### Features
- Dataset
    - Tiff, Zarr, numpy array
    - Optional targets
    - Add a maximum number of iterations
- Patcher
    - Exclude background coordinates (here or in sampler?)
    - Random patching for training
    - Save train / val coordinates for reproducibility?
    - Seed-able
    - Possibility to fix coordinates (sequential?) for validation
    - Multi-file patching
- Sampler
- Data statistics
    - Welford mean/std
    - Quantile normalization
    - No normalization
    - Pre-computed
- Metadata extractor
    - returns shape of each file for a number of files
    - if metadata is unavailable, defaults to reading the file
- Augmentations
    - Normalization handling
    - Random rotate 90 degrees
    - Random X or Y flip
    - Add noise

### Which part of the code?

- https://github.com/CAREamics/careamics/blob/main/src/careamics/dataset/in_memory_dataset.py
- https://github.com/CAREamics/careamics/blob/main/src/careamics/dataset/iterable_dataset.py
- other datasets in https://github.com/CAREamics/careamics/tree/main/src/careamics/dataset

## Potential solutions

@veegalinova and @CatEek have worked out a potential architecture, please post the schematic here for reference once we will be more confident this is the way to go.

## Relevant issues

https://github.com/CAREamics/careamics/issues/293: Exclude background patches during training.
https://github.com/CAREamics/careamics/issues/282: Why the difference between patching (random vs sequential)?
https://github.com/CAREamics/careamics/issues/297: Should we save train/val patches info for reproducibility?
[...]

PR #373, opened 2025-01-23
Feature(Next-Gen Datasets): Patch extraction agnostic to underlying data storage

melisande-c:
## Description

Opening this draft PR to discuss work so far on the `PatchExtractor` class for the new dataset development. There are no docs or tests for the time being until we have agreed on the final structure!

I have deviated slightly from what was discussed in our meeting, but I hope the design makes sense, otherwise very happy to take suggestions!

This will be an iterative process, so I'm sure changes will be made to these classes in the future, just hope this is a decent starting point!

### `ImageStack` class 
- An interface to extract patches from an array type objection. Must implement the `extract_patch` method. Must have a `data_shape` attribute and a `source` attribute. (The `source` attribute is not used yet but I thought it might be handy to keep track of where data is coming from).

- **Concrete implementations**:
    - `InMemoryImageStack`: The underlying array is a numpy array. Has constructor class methods to initialise from a tiff file, an existing numpy array, or a custom file type.
    - `ZarrImageStack`: The underlying array should point to a zarr array stored on disk. (Not implemented yet!)

### `PatchExtractor` class
- contains multiple `ImageStack` classes and can extract patches from each of them.

### `RandomPatchSpecsGenerator`
- Not settled on the structure of this, implemented it more as of a proof of concept. Should probably be kept decoupled from the `ImageStack` classes and the `PatchExtractor`.
- Ported the existing random patching logic to the `generate` method, but now only the `PatchSpecs` ("patch specifications") are generated.

`PatchSpecs` is a typed dict that can be unpacked in the `PatchExtractor.extract_patch` method or a sequence of `PatchSpecs` can be passed to the `PatchExtractor.extract_patches` method.

### Patch extractor factory

Added a `dataset_ng/patch_extractor/patch_extractor_factory.py` file which contains a `create_patch_extractors` function. 
- Chooses the correct `PatchExtractor` constructor according to information in the `DataConfig`.
- Takes the train data and optional val data, train target data and val target data and returns corresponding `PatchExtractors`.
- Probably subject to change (probably want to separate the training set creation from the validation set creation).
- But some version of this will be called within the new dataset initialisation.

I have also included two demo notebooks: 
- `dataset_ng/demo_data_reader.ipynb`
- `dataset_ng/patch_extractor_factory_demo.ipynb`

### Related Issues

Part of the development of #292

- A note on organisation: I have put everything in a new directory `dataset_dev` so it wouldn't get too confusing, if people agree?
EDIT: has been moved to `dataset_ng`

- Obviously pre-commit does not pass atm because there are no docs, but I made sure mypy passes. 
EDIT: ruff now ignores docstrings for file in the `dataset_ng` directory. numpy docstring hook has been disabled for `dataset_ng` entirely.

EDIT: Name updates and *Patch extractor factor* section.

---

- [x] Pre-commit passes


melisande-c:
Main question I have to everyone more familiar with the zarr library: will it be possible to implement the `ZarrArrayReader` as I have envisioned? Will it get complicated with when/where the zarr files are opened? (Asking only from my experience with working with HDF5 files.) Are there any performance concerns?

@CatEek @veegalinova

jdeschamps:
Awesome! It is starting!! 🎉 

>     * A note on organisation: I have put everything in a new directory `dataset_dev` so it wouldn't get too confusing, if people agree?
> 

Makes total sense! I'd suggest the name `ng_dataset` for "next-generation dataset" to be fancy, but this is totally fine as it. 😄

src/careamics/dataset_dev/demo_data_reader.ipynb
```diff
@@ -0,0 +1,123 @@
+{
```
jdeschamps: If we are adding back notebooks to the repo, it might be worth also adding back the pre-commit that removes all outputs from notebooks to avoid committing binaries (images).

src/careamics/dataset_dev/data_reader/array_reader/array_reader_protocol.py
```diff
@@ -0,0 +1,17 @@
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Literal, Protocol, Union
+
+from numpy.typing import NDArray
+
+
+class ArrayReader(Protocol):
+
+    # TODO: not sure how compatible using Path will be for a zarr array
```
jdeschamps: I guess a path going to an internal group will not be a valid path.

Not elegant, but if were to pass such a path, we could infer which is the path to the zarr and split the internal path.

src/careamics/dataset_dev/data_reader/array_reader/array_reader_protocol.py
```diff
@@ -0,0 +1,17 @@
...
+    #   (for a zarr array need to specify file path and internal zarr path)
+    source: Union[Path, Literal["array"]]
```
jdeschamps: Would it make sense to have the source be `Path | np.ndarray`?

Might simplify a tad the parameters to the InMemoryReader

src/careamics/dataset_dev/data_reader/array_reader/in_memory_array_reader.py
```diff
@@ -0,0 +1,51 @@
...
+        self._data = data
+        self.data_shape: Sequence[int] = self._data.shape
+
+    def extract_patch(
+        self, sample_idx: int, coords: Sequence[int], extent: Sequence[int]
+    ) -> NDArray:
+        if len(coords) != len(extent):
+            raise ValueError("Length of coords and extent must match.")
+        # TODO: test for 2D or 3D?
+        return self._data[
```
jdeschamps: if you put the type ignore on line 26, does it remove the need for the others? I never remember how it works.

src/careamics/dataset_dev/data_reader/array_reader/in_memory_array_reader.py
```diff
@@ -0,0 +1,51 @@
...
+class InMemoryArrayReader:
+
+    def __init__(self, source: Union[Path, Literal["array"]], data: NDArray):
+        self.source: Union[Path, Literal["array"]] = source
+        # data expected to be in SC(Z)YX shape, reason to use from_array constructor
+        self._data = data
+        self.data_shape: Sequence[int] = self._data.shape
+
+    def extract_patch(
+        self, sample_idx: int, coords: Sequence[int], extent: Sequence[int]
```
jdeschamps: I understand `extent`, I'd use `patch_size` to stick with the configuration.

jdeschamps (approved):

src/careamics/dataset_dev/demo_data_reader.ipynb
melisande-c: The outputs were still removed! But also I will delete this at some point, just thought it was an easier way to share an example than a long markdown code block

src/careamics/dataset_dev/data_reader/array_reader/in_memory_array_reader.py
melisande-c: yeah you do. The numpy type hints for slicing are a bit messed up I guess, because it complained about all 3.

src/careamics/dataset_dev/data_reader/array_reader/in_memory_array_reader.py
melisande-c: Yeah sure, I did think that tbf

src/careamics/dataset_dev/data_reader/array_reader/array_reader_protocol.py
melisande-c: hmm true, I will try it out

melisande-c: Wait no this doesn’t work, the `source` attribute was more intended as a label. The `InMemoryArrayReader` is implemented for both initialising from an array and initialising from a file e.g. tiff. When initialising from a file source is the file path, and the `._data` attribute stores the array, if I do as you suggest for the case of an array then both `source` and `_data` will have to store a copy of the array for the other logic to work

melisande-c: But also as I mentioned in the description, `source` is not actually used for anything, I just thought it might be handy in the future, but I could easily remove it entirely for now

src/careamics/dataset_dev/demo_data_reader.ipynb
jdeschamps: Ah my bad, the pre-commit hook was never removed. ^^

src/careamics/dataset_dev/data_reader/array_reader/array_reader_protocol.py
jdeschamps: Alright! I just found it odd to have both `source` and `data` (or whatever was the name) passed to the `ArrayReader.__init__`.

melisande-c: so my intention is that `InMemoryArrayReader` should never be constructed using the `__init__` method and only `from_array`, `from_tiff` and `from_custom_file_type` should be used. And I would explain this in the docs.

melisande-c: For example:
```python
>>> array = np.arange(36).reshape(6, 6)
>>> array_reader = InMemoryArrayReader.from_array(data=array, axes="YX")
>>> array_reader.extract_patch(sample_idx=0, coords=(2, 2), extent=(3, 3))
```
```
array([[[14, 15, 16],
        [20, 21, 22],
        [26, 27, 28]]])
```

melisande-c: The other option is to have separate classes for the each construction case. But in the interest of reducing code bloat, I decided to go for the classmethod constructors.

src/careamics/dataset_ng/data_reader/array_reader/in_memory_array_reader.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import Literal, Union
+
+from numpy.typing import NDArray
+from typing_extensions import Self
+
+from careamics.dataset.dataset_utils import reshape_array
+from careamics.file_io.read import ReadFunc, read_tiff
+
+
+class InMemoryArrayReader:
```
veegalinova: Could be `InMemoryArray` or `LoadedArray`

src/careamics/dataset_ng/data_reader/array_reader/zarr_array_reader.py
```diff
@@ -0,0 +1,19 @@
+from collections.abc import Sequence
+from pathlib import Path
+
+from numpy.typing import NDArray
+
+
+class ZarrArrayReader:
```
veegalinova: Same as with the `InMemoryArray`, could be `ZarrArray` or `LoadedZarrArray`

src/careamics/dataset_ng/data_reader/data_reader.py
```diff
@@ -0,0 +1,41 @@
+from collections.abc import Sequence
+from pathlib import Path
+
+from numpy.typing import NDArray
+
+from ..patching import PatchSpecs
+from .array_reader import ArrayReader, InMemoryArrayReader
+
+
+class DataReader:
```
veegalinova: Could be `MultiArray` or `MultiFileArray`

veegalinova:
I like this direction a lot! 

A random thought I had while reading the code - what if instead of classmethod initialization, there would be separate functions implementing the data reading part? 
For example, instead of `from_tiff`, there will be:
```python
def load_tiff(path, axes): 
       ... 
       return InMemoryArrayReader(params)
```
For the DataReader, there could be `load_many_arrays`, `load_many_tiffs` and so on. 
This way, an array reader will have a single responsibility, and `InMemoryArrayReader`, `ZarrArrayReader` and `DataReader` will all be equal from the user perspective. Currently, you'll need to remember which reader to use for which data type.

Another random thought / question (I don't have an answer for) - how much logic could there actually be behind `extract_patch` in the zarr array? As I currently understand it, the only difference between it and the numpy array is adding `.compute()` or `.read()` after indexing, depending on the underlying zarr library. In that case, is there actually a need for a different dataset class? Maybe a simple `if` statement in the main `InMemoryArrayReader` is enough? 

I'll also add some naming suggestions (which are also based on the first point), but feel free to discard them.

src/careamics/dataset_ng/data_reader/array_reader/array_reader_protocol.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class ArrayReader(Protocol):
+
+    # TODO: not sure how compatible using Path will be for a zarr array
+    #   (for a zarr array need to specify file path and internal zarr path)
+    source: Union[Path, Literal["array"]]
+    data_shape: Sequence[int]
+
+    def extract_patch(
```
veegalinova: I find `patch` a bit clashing with the actual patching, might be `slice` or `chunk`

veegalinova (approved):

melisande-c:
@veegalinova thanks for the review! I explain a few of my design choices further in response, but happy to change stuff. Also I was struggling with the naming a little, so always happy to take suggestions 😄.

### Classmethond constructors vs. external function constructors.

Having the method bound to the class or not doesn't make too much difference I guess. I chose to use classmethod constructors because I thought it would be easier to communicate that the `InMemoryArrayReader` is not intended to be initialised using its `__init__` method, but instead only with `from_array`, `from_tiff` etc. And it keeps all of the constructors together.

Side note: additional functionality you get when using a constructor classmethod is that, if you inherit from the class then the constructor will still work for the subclass, i.e. it will return the subclass and not the parent class. But I don't think we'll be using this functionality here, so I'm easy either way if people aren't a fan of classmethod constructors.

### DataReader class

Ok yeah `MultiArrayReader` probably does make sense as an alternative name 😄. I'm not sure I totally get your point here but I will make a few comments and I hope to answer any questions! 
- I was considering trying to make the `DataReader`/`MultiArrayReader` and the `ArrayReader` have the same interface but I decided it was too difficult. (If there is only 1 array, `DataReader` can still be used with a list of 1).
- I expect we will write `Dataset` so that users won't really interact with the lower level `DataReader` and `ArrayReader`. The chain of interaction is intended to be User -> `Dataset` -> `DataReader` -> `ArrayReader`. 
- Back onto the classmethod constructors, for the `DataReader` it does make sense to also have an external function I think. If we want to keep the `DataConfig` the same and also the `TrainDataModule` initialisation the same as it is currently then I think we will have something like the following (not polished):
    ```python
    def create_datareaders(
        data_config: GeneralDataConfig,
        train_data: Union[Sequence[NDArray], Sequence[Path]],
        val_data: Optional[Union[Sequence[NDArray], Sequence[Path]]] = None,
        train_data_target: Optional[Union[Sequence[NDArray], Sequence[Path]]] = None,
        val_data_target: Optional[Union[Sequence[NDArray], Sequence[Path]]] = None,
    ) -> DataReader:
    
        # TODO: I can see a way to make this DRYer
        #   - Need to define a `DataReaderConstuctor` protocol
        #   - Then can iterate through a mapping of SupportedData and constructors
    
        if data_config.data_type == SupportedData.ARRAY:
            # --- train reader
            train_reader = DataReader.from_arrays(train_data, data_config.axes)
            # --- val reader
            if val_data is not None:
                val_reader: Optional[DataReader] = DataReader.from_arrays(
                    val_data, data_config.axes
                )
            else:
                val_reader = None
            # --- train target reader
            if train_data_target is not None:
                train_target_reader: Optional[DataReader] = DataReader.from_arrays(
                    train_data_target, data_config.axes
                )
            else:
                train_target_reader = None
            # --- val target reader
            if val_data_target is not None:
                val_target_reader: Optional[DataReader] = DataReader.from_arrays(
                    val_data_target, data_config.axes
                )
            else:
                val_target_reader = None
    
        if data_config.data_type == SupportedData.TIFF:
            ... # similar to above but use DataReader.from_tiff_files
        
        # if zarr
        # if custom
        # etc.
    
        return train_reader, val_reader, train_target_reader, val_target_reader
    ```

### ZarrArrayReader separation

Ok I separated the classes because I have never really used the zarr library but if you think they can be combined then sure we can do it. (We do want to leave room for potential other readers though, e.g. CZI)

I'd also like to make a little point that with this set up we are pretty flexible to designing many patching strategies. For example we could write a "chunked random patch generator" that could be given the chunk sizes so it could include logic to prevent patches being selected that span multiple chunks in chunked data files.

src/careamics/dataset_ng/data_reader/array_reader/array_reader_protocol.py
melisande-c: Sure this is just slicing, but I chose `extract_patch` for the following reason. Currently I have defined the typed dict `PatchSpecs`, that can be unpacked directly into `DataReader.extract_patch` (or a sequence of `PatchSpecs` objects can be passed to `DataReader.extract_patches`). So I was thinking if you provide "patch specifications" to a method called `extract_patch` it would make sense to expect to be returned the corresponding patch. But we can discuss further!

melisande-c:
### Renaming updates!

I changed my mind on the names like 10 times as you can see from the commits...

`ArrayReader` -> `ImageStack`
- I think this now communicates that it's intended to store image data, and now the `sample_idx` argument in the `extract_patches` method makes more sense in the context – it indexes one sample from the image stack, to then extract the patch from.

`DataReader` -> `PatchExtractor`
- I tried `MultiImageStack` but it felt clunky when I was trying to develop what is the now `create_patch_extractors` function. I justify this choice by the fact the class does extract patches, but still open to suggestions.

directories and file names have also been renamed accordingly, additionally the `patching` directory has been renamed to `patching_stategies`.

### Patch extractor factory

I have added the `patch_extractor/patch_extractor_factory.py` module which has a `create_patch_extractors` function. I managed to avoid a giant set of if statements, how I was trying to explain in our meeting, but it probably still needs a little work. Especially organisation, like where to keep the `PatchExtractorConstructor` protocol definition.

Let me know what you think! @veegalinova @jdeschamps  @CatEek

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
```diff
@@ -0,0 +1,83 @@
...
+from careamics.dataset_ng.patch_extractor import (
+    PatchExtractor,
+    PatchExtractorConstructor,
+)
+
+
+def build_patch_extractor_constructor_kwargs(
+    data_config: GeneralDataConfig, **custom_kwargs: Any
+) -> dict:  # TODO: return union of TypedDicts?
+    if data_config.data_type == SupportedData.ARRAY:
```
jdeschamps: Since we are not enforcing `py311`, our `Enums` are not `StrEnum` and we used to have to do `SupportedData.ARRAY.value` to ensure that the comparison was sane.

I am not sure what the status of that, but that's why you will see it through the code. There were cases were the comparison would not work... Can't remember the exact situation right now.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
```diff
@@ -0,0 +1,83 @@
...
+    elif data_config.data_type == SupportedData.CUSTOM:
+        return {"axes": data_config.axes, **custom_kwargs}
+    else:
+        raise ValueError(f"Data type {data_config.data_type} is not supported.")
+
+
+def get_patch_extractor_constructor(
+    data_config: GeneralDataConfig,
+) -> PatchExtractorConstructor:
+    if data_config.data_type == SupportedData.ARRAY:
```
jdeschamps, line 18: If we drop `py39`, we can use `match case` clauses (just planting the seed, nothing actionable right now).

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
```diff
@@ -0,0 +1,83 @@
...
+    Optional[PatchExtractor],
+    Optional[PatchExtractor],
+    Optional[PatchExtractor],
+]:
+
+    # get correct constructor
+    constructor = get_patch_extractor_constructor(data_config)
+
+    # build key word args
+    constructor_kwargs = build_patch_extractor_constructor_kwargs(
```
jdeschamps: Can we pass the `data.axes` and `**custom_kwargs` directly to `constructor` rather than going through this additional method?

jdeschamps (approved):

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
jdeschamps: I find this a bit convoluted here, but you might have intended it for flexibility to more complex cases.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
melisande-c: I just tested this in a python 3.9 env and it works
```python
from enum import Enum

class Letters(str, Enum):
    A = "a"
    B = "b"
    C = "c"

print("a" == Letters.A)
```
prints
```
True
```
I think that now that the supported enums inherit from `str` it works :)

On a side note, I think it would be better if the pydantic class attributes weren't typed as `Literal` but typed as the `Enum` classes. If you want to narrow the accepted `Enum` types (say for sub configurations), then you can also do `Literal[Letters.A, Letters.B]`. I think this way if we add another supported value, then there will be less places to change all the literals, we could keep the narrowed type aliases in the same module as the Support class they refer to so it's easier to change everything at once.

I have found out recently if you do `GenericPydanticModel.model_dump(mode="json")` then any enum types will be converted to just their value.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
melisande-c: yep, I love the `match case` blocks for this kind of situation!

melisande-c: I did this just in case future SupportedTypes need additional parameters from the data config, or don't need the axes parameter. But it's true this is unnecessary for now, I can remove it.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
melisande-c: Yeah I did this just in case future `SupportedTypes` (zarr) need additional parameters from the data config, or don't need the axes parameter. But it's true this is unnecessary for now, I can remove it.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
jdeschamps: Yeah I think what I was trying to remember was basically: it works for comparisons, but you should not use it to set values because it is not really a string. 

I agree on the typing, and that's exactly what we had for a long time. It led to a lot of issues in the yaml import/export, and we decided to just kick it out. I can't remember the specifics now, I need to try it out with the code.

There is also (or used to be) an `enum_as_values` (approximate name) keyword in the Pydantic configuration to use values of the enum rather than the enum themselves upon model dump.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
jdeschamps: Then let's remember that, dropping `py39` could be a quick PR next week.

jdeschamps:
We should merge it!

veegalinova:
Sorry that it took me a long time to look through 😅 
Looks good, I don't have any other comments, let's merge it!

veegalinova (approved):

PR #381, opened 2025-01-29
update to expose ModelCheckpoint parameters in convenience configuration functions

### LINKED ISSUE #353: Expose `ModelCheckpoint` parameters in convenience configuration functions
## Problem

If users want to set the parameters of the checkpoointing, they need to modify the values directly into the Pydantic configuration:

```python
from careamics.config import create_care_configuration

config = create_care_configuration(...)
config.training_config.checkpoint_callback.every_n_epochs = 5
```

We could instead allow a `checkpoint_kwargs`  or better yet the `ModelCheckpoint` Pydantic model in the API for the convenience functions.

diyasrivas:
### Description

- **What**: 
Added ModelCheckpoint import to training_model.py
Added arbitrary_types_allowed=True to the model configuration
Exposed ModelCheckpoint parameter in create_care_configuration() function
Propagated ModelCheckpoint parameter through configuration factory functions
- **Why**: 
Users need to modify checkpoint configuration after creating the configuration object by directly accessing training_config.checkpoint_callback. These changes allow users to set ModelCheckpoint parameters directly during configuration creation, making the API more user-friendly and explicit. 
- **How**: 
The ModelCheckpoint import was added to training_model.py and arbitrary_types_allowed flag was enabled to properly support the ModelCheckpoint type. ModelCheckpoint parameter was added to create_care_configuration() and propagated through the configuration factory functions.

### Changes Made

- **Added**: 
ModelCheckpoint import from pytorch_lightning.callbacks
model_checkpoint parameter to configuration creation functions
arbitrary_types_allowed=True to model configuration
- **Modified**: 
Updated documentation for configuration functions
Modified configuration factory chain to support ModelCheckpoint

### Related Issues
- Closes #353 

### Example config:
```python
config = create_care_configuration(
    experiment_name="experiment_name",
    data_type="tiff",  
    axes="ZYX",
    patch_size=[16, 64, 64],
    batch_size=1,
    num_epochs=1,
    model_checkpoint=ModelCheckpoint(  
        monitor="val_loss",
        save_top_k=10,
        save_last=True,
        mode="min",
        every_n_epochs=5,
        auto_insert_metric_name=False
    )
)
``` 
---

- [ ] Pre-commit passes

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -52,6 +52,8 @@ class CAREamist:
         by default None.
     callbacks : list of Callback, optional
         List of callbacks to use during training and prediction, by default None.
+        Note: ModelCheckpoint configuration should be set through the Configuration 
+        object (config.training_config.model_checkpoint) rather than through callbacks.
 
     Attributes
     ----------
@@ -219,6 +221,11 @@ class CAREamist:
     def _define_callbacks(self, callbacks: Optional[list[Callback]] = None) -> None:
         """Define the callbacks for the training loop.
 
+        ModelCheckpoint configuration should be provided through the Configuration 
+        object (config.training_config.model_checkpoint) rather than through callbacks.
+        If no ModelCheckpoint is specified in the configuration, default checkpoint 
+        settings will be used.
+
         Parameters
         ----------
         callbacks : list of Callback, optional
@@ -247,7 +254,9 @@ class CAREamist:
         self.callbacks.extend(
             [
                 HyperParametersCallback(self.cfg),
-                ModelCheckpoint(
+                self.cfg.training_config.model_checkpoint
+                if self.cfg.training_config.model_checkpoint is not None
+                else ModelCheckpoint(
                     dirpath=self.work_dir / Path("checkpoints"),
                     filename=self.cfg.experiment_name,
                     **self.cfg.training_config.checkpoint_callback.model_dump(),
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -221,11 +221,14 @@ class CAREamist:
     def _define_callbacks(self, callbacks: Optional[list[Callback]] = None) -> None:
         """Define the callbacks for the training loop.
 
+<<<<<<< HEAD
         ModelCheckpoint configuration should be provided through the Configuration
         object (config.training_config.model_checkpoint) rather than through callbacks.
         If no ModelCheckpoint is specified in the configuration, default checkpoint
         settings will be used.
 
+=======
+>>>>>>> 3802fcd (Fix careamist.py, configuration_factories.py, and training_model.py per review feedback (PR #381))
         Parameters
         ----------
         callbacks : list of Callback, optional
@@ -254,6 +257,7 @@ class CAREamist:
         self.callbacks.extend(
             [
                 HyperParametersCallback(self.cfg),
+<<<<<<< HEAD
                 (
                     self.cfg.training_config.model_checkpoint
                     if self.cfg.training_config.model_checkpoint is not None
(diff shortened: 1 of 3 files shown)


src/careamics/config/training_model.py
```diff
@@ -55,6 +57,9 @@ class TrainingConfig(BaseModel):
     """Checkpoint callback configuration, following PyTorch Lightning Checkpoint
     callback."""
 
+    model_checkpoint: Optional[ModelCheckpoint] = None
+    """Optional override for the model checkpoint configuration."""
```
melisande-c: Don't need this because there is the `checkpoint_callback` parameter above!

src/careamics/config/training_model.py
```diff
@@ -28,6 +29,7 @@ class TrainingConfig(BaseModel):
     # Pydantic class configuration
     model_config = ConfigDict(
         validate_assignment=True,
+        arbitrary_types_allowed=True,  # test - Diya 27.1.25
```
melisande-c, line 33: This means the configuration won't be serialised correctly when exporting to yaml.

src/careamics/config/configuration_factories.py
```diff
@@ -225,6 +226,7 @@ def _create_configuration(
     use_n2v2: bool = False,
     model_params: Optional[dict] = None,
     dataloader_params: Optional[dict] = None,
+    model_checkpoint: Optional[ModelCheckpoint] = None,
```
melisande-c: Want to pass either `model_checkpoint_kwargs` (a dictionary) or the `CheckpointModel` object here, which will be passed to the `TrainingConfig`, e.g.

```python
TrainingConfig(
    num_epochs=1,
    batch_size=8,
    checkpoint_callback={
        # all model_chekpoint params can be passed as a dict in the following way
        # An initialised CheckpointModel object will also be accepted
        "every_n_train_steps": 5
    },
)
```

src/careamics/careamist.py
```diff
@@ -247,10 +254,14 @@ def _define_callbacks(self, callbacks: Optional[list[Callback]] = None) -> None:
...
-                    filename=self.cfg.experiment_name,
-                    **self.cfg.training_config.checkpoint_callback.model_dump(),
+                (
+                    self.cfg.training_config.model_checkpoint
+                    if self.cfg.training_config.model_checkpoint is not None
+                    else ModelCheckpoint(
+                        dirpath=self.work_dir / Path("checkpoints"),
+                        filename=self.cfg.experiment_name,
+                        **self.cfg.training_config.checkpoint_callback.model_dump(),
+                    )
```
melisande-c: This should return to how it was originally, once the other changes have been made to allow for the `training_config.checkpoint_callback` to be set in the configuration factory convenience functions.

melisande-c (changes requested):
The training config already has a `CheckpointModel` object! And allowing arbitrary types in the `Configuration` class will cause a problem because they will not serialise correctly when the configuration is exported to a yaml file.

`CheckpointModel` is a pydantic model that is defined within CAREamics, it contains all the parameters necessary to initialise a `ModelCheckpoint` object which is the actual lightning class used for saving checkpoints. (The names are confusing though). To initialise a `ModelCheckpoint` with a `CheckpointModel` you do:

```python
checkpoint_params = CheckpointModel(
    every_n_train_steps=5,
    # any of the other params you choose
)
checkpoint_callback = ModelCheckpoint(**checkpoint_params.model_dump())
```

The configuration factories should allow for the user to pass model checkpoint params, as a dictionary (or potentially the `CheckpointModel`), which can then be passed to the `TrainingConfig` to override the default `CheckpointModel`. Then the callback should be created in the `CAREamist` class as it did before with no modifications.

src/careamics/config/configuration_factories.py
melisande-c: or in the context of the `_create_configuration` function, the following will work:
```python
def _create_configuration(
    # other args ...
    model_checkpoint: Union[dict[str, Any], CheckpointModel],
) -> Configuration:
    
    # function logic
    # ...

    # training model
    training = TrainingConfig(
        num_epochs=num_epochs,
        batch_size=batch_size,
        logger=None if logger == "none" else logger,
        checkpoint_callback=model_checkpoint,
    )

    # ... rest of function
```

jdeschamps:
Superseded by https://github.com/CAREamics/careamics/pull/483

PR #387, opened 2025-02-03
Feature: Add ability to set dataloader kwargs for validation dataset

### LINKED ISSUE #342: Validation dataloder parameters
## Problem

Currently, the validation dataloader does not accept the usual PyTorch `kwargs`, but this could easily be added to the configuration in the `DataConfig`.

Also potentially relevant for https://github.com/CAREamics/careamics/issues/292.

## Potential solution

```python
class DataConfig(BaseModel):
    ...

    train_dataloader_params: Optional[dict] = None

    val_dataloader_params: Optional[dict] = None
```

melisande-c:
## Description

> [!NOTE]  
> **tldr**:  This PR gives users the ability to set the dataloader parameters for the validation dataset. Also included is a minor refactor for the train dataloader parameters to be validated in the `GeneralDataConfig` rather than be modified in the `TrainDataModule.train_dataloader` method.

Previously there was no way in which users could modify the dataloader parameters for the validation dataset; not being able to set something like the number of workers could unnecessarily slow down users code. 

Validation that `train_dataloader_params` has a `shuffle` parameter was added to the `GeneralDataConfig` because it is better to notify a user when their configuration is wrong rather than silently change behaviour. One small unfortunate thing however, is that the `IterableDataset` cannot be shuffled; so in this case the `shuffle` argument is not given to the dataloader. There are plans to remove the iterable dataset, so this won't be necessary in the future. 

### Overview - what changed?

- `GeneralDataConfig` now has `train_dataloader_params` and `val_dataloader_params` rather than just `dataloader_params`. 
- All configuration creation convenience functions now accept `train_dataloader_params` and `val_dataloader_params` rather than just `dataloader_params`. 

Added new fields to `GeneralDataConfig` and updated configuration creation convenience functions accordingly.

- `GeneralDataConfig`
  - added `val_dataloader_params` field.
  - added `train_dataloader_params` field.
  - removed `dataloader_params` field.
  - added field validator to ensure `train_dataloader_params` has shuffle key.
- all configuration creation convenience functions
  - added `val_dataloader_params` argument
  - added `train_dataloader_params` argument
  - removed `dataloader_params` argument.
- `TrainDataModule`
  - `train_dataloader` method
  - `val_dataloader` method

### Removed features or files

- `TrainDataModule.dataloader_params` attribute 
  - just access from data_config no need to copy the attribute.

## How has this been tested?

- Modified existing `test_create_configuration` to also check `val_dataloader_params` are set as expected.
- Added an additional test to ensure than an error is raised if `train_dataloader_params` does not include a "shuffle" key.

## Related Issues

- Resolves #342 

## Breaking changes

Old configuration that include `dataloader_params`.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/configuration_factories.py
+++ b/src/careamics/config/configuration_factories.py
@@ -224,7 +224,8 @@ def _create_configuration(
     logger: Literal["wandb", "tensorboard", "none"],
     use_n2v2: bool = False,
     model_params: Optional[dict] = None,
-    dataloader_params: Optional[dict] = None,
+    train_dataloader_params: Optional[dict[str, Any]] = None,
+    val_dataloader_params: Optional[dict[str, Any]] = None,
 ) -> Configuration:
     """
     Create a configuration for training N2V, CARE or Noise2Noise.
@@ -262,8 +263,10 @@ def _create_configuration(
         Whether to use N2V2, by default False.
     model_params : dict
         UNetModel parameters.
-    dataloader_params : dict
-        Parameters for the dataloader, see PyTorch notes, by default None.
+    train_dataloader_params : dict
+        Parameters for the training dataloader, see PyTorch notes, by default None.
+    val_dataloader_params : dict
+        Parameters for the validation dataloader, see PyTorch notes, by default None.
 
     Returns
     -------
@@ -294,8 +297,12 @@ def _create_configuration(
         "patch_size": patch_size,
         "batch_size": batch_size,
         "transforms": augmentations,
-        "dataloader_params": dataloader_params,
     }
+    # Don't override defaults set in DataConfig class
+    if train_dataloader_params is not None:
+        data["train_dataloader_params"] = train_dataloader_params
+    if val_dataloader_params is not None:
+        data["val_dataloader_params"] = val_dataloader_params
 
     # training model
     training = TrainingConfig(
@@ -331,7 +338,8 @@ def _create_supervised_configuration(
     n_channels_out: Optional[int] = None,
     logger: Literal["wandb", "tensorboard", "none"] = "none",
     model_params: Optional[dict] = None,
-    dataloader_params: Optional[dict] = None,
+    train_dataloader_params: Optional[dict[str, Any]] = None,
+    val_dataloader_params: Optional[dict[str, Any]] = None,
 ) -> Configuration:
     """
     Create a configuration for training CARE or Noise2Noise.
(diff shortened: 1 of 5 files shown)


src/careamics/config/data/data_model.py
```diff
@@ -100,8 +101,13 @@ class GeneralDataConfig(BaseModel):
     """List of transformations to apply to the data, available transforms are defined
     in SupportedTransform."""
 
-    dataloader_params: Optional[dict] = None
-    """Dictionary of PyTorch dataloader parameters."""
+    train_dataloader_params: dict[str, Any] = Field(
```
jdeschamps, line 104: This might clash with the modifications in https://github.com/CAREamics/careamics/pull/365 to the same file, but I think it may be easier to merge this one first.

jdeschamps (approved):
Nice!

src/careamics/config/configuration_factories.py
```diff
@@ -648,8 +674,10 @@ def create_n2n_configuration(
         Logger to use, by default "none".
     model_params : dict, optional
         UNetModel parameters, by default {}.
-    dataloader_params : dict, optional
-        Parameters for the dataloader, see PyTorch notes, by default None.
+    train_dataloader_params : dict
+        Parameters for the training dataloader, see PyTorch notes, by default None.
+    val_dataloader_params : dict
+        Parameters for the validation dataloader, see PyTorch notes, by default None.
```
melisande-c: I just realised this should be `dict, optional` will update tomorrow.

src/careamics/config/configuration_factories.py
```diff
@@ -618,6 +639,11 @@ def create_n2n_configuration(
     transforms, a list of transforms can be passed to the `augmentations` parameter. To
     disable the transforms, simply pass an empty list.
 
+    By setting `train_dataloader_params` and `val_dataloader_params` to `None` the
+    defaults set in the `DataConfig` will be used. This is the dictionary
+    {"shuffle": True} for the `train_dataloader_params` and the empty dict {} for
+    the `val_dataloader_params`.
+
```
melisande-c: I've decided I will just include this information in the `Parameters` section, as it will probably be easier to see. Will update tomorrow.

src/careamics/config/data/data_model.py
melisande-c: @CatEek are you ok to merge?

PR #395, opened 2025-02-10
fix: Add CAREamics version to BMZ

### LINKED ISSUE #385: BMZ-export environment file installs CAREamics from github not pypi
The environment file that is created during the BMZ export installs CAREamics from github and not pypi. The current environment file will install CAREamics from the main branch, which may not be compatible with the exported config file and/or model anymore. The environment file should install the version of CAREamics the model was created with.

See environment file text that is created below.
https://github.com/CAREamics/careamics/blob/426ba1d904ee3746a58b02deecfa1ec68ec1a527/src/careamics/model_io/bioimage/bioimage_utils.py#L43-L52

### Suggested change

```
    f"- careamics={__version__}"
```

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Make BMZ use the latest PyPI version of CAREamics.

Following https://github.com/CAREamics/careamics/issues/385, this PR adds utilities to extract the version of CAREamics and modifies the BMZ environment yaml to fetch this version from PyPI.

The BMZ env yaml was fetching and installing CAREamics via the main branch in Github, this can cause errors as the main branch might have breaking changes compared to the latest PyPI or even older versions users might have installed.

### Overview - what changed?

This PR introduces a new utils function to get the `Major.minor.patch` version of CAREamics, and builds the BMZ env yaml such that `torch`, `torchvision`, and `CAREamics` are fetched from PyPI according to their know versions.

Locally (e.g. when installed via `pip install -e .`) the CAREamics `__version__` is returned as a dev versions: e.g. `0.0.7.dev10a43fe0b087f`. The new `get_clean_version` utility only keeps the `Major.minor.patch` version.

Finally, the function building up the BMZ env yaml now calls this utility and fetches `torch`, `torchvision` and `CAREamics` from PyPI.

Note that the choice of moving `torch` and `torchvision` to PyPI is motivated by [PyTorch's conda package discontinuation](https://github.com/pytorch/pytorch/issues/138506).

### New features or files

- `careamics.utils.version.get_clean_version` to get the CAREamics version, and corresponding test.

- `careamics.model_io.bioimage.bioimage_utils.create_env_text` now includes CAREamics version and installs `torch` and `torchvision` from pip.

## How has this been tested?

Only via `pytest`. 

> [!WARNING]  
> This should be tested with an install from PyPI (or with a tagged commit) and with upload to the BMZ to verify that the env is valid.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/385

> [!WARNING]  
> If working on a local editable install, `pip` automatically increases the patch version from the latest **known**. If you `pip install -e .` again, this might increase the latest known version. BMZ models should therefore preferentially be uploaded from an env installed from PyPI.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/model_io/bioimage/bioimage_utils.py
+++ b/src/careamics/model_io/bioimage/bioimage_utils.py
@@ -3,6 +3,8 @@
 from pathlib import Path
 from typing import Union
 
+from careamics.utils.version import get_clean_version
+
 
 def get_unzip_path(zip_path: Union[Path, str]) -> Path:
     """Generate unzipped folder path from the bioimage.io model path.
@@ -44,11 +46,11 @@ def create_env_text(pytorch_version: str, torchvision_version: str) -> str:
         f"name: careamics\n"
         f"dependencies:\n"
         f"  - python=3.10\n"
-        f"  - pytorch={pytorch_version}\n"
-        f"  - torchvision={torchvision_version}\n"
         f"  - pip\n"
         f"  - pip:\n"
-        f"    - git+https://github.com/CAREamics/careamics.git\n"
+        f"    - torch=={pytorch_version}\n"
+        f"    - torchvision=={torchvision_version}\n"
+        f"    - careamics=={get_clean_version()}"
     )
 
     return env
--- a/src/careamics/model_io/bioimage/bioimage_utils.py
+++ b/src/careamics/model_io/bioimage/bioimage_utils.py
@@ -3,7 +3,7 @@
 from pathlib import Path
 from typing import Union
 
-from careamics.utils.version import get_clean_version
+from careamics.utils.version import get_careamics_version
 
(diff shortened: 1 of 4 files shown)


melisande-c (approved):
Looks good,

Only addition I could think of is maybe a warning to let the user know if they are not on a clean install of CAREamics, in the unlikely case they have changed something that would make it not compatible with the true version. But I think it's unlikely so not that important.

Also, in the exported README the CAREamics version is mentioned, we might want to also use the `get_clean_version` function here. https://github.com/CAREamics/careamics/blob/2c25e2a86e8aa0a0714d49ff0402d008ea14ba1c/src/careamics/model_io/bmz_io.py#L142

jdeschamps:
Great ideas! Implemented!

PR #413, opened 2025-02-18
Enforce `shuffling` in convenience functions

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Prevent users from running into errors as soon as they add training dataloader parameters.

I ran into this error while fixing the docs: if one wishes to change `num_workers` in the train dataloader, then an error would be raised (Pydantic validation) if the `shuffle` parameter is also absent.

This PR enforces the presence of the `shuffle` parameter for users.

The current mindset of the configuration is the following:
- Pydantic configuration do not change the user input and the sole responsibility is coherence and clarity of errors (at least those we write).
- Convenience functions are the entry point for users and these should minimize the potential for misconfigurations (and therefore validation errors).

The current behaviour leads to easy to stumble upon errors: any parameter but `shuffle` passed to the `train_dataloader_params` parameter of the convenience functions will lead to an error thrown at the user.

### Overview - what changed?

This PR enforces the addition of `shuffle` to the `train_dataloader_params` if it was not passed by users to the convenience functions.

The convenience functions call `_create_data_configuration`, which now test for the presence of `shuffle` and adds it if absent.

- `_create_data_configuration` checks for presence in `train_dataloader_params` and add `"shuffle" : True` if absent
- Updated the description of the parameter to mention the necessity to have the `shuffle` parameter.
- Added a test to check the behaviour.

## How has this been tested?

Locally and via the tests.

## Breaking changes

No breaking changes I can think of.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/configuration_factories.py
+++ b/src/careamics/config/configuration_factories.py
@@ -242,11 +242,11 @@ def _create_data_configuration(
     }
     # Don't override defaults set in DataConfig class
     if train_dataloader_params is not None:
-        data["train_dataloader_params"] = train_dataloader_params
-
         # DataConfig enforces the presence of `shuffle` key in the dataloader parameters
         if "shuffle" not in train_dataloader_params:
             train_dataloader_params["shuffle"] = True
+
+        data["train_dataloader_params"] = train_dataloader_params
 
     if val_dataloader_params is not None:
         data["val_dataloader_params"] = val_dataloader_params
--- a/src/careamics/config/configuration_factories.py
+++ b/src/careamics/config/configuration_factories.py
@@ -243,6 +243,11 @@ def _create_data_configuration(
     # Don't override defaults set in DataConfig class
     if train_dataloader_params is not None:
         data["train_dataloader_params"] = train_dataloader_params
+
+        # DataConfig enforces the presence of `shuffle` key in the dataloader parameters
+        if "shuffle" not in train_dataloader_params:
+            train_dataloader_params["shuffle"] = True
+
     if val_dataloader_params is not None:
         data["val_dataloader_params"] = val_dataloader_params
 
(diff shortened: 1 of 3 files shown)


melisande-c (approved):
Yeah makes sense :)

I think I didn't do this when I added the `train_dataloader_params` argument because I didn't want to set `shuffle=True` in two places (pydantic model defaults + convenience functions). But from a user experience perspective it does make sense to try and prevent this kind of error, and this is the simplest way to solve the problem.

Not necessarily the case here, but I'm gonna just bring up that one small annoyance I have with the convenience functions is sometimes the mismatch of defaults with the pydantic models. Or if we want to change a default we have to change it in two places, which can be easy to forget and can cause confusion. But not fully sure on a solution

jdeschamps:
> Yeah makes sense :)
> 
> I think I didn't do this when I added the `train_dataloader_params` argument because I didn't want to set `shuffle=True` in two places (pydantic model defaults + convenience functions). But from a user experience perspective it does make sense to try and prevent this kind of error, and this is the simplest way to solve the problem.
> 
> Not necessarily the case here, but I'm gonna just bring up that one small annoyance I have with the convenience functions is sometimes the mismatch of defaults with the pydantic models. Or if we want to change a default we have to change it in two places, which can be easy to forget and can cause confusion. But not fully sure on a solution

Yes that is indeed annoying, although I imagine there are not so many instances (at least right now I can't think of many).

Not super elegant, but we can instantiate the particular Pydantic model and get the default value. Or use the Pydantic model class itself and get the defaults through its schema. It will be quite verbose though (the latter especially).

melisande-c:
> Not super elegant, but we can instantiate the particular Pydantic model and get the default value. Or use the Pydantic model class itself and get the defaults through its schema. It will be quite verbose though (the latter especially).

Yeah or we have a pydantic model `DataloaderParams` which has `shuffle=True` by default and then do `DataloaderParams(**train_dataloader_params)` in the convenience function.

But to fully remove all defaults from convenience functions also might not be very intuitive for users since all arguments will have to be set to `None`.

So yeah I think we just leave this as is for now.

CatEek (approved):

PR #417, opened 2025-02-19
Feature(Next-Gen Datasets): Zarr patch reading

### LINKED ISSUE #397: Next-Gen Dataset: Zarr patch reading
### Overview

One of the features of CAREamics's Next-Gen Datasets will be to be able to read patches from Zarr datasets. The core dataset should be agnostic to whether the data is stored on in memory or on disk (in a zarr file).

### Background

#373 introduced the `ImageStack` class, an interface for storing and reading image data and a `PatchExtractor` class for extracting patches from multiple images.

### TODO:
- [ ] Implement `ZarrImageStack`
    - A placeholder for this class was added in #373. It needs to have its `extract_patch` method implemented, which will also need to handle axis manipulation. As the extracted patch is expected to have C(Z)YX dimension order.
- [ ] Test how the newly implemented `ZarrImageStack` will work with the `PatchExtractor`.
    - References to the Zarr file will need to be kept somewhere, likely in the `PatchExtractor`. Either the original `PatchExtractor` will be modified to handle both the `InMemoryImageStack` and the `ZarrImageStack`. Or a separate `ZarrPatchExtractor` can be created.

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Implementing the backend for reading patches from zarr files by implementing a `ZarrImageStack` that was first introduced as a not-implemented class in #373.

In the Next-Gen PyTorch Dataset class we want to be able to seamlessly read data from multiple sources. `ImageStack`
defines an interface which allows for different concrete implementations to be swapped without higher level classes needing any additional information or functionality.

### Overview - what changed?

Added the concrete implementation of the `ZarrImageStack` and a class constructor method `from_ome_zarr`, which can be used if to instantiate the class if it follows the ome standard.

This potentially leaves room for advanced users to write a custom loading function for arbitrary zarr file layouts.

The `ZarrImageStack` is instantiated with a zarr `store`, the internal zarr path file `data_path`, and the original data `axes` (a subset of `STCZYX`), the zarr store is currently limited to  currently limited to the [`FSStore`](https://zarr.readthedocs.io/en/v2.18.4/api/storage.html#zarr.storage.FSStore), but we can look into expanding this. Also note the zarr store classes have changed quite a lot between v2 and v3.

The `extract_patch` method has to transform the `sample_idx`, `coords` and `patch_size` arguments to index the zarr array in the original axes order, and then reshape the resulting patch to the expected `C(Z)YX` axes. Indexing the `zarr.Array` object will load the requested data and return a numpy array. 

### New features or files

- Implemented `ZarrImageStack` class

- `PatchExtractor` class: add `from_ome_zarr` class constructor
- `PatchExtractor` class: fixed some variable names
- `ImageStack` protocol: updated `source` and `data_shape` to properties for greater flexibility

## How has this been tested?

EDIT: One test for the extract patch method has been added, it tests extracting a patch for multiple underlying axes orders.

Still no tests (or good docs), I want to take some time to properly plan them. But use the example below to see the class in action working on a zarr stored in S3.

```python
from matplotlib import patches
from careamics.dataset_ng.patch_extractor.image_stack import ZarrImageStack

# found this link online
path = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr"
image_stack = ZarrImageStack.from_ome_zarr(path=path)
print("Image stack axes SCZYX =", image_stack.data_shape)

# slightly abusing the `extract_patch` method to extract z-slice for visualisation
z_idx = 100
z_slice = image_stack.extract_patch(
    sample_idx=0, coords=(z_idx, 0, 0), patch_size=(1, *image_stack.data_shape[-2:])
)
print("Z-Slice axes CZYX =", z_slice.shape)

patch_2d_coords = (50, 200)
patch_2d_size = (64, 64)
z_slice_patch = image_stack.extract_patch(
    sample_idx=0, coords=(z_idx, *patch_2d_coords), patch_size=(1, *patch_2d_size)
)

fig, axes = plt.subplots(1, 2)
axes[0].imshow(z_slice[0, 0])
patch_rect = patches.Rectangle(
    (patch_2d_coords[1], patch_2d_coords[0]),
    width=patch_2d_size[1],
    height=patch_2d_size[0],
    fc=[0, 0, 0, 0],
    ec="r",
)
axes[0].add_artist(patch_rect)
axes[1].imshow(z_slice_patch[0, 0])

plt.show()
```
```
Image stack axes SCZYX = (1, 2, 236, 275, 271)
Z-Slice axes CZYX = (2, 1, 275, 271)
```
zarr_image_stack_example

## Related Issues

- Resolves #397

Apologies some stuff is maybe a little ugly, if anything is unclear let me know and I will add more comments, but I thought it was important to get a working version finished. 

Going forward, I think we should consider how to make a class that will be easy for advanced users to implement / inherit from, for example I think most NGFF formats will need to apply axes transformations in `extract_patch` so maybe this can move to a base class that will automatically apply this without the user having to worry about it.

---

- [x] Pre-commit passes


src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
```diff
@@ -1,24 +1,152 @@
...
+            return sample_idx
+
+    def _get_Z_slice(self, coords: Sequence[int], patch_size: Sequence[int]) -> slice:
+        """Get z slice given `coords` and `patch_size`"""
+        if "Z" not in self._original_axes:
+            raise ValueError("No 'Z' axis specified in original data axes.")
+        idx = 0
+        return slice(coords[idx], coords[idx] + patch_size[idx])
+
+    def _get_Y_slice(self, coords: Sequence[int], patch_size: Sequence[int]) -> slice:
```
CatEek: Unless I'm missing something, why do we need 3 separate methods for spatial coords ?

jdeschamps: I also find this rather convoluted. Couldn't we simply do:

```python
return [
     slice(coords[i], coords[i] + patch_size[i]) for i in range(len(coords))
]
```

or

```python
return [
    slice(coord, coord + patch_s) for coord, patch_s in zip(coords, patch_size)   
]
```

jdeschamps: Ok, I get it: the code currently does not assume that the spatial axes are following each other (e.g. "XTZCY") or always in the same order.

This has the advantage of allowing users to specify "ZYX", "YXZ", "XYZ" etc.

jdeschamps (approved):

jdeschamps:
> ## Additional Notes and Examples
> 
> Apologies some stuff is maybe a little ugly, if anything is unclear let me know and I will add more comments, but I thought it was important to get a working version finished.

We might discuss to see if there is a more elegant way to structure the same logic. I'd say maybe a bit more comments could be useful because it is a bit hard to grasp right away!

 
> Going forward, I think we should consider how to make a class that will be easy for advanced users to implement / inherit from, for example I think most NGFF formats will need to apply axes transformations in `extract_patch` so maybe this can move to a base class that will automatically apply this without the user having to worry about it.

I very much would like that!

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
melisande-c: @jdeschamps yep exactly, correct me if I'm wrong but there is no guarantee on the order of the original data axes. I only wrote the `_get_X_axes` etc. functions to condense the `if-else` block in `extract_patch` and are not really intended for re-use, I can remove the functions and put the logic back into the `extract_patch` function if you think it is clearer. I will also add some more comments explaining `extract_patch`.

But yeah, the function extract patch has to transform the indices to the actual shape of the data and then transform the patch to return the expected C(Z)YX shape. The first way I thought to do this is to just loop through the original axes and work out what the index or slice should be from the given arguments. There is probably a better solution, or a better way to organise the logic, so please if you have any suggestions let me know @jdeschamps @CatEek

jdeschamps: In the vast majority of cases you will have similar orders, but X-Y can be switched, and C can be at the end or before the XY. Same for Z, but probably very rare.

So being flexible is a good choice!

melisande-c: Ok so I removed all the "get spatial index" functions and put the logic back into the `extract_patch` method so hopefully now it is clearer what is going on. Also the "get spatial index" functions were pretty pointless, I did kinda rush to finish this last week, so definitely could use some refining on the readability side, thanks for the feedback!

jdeschamps:
Happy to merge

melisande-c:
I was thinking I might just add some very basic tests for loading data with different dimension to make sure everything is ok.

EDIT: Yeah I think I can see one last problem with the data reshaping, so I will add some tests and fix that.

melisande-c:
Update: Added 1 test for extracting patches with a range of different underlying axes orders. All the cases worked so no further modification of the code was necessary.

jdeschamps:
I almost wrote to our channel yesterday, but we should start writing tests indeed. Otherwise this code base will grow rapidly unchecked.

melisande-c:
> I almost wrote to our channel yesterday, but we should start writing tests indeed. Otherwise this code base will grow rapidly unchecked.

Yeah the original justification for not having tests yet was that the structure of the code might change rapidly so then it would be annoying to write tests that would very quickly not be applicable. But now we are reaching a point where it seems a few interfaces are a bit more solid so it does make sense to have some tests.

Maybe we should have a meeting to figure out the minimum testing needed for development, and to maybe plan a structure for the tests, e.g. unit tests separated from main interface tests in some way. Just to avoid the tests ballooning like they already have in the main repo.

I will open an issue about this.

veegalinova (approved):

PR #420, opened 2025-02-25
Feature(Next-Gen Datasets): Dataset class prototype

### LINKED ISSUE #398: Next-Gen Dataset: Core dataset class skeleton
Add an outline of the core dataset class.

This will give us an idea of how the different components inside the dataset will work together, and what makes most sense as an interface for each of these components.

### LINKED ISSUE #401: Next-Gen Dataset: Augmentations
Add augmentations to the next-gen dataset class.

We will probably keep this very similar to how it is currently implemented.

Relevant issues: #167.

veegalinova:
## Description

> [!NOTE]  
> **tldr**: 
This is a prototype implementation of a dataset that aims to combine the features of the existing datasets in one flexible class. 

Current code base has multiple dataset implementations with overlapping functionality, making it difficult to use and maintain. This PR aims to create a more streamlined class that should be more easily extensible and maintainable.

### Overview - what changed?

This PR introduces the CareamicsDataset class, a map-style PyTorch dataset that supports both training and inference workflows. It has configurable data reading, patch extraction, and transformation steps. The implementation aims to work with file-based, in-memory-based, and custom data sources. 

Implemented `CareamicsDataset` class, made a few small changes to the `ImageStack`, `PatchExtractor` and `PatchSpecsGenerator`

### New features or files

- add CareamicsDataset class in `dataset.py` 
  - Takes a config (either DataConfig or InferenceConfig), array or path inputs and optional array or path targets, optional reading function for custom data types. 
  - Outputs an `ImageRegionData` named tuples, containing created data patch, image metadata and patch metadata

Opinions are really welcome for any of these changes! 

- added attribute `data_dtype` to the ImageStack protocol and it's implementations to track data dtype
- added `shape` property to the PatchExtractor class that returns the shapes of the internal data
- changed `create_patch_extractors` method to only return one extractor at a time 
- changed `PatchSpecsGenerator` protocol - moved `data_shapes` argument to `generate` function, moved the rest of the arguments from `generate` to the `init`. Motivation - I think `patch_size` belongs to the class instance, and the data_shapes is a `generate` input. 
- added `SequentialPatchSpecsGenerator` - it's an incomplete copy from the current sequential patching implementation, not tested yet! 

## How has this been tested?

I created a demo notebook  [demo_dataset.ipynb](src/careamics/dataset_ng/demo_dataset.ipynb) that shows some usage scenarious for basic functionality.
I plan to add unit tests after review feedback and the following possible changes to the overall structure. 

## Related Issues

- Resolves #398 

--- 

- [ ] Pre-commit passes


src/careamics/dataset_ng/patch_extractor/patch_extractor.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, data_readers: Sequence[ImageStack]):
         self.image_stacks: list[ImageStack] = list(data_readers)
 
     @classmethod
-    def from_arrays(cls, source: Sequence[NDArray], *, axes: str) -> Self:
+    def from_arrays(cls, source: Sequence[NDArray], *, axes: str, **kwargs) -> Self:
```
melisande-c: Why did you add `**kwargs` to the function signature?

Also on `from_tiff_files` below.

src/careamics/dataset_ng/patch_extractor/patch_extractor.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, data_readers: Sequence[ImageStack]):
         self.image_stacks: list[ImageStack] = list(data_readers)
```
melisande-c: Also whoops, `data_readers` argument here should have been updated to `image_stacks`. I guess I missed this instance when I was renaming everything.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
```diff
@@ -25,45 +26,18 @@ def get_patch_extractor_constructor(
...
 
-def create_patch_extractors(
-    data_config: DataConfig,
-    train_data: Union[Sequence[NDArray], Sequence[Path]],
-    val_data: Optional[Union[Sequence[NDArray], Sequence[Path]]] = None,
-    train_data_target: Optional[Union[Sequence[NDArray], Sequence[Path]]] = None,
-    val_data_target: Optional[Union[Sequence[NDArray], Sequence[Path]]] = None,
+def create_patch_extractor(
+    data_config: Union[DataConfig, InferenceConfig],
+    data: Union[Sequence[NDArray], Sequence[Path]],
```
melisande-c: Makes way more sense to create one at a time! 😄

src/careamics/dataset_ng/patching_strategies/patch_specs_generator.py
```diff
@@ -12,42 +13,43 @@
...
-    def __init__(self, data_shapes: Sequence[Sequence[int]]):
-        self.data_shapes = data_shapes
+    def __init__(self, patch_size: Sequence[int], random_seed: int):
+        self.patch_size = patch_size
+        self.random_seed = random_seed
 
-    def generate(self, patch_size: Sequence[int], seed: int) -> list[PatchSpecs]:
-        rng = np.random.default_rng(seed=seed)
+    def generate(self, data_shapes: Sequence[Sequence[int]]) -> list[PatchSpecs]:
+        rng = np.random.default_rng(seed=self.random_seed)
```
melisande-c, line 34: So I wrote the `PatchSpecsGenerator` without full plan. But I guess my reasoning for having the `data_shapes` as an attribute and the `patch_size` as an argument was that: for a given instance of the dataset class the data shapes wouldn't change, but you might want to generate another set of patches with a different patch size from the same set of images? But tbh I don't really see a use case for that...

Maybe both `data_shapes` and `patch_size` should be an argument to the `generate` method?

src/careamics/dataset_ng/patch_extractor/image_stack/in_memory_image_stack.py
```diff
@@ -17,8 +17,9 @@ class InMemoryImageStack:
     def __init__(self, source: Union[Path, Literal["array"]], data: NDArray):
         self.source: Union[Path, Literal["array"]] = source
         # data expected to be in SC(Z)YX shape, reason to use from_array constructor
-        self._data = data
+        self._data: NDArray = data
         self.data_shape: Sequence[int] = self._data.shape
+        self.data_dtype: DTypeLike = self._data.dtype
```
melisande-c, line 22: Adding dtype attribute is a good idea!

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+            self.config, data=inputs, read_func=read_func
+        )
+        if targets is not None:
+            self.target_extractor: Optional[PatchExtractor] = create_patch_extractor(
+                self.config, data=targets, read_func=read_func
+            )
+        else:
+            self.target_extractor = None
+
+        self.patch_specs = self._initialize_patch_specs()
```
melisande-c, line 74: Hmm, with the random patching strategy we probably want them to be different random patches each epoch.

We either reinitialise the `patch_specs` every time the data is fully cycled through, which would probably involve some janky index incrementing. Or, maybe there is a way to make the `RandomPatchSpecs` generator iterable so it just provides a random `PatchSpec` during `__getitem__`. Although this would have to stay compatible with deterministic `PatchSpecsGenerators` like the sequential one... (I don't think this needs to be implemented now, just throwing some ideas out)

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+
+        self.input_stats, self.target_stats = self._initialize_statistics()
+
+        self.transforms = self._initialize_transforms()
```
melisande-c, line 78: Good to see existing transforms can just be integrated. We can probably close #401, unless anyone has some more thoughts @jdeschamps @CatEek?

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+    def _initialize_patch_specs(self) -> list[PatchSpecs]:
+        if isinstance(self.config, DataConfig):
+            # TODO: how to fix the random seed properly?
+            #  how to change it between epochs?
+            patch_generator: PatchSpecsGenerator = RandomPatchSpecsGenerator(
+                patch_size=self.config.patch_size,
+                random_seed=getattr(self.config, "random_seed", 42),
+            )
+        elif isinstance(self.config, InferenceConfig):
+            # TODO: how to handle the whole images?
```
melisande-c, line 87: In answer to TODO, maybe the `ImageStack` classes can have an additional method that extract a whole sample, or there could be a `WholeSamplePatchSpecGenerator` (these names are getting too long 😅) that just generates a "patch" which is just the size of the whole data. (Also don't need to implement now)

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+        if target_patch is not None and self.target_extractor is not None:
+            target_data = self._create_image_region(
+                patch=target_patch,
+                patch_spec=patch_spec,
+                extractor=self.target_extractor,
+            )
+        else:
+            target_data = None
+
+        return input_data, target_data
```
melisande-c, line 166: I was thinking maybe we could actually output a dictionary instead of tuples (I am pretty sure dicts work with collate). Something like

```python
class SelfSupervisedData(TypedDict):
    input: NDArray

class SupervisedData(TypedDict):
    input: NDArray
    target: NDArray

class CareamicsDataset(Dataset):

    # ...

    def __getitem__(self, index: int) -> Union[SelfSupervisedData, SupervisedData]:
        # ...
        if self.target_extractor is not None:
            return {
                "input": input_data,
                "target": target_data
            }
        else:
            return {
                "input": input_data
            }
```

Obviously this means the lightning modules would also have to be slightly refactored, but it hopefully prevents a lot of the imo confusing unpacking and `if target is None` checks.

melisande-c: Oh wait, I just realised this doesn't work with the `ImageRegionData`, which I do like...

Maybe there is some way to combine the concepts

melisande-c (approved):
Nice to see everything coming together! This gives us a really good starting point to see how everything works together and how to integrate new features. My comments are mostly discussion points and ideas I had while reading through, we can add them to relevant issues.

src/careamics/dataset_ng/dataset/dataset.py
jdeschamps: I agree!

If there really are concerns about the way the augmentations are implemented, then we can open an issue detailing what these concerns are.

src/careamics/dataset_ng/patch_extractor/patch_extractor.py
veegalinova: If I remember correctly there was a mypy or pyright warning that `from_arrays` and `from_tiff_files` do not match the `PatchExtractorConstructor`protocol

src/careamics/dataset_ng/patching_strategies/patch_specs_generator.py
veegalinova: Agreed! 
I was thinking the other way around - there could be a case where there are several input/target readers with different shapes, and the `patch_size` is currently a constant during the pipeline, so one PatchSpecGenerator can be reused throughout (this could be different for the micro split, though...) Also do not see a concrete use case for that

src/careamics/dataset_ng/dataset/dataset.py
veegalinova: I also really like using dicts in datasets! They would be especially useful with more complicated stuff that needs several items, like n2v masks and maybe microsplit. 
I implemented it in a similar way at first but did not see a good way to combine it with `ImageRegionData`, but we should definitely try again

src/careamics/dataset_ng/patch_extractor/patch_extractor.py
melisande-c: hmm I tested and for some reason mypy doesn't flag in the `get_patch_extractor_constructor` function, but it does flag it if I do (even with **kwargs).
```python
if __name__ == "__main__":

    constructor: PatchExtractorConstructor = PatchExtractor.from_arrays`
```
So yeah I don't think I wrote that protocol correctly... The whole constructor thing can def be improved, keeping in mind user customisation (which I will open an issue for shortly).

melisande-c:
Resolves #401

src/careamics/dataset_ng/patch_extractor/patch_extractor.py
melisande-c: But yeah this is my mistake, I will try to find a fix soon and make a PR.

melisande-c: This should be resolved since #434

src/careamics/dataset_ng/patch_extractor/patch_extractor.py
melisande-c: I think I updated this in #417

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: Opened issue #425

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: Opened issue #425

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: Let's leave this for now so we don't have to refactor the lightning modules too much, but keep as a TODO for future, I will reference this comment in the main Next-Gen dataset issue

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+
+        self.patch_specs = self._initialize_patch_specs()
+
+        self.input_stats, self.target_stats = self._initialize_statistics()
+
+        self.transforms = self._initialize_transforms()
+
+    def _initialize_patch_specs(self) -> list[PatchSpecs]:
+        if isinstance(self.config, DataConfig):
+            # TODO: how to fix the random seed properly?
```
jdeschamps: Do we have issues open already or notes somewhere that reproducibility should be possible?

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+                patch_size=self.config.patch_size,
+                random_seed=getattr(self.config, "random_seed", 42),
+            )
+        elif isinstance(self.config, InferenceConfig):
+            # TODO: how to handle the whole images?
+            if self.config.tile_size is None or self.config.tile_overlap is None:
+                raise ValueError(
+                    "InferenceConfig must specify tile_size and tile_overlap"
+                )
+            patch_generator = SequentialPatchSpecsGenerator(
```
jdeschamps: This should probably be renamed to TiledPatchSpecsGenerator rather than Sequential?

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+
+        self.transforms = self._initialize_transforms()
+
+    def _initialize_patch_specs(self) -> list[PatchSpecs]:
+        if isinstance(self.config, DataConfig):
+            # TODO: how to fix the random seed properly?
+            #  how to change it between epochs?
+            patch_generator: PatchSpecsGenerator = RandomPatchSpecsGenerator(
+                patch_size=self.config.patch_size,
+                random_seed=getattr(self.config, "random_seed", 42),
```
jdeschamps, line 84: Do we have an issue opened to accumulate ideas and criticisms on the Dataconfig? e.g. adding a random seed parameter?

src/careamics/dataset_ng/dataset/dataset.py
jdeschamps: Agreed with the TODO!

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            raise ValueError(f"Data config type {type(self.config)} is not supported.")
+
+        patch_specs = patch_generator.generate(data_shapes=self.input_extractor.shape)
+        return patch_specs
+
+    def _initialize_transforms(self) -> Optional[Compose]:
+        if isinstance(self.config, DataConfig):
+            return Compose(
+                transform_list=list(self.config.transforms),
```
jdeschamps, line 104: Config currently does not include normalization, since normalization is always applied.

We could add `no_normalization` parameter to the configuration and an `if` clause here to add it or not to the list before the compose.

jdeschamps: No need to do it in this PR, but it should be a TODO and an issue (if not there already).

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+        # TODO: add TTA
+        return None
+
+    def _initialize_statistics(self) -> tuple[Stats, Stats | None]:
+        # TODO: add running stats
+        # Currently assume that stats are provided in the configuration
+        input_stats = Stats(self.config.image_means, self.config.image_stds)
+        target_means = getattr(self.config, "target_means", None)
```
jdeschamps: Not a big fan of `getattr` as the parameter names might change. I assume this is because of the Dataconfig/InferenceDataConfig mypy issue?

jdeschamps: Worst case, we could just have the `if` clause. Best case, we figure out an elegant solution on Pydantic side.

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -0,0 +1,154 @@
...
+    PatchSpecsGenerator,
+    RandomPatchSpecsGenerator,
+    SequentialPatchSpecsGenerator,
+)
+from careamics.transforms import Compose
+
+
+class ImageRegionData(NamedTuple):
+    data: NDArray
+    filename: Path | Literal["array"]
```
jdeschamps: `filename` is a bit misleading for an array

jdeschamps (approved):

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: Ah yeah because we need the crop coords as well

melisande-c: Since I'm working on improving the patching strategies right now, I can also have a think about how to do the tiling.

src/careamics/dataset_ng/dataset/dataset.py
veegalinova: Yes, see https://github.com/CAREamics/careamics/issues/436

src/careamics/dataset_ng/dataset/dataset.py
veegalinova: Added it here: https://github.com/CAREamics/careamics/issues/436

src/careamics/dataset_ng/dataset/dataset.py
veegalinova: Renamed to TiledPatchSpecsGenerator, will need to rename it in https://github.com/CAREamics/careamics/pull/439 also

src/careamics/dataset_ng/dataset/dataset.py
veegalinova: Agreed, added an if statement instead

src/careamics/dataset_ng/dataset/dataset.py
veegalinova: renamed to `source` for now, but I expect that this thing will change a lot in the near future anyways

tests/dataset_ng/dataset/test_dataset.py
```diff
@@ -0,0 +1,157 @@
...
+)
+def test_from_tiff(data_shape, patch_size, expected_dataset_len):
+    rng = np.random.default_rng(42)
+    example_input = rng.random(data_shape)
+    example_target = rng.random(data_shape)
+
+    with (
+        tempfile.NamedTemporaryFile(suffix=".tiff", delete=False) as input_file,
+        tempfile.NamedTemporaryFile(suffix=".tiff", delete=False) as target_file,
+    ):
```
melisande-c, line 62: you can use the pytest built in `tmp_path` fixture, it will create a temporary directory for the test that will be cleaned up after the test has completed :) 

https://docs.pytest.org/en/stable/how-to/tmp_path.html

PR #434, opened 2025-03-07
Refac(Next-Gen Dataset): More easily user customisable `PatchExtractor` construction

### LINKED ISSUE #424: Next-Gen Dataset: User customisability
It is important that the new dataset allows for custom user file types. Loading a whole file into memory is a pretty easy problem and has already been solved with the `InMemoryImageStack.from_custom_file_type` function.
https://github.com/CAREamics/careamics/blob/33c6e3f8b5e22e7a270861f4844e75b510c9e40b/src/careamics/dataset_ng/patch_extractor/image_stack/in_memory_image_stack.py#L48-L54

The harder case is for users writing custom `ImageStacks`. We need to make sure that there is an easy way to tell the dataset class how to construct the custom image stack, and users might want to have a custom loading function for multiple image stacks.

melisande-c:
## Description

> [!NOTE]  
> **tldr**: A refactor of the `PatchExtractor` construction that should make it easier for advanced users to define their own custom `ImageStacks` and how to load them.

In the new dataset class we want it to be possible for users to define their own `ImageStacks`; in particular, for custom NGFF formats such as HDF5, CZI and IMS, which require more custom patch reading than simply loading the entire array into memory.

### Overview - what changed?

I removed the `PatchExtractorConstructor` protocol and instead created a protocol that defines a callable that should return a sequence of `ImageStacks`, which will then in turn instantiate the `PatchExtractor`.

I made this change because I thought it would be conceptually clearer for users: if they define their own `ImageStack` class then they also have to define a function to load a series of them.

Now there are three ways that a user can customise loading their own data:
1. Define a custom `ImageStack` _and_ a custom function to load their `ImageStack`s.
2. Define a custom function function to load an existing `ImageStack` implementation, I provide an example of this method below.
3. Define a simple `ReadFunc` that will load all the data in memory as an `InMemoryImageStack`.

To do this, I removed the `PatchExtractorConstructor` protocol and I also removed all the class constructors from the `PatchExtractor`. I then created the `ImageStackLoader` protocol and converted the old `PatchExtractor` class constructors to `ImageStackLoader` implementations within a new `image_stack_loader.py` file.

An `ImageStackLoader` is a callable that must take the CAREamics `DataConfig` as the first argument and the `source` of the data as the second argument. Additional `*args` and `**kwargs` are allowed, but they should only be used to determine _how_ the data is loaded, not _what_ data is loaded. The `source` argument has to wholly determine _what_ data is loaded, this is because, downstream both an input-source and a target-source have to be specified but they will share `*args` and `**kwargs`.

I then created a new `create_patch_extractor` function with overloads for each of the different cases.

I updated the existing `create_patch_extractors` function, so it could still work with `CAREamicsDataset` class in #420 but maybe it can be removed, see TODO. Note I cherry-picked the commit from #420 that  modified the `create_patch_extractors` so it would be compatible (but I changed the arguments so there will still be conflicts).

### New features or files

- `dataset_ng/patch_extractor/image_stack_loader.py` file
- `ImageStackLoader` protocol class
- A few `ImageStackLoader` implementations
- `create_patch_extractor` function
- Demo notebook `dataset_ng/patch_extractor/demo_custom_image_stack_loader.ipynb` 

- `PatchExtractor` class (removed constructors)
- `create_patch_extractors` function

### Removed features or files

- `PatchExtractorConstructor` protocol class
- `get_patch_extractor_constructor` function

## How has this been tested?

So far tested in the demo notebook. But I also plan to add some tests for the existing image stack loader functions.

## Related Issues

- Resolves #424 

## Breaking changes

Interacts with #420 and will cause some small conflicts with calling `create_patch_extractors`.

This is the example that I tested in the provide demo notebook. It is intended for loading non-OME Zarrs that are all contained within one Zarr file. I also included this example in the docstring of `ImageStackLoader`.

```python
class ZarrSource(TypedDict):
    store: FSStore
    data_paths: Sequence[str]

def custom_image_stack_loader(
    data_config: DataConfig, source: ZarrSource, *args, **kwargs
) -> list[ZarrImageStack]:
    axes = data_config.axes
    image_stacks = [
        ZarrImageStack(store=source["store"], data_path=data_path, axes=axes)
        for data_path in source["data_paths"]
    ]
    return image_stacks
```
This can then initialise a `PatchExtractor` with the following
```python
patch_extractor = create_patch_extractor(
    data_config=data_config,
    source={"store": store, "data_paths":data_paths},
    image_stack_loader=custom_image_stack_loader,
)
```
See the notebook to test it out!

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
@@ -3,12 +3,12 @@ from pathlib import Path
 from typing import (
     Any,
     Optional,
-    ParamSpec,
     Protocol,
     Union,
 )
 
 from numpy.typing import NDArray
+from typing_extensions import ParamSpec
 
 from careamics.config import DataConfig
 from careamics.config.support import SupportedData
--- a/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
@@ -12,10 +12,15 @@ from typing_extensions import ParamSpec
 
 from careamics.config.support import SupportedData
 from careamics.file_io.read import ReadFunc
+from careamics.utils import BaseEnum
 
 from .image_stack import ImageStack, InMemoryImageStack, ZarrImageStack
 
 P = ParamSpec("P")
+
+
+class SupportedDataDev(str, BaseEnum):
+    ZARR = "zarr"
 
 
 class ImageStackLoader(Protocol[P]):
@@ -117,7 +122,7 @@ def from_ome_zarr_files(
 
 
 def get_image_stack_loader(
-    data_type: Union[SupportedData, str],  # Union with string temp for zarr
+    data_type: Union[SupportedData, SupportedDataDev],
     image_stack_loader: Optional[ImageStackLoader] = None,
 ) -> ImageStackLoader:
     if data_type == SupportedData.ARRAY:
--- a/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
@@ -62,6 +62,7 @@ class ImageStackLoader(Protocol[P]):
     ...         for data_path in source["data_paths"]
     ...     ]
     ...     return image_stacks
+
     TODO: show example use in the `CAREamicsDataset`
 
     The example above defines a `ZarrSource` dict because to determine _which_ ZARR
--- a/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
@@ -10,7 +10,6 @@ from typing import (
 from numpy.typing import NDArray
 from typing_extensions import ParamSpec
 
-from careamics.config import DataConfig
 from careamics.config.support import SupportedData
 from careamics.file_io.read import ReadFunc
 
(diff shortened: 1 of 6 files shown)


CatEek (approved):
Lgtm, I don't have any strong opinions at this point. I think we'll understand more once we're done with the whole pipeline and will test it on someone

src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def from_arrays(
+    data_config: DataConfig, source: Sequence[NDArray], *args, **kwargs
+) -> list[InMemoryImageStack]:
+    axes = data_config.axes
+    return [InMemoryImageStack.from_array(data=array, axes=axes) for array in source]
+
+
+# TODO: change source to directory path? Like in current implementation
```
jdeschamps, line 88: Both could be nice to have. Leaving users in charge expose them to an error prone process. In my experience, people like being able to place everyhting in a folder with matching names between two folders, so it is a desirable feature.

src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
```diff
@@ -0,0 +1,138 @@
...
+
+    TODO: show example use in the `CAREamicsDataset`
+
+    The example above defines a `ZarrSource` dict because to determine _which_ ZARR
+    images will be loaded both a ZARR store and the internal data paths need to be
+    specified.
+    """
+
+    def __call__(
+        self, data_config: DataConfig, source: Any, *args: P.args, **kwargs: P.kwargs
```
jdeschamps: Shouldn't InferenceDataConfig also be accepted?

Until we solve that problem (see Dataset PR)

melisande-c: Ah right yeah... 

So the only thing that most of the `ImageStackLoaders` need from the `data_config` is the axes, so I could remove the `data_config` argument and replace with `axes`, which is what I had at first. I decided to use the `DataConfig` incase we add additional fields in the future that relate to the loading, but I think I am just over complicating things again... I think I will just revert to just having the axes field

melisande-c:
### Update

- Removed all the `PatchExtractor` constructors I thought I already removed.
- Removed all dependency on `DataConfig` and the functions just have arguments for the fields that they need, namely `axes` and `data_type`

PR #435, opened 2025-03-07
Fix: Move `enable_progress_bar` from `TrainingConfig` to `CAREamist` API initialisation 

### LINKED ISSUE #389: Cannot disable progress bar with `TrainingConfig.enable_progress_bar` when using `CAREamist` API
## Describe the bug

Setting `enable_progress_bar=False` in `TrainingConfig` results in `MisconfigurationException` from lightning because `ProgressBarCallback` is added to the callbacks in `CAREamist._define_callbacks` regardless.

## To Reproduce

```python
    experiment_name="ht_lif24",
    data_type="array",
    axes="SYXC",
    n_channels=2,
    patch_size=(64, 64),
    batch_size=64,
    num_epochs=3,
)
config.training_config.enable_progress_bar = False
careamist = CAREamist(source=config)
careamist.train(train_source=train_data,val_minimum_split=5)
```

Error message:
```pytb
---------------------------------------------------------------------------
MisconfigurationException                 Traceback (most recent call last)
Cell In[9], line 1
----> 1 careamist = CAREamist(source=config)
      2 careamist.train(train_source=train_data,val_minimum_split=5)

File /localscratch/miniforge3/envs/microSplit/lib/python3.9/site-packages/careamics/careamist.py:201, in CAREamist.__init__(self, source, work_dir, callbacks)
    198     experiment_logger = [csv_logger]
    200 # instantiate trainer
--> 201 self.trainer = Trainer(
    202     max_epochs=self.cfg.training_config.num_epochs,
    203     precision=self.cfg.training_config.precision,
    204     max_steps=self.cfg.training_config.max_steps,
    205     check_val_every_n_epoch=self.cfg.training_config.check_val_every_n_epoch,
    206     enable_progress_bar=self.cfg.training_config.enable_progress_bar,
    207     accumulate_grad_batches=self.cfg.training_config.accumulate_grad_batches,
    208     gradient_clip_val=self.cfg.training_config.gradient_clip_val,
    209     gradient_clip_algorithm=self.cfg.training_config.gradient_clip_algorithm,
    210     callbacks=self.callbacks,
    211     default_root_dir=self.work_dir,
    212     logger=experiment_logger,
    213 )
    215 # place holder for the datamodules
    216 self.train_datamodule: Optional[TrainDataModule] = None

...

File /localscratch/miniforge3/envs/microSplit/lib/python3.9/site-packages/pytorch_lightning/trainer/connectors/callback_connector.py:134, in _CallbackConnector._configure_progress_bar(self, enable_progress_bar)
    131     # otherwise the user specified a progress bar callback but also
    132    # elected to disable the progress bar with the trainer flag
    133     progress_bar_callback = progress_bars[0]
--> 134     raise MisconfigurationException(
    135         "Trainer was configured with `enable_progress_bar=False`"
    136         f" but found `{progress_bar_callback.__class__.__name__}` in callbacks list."
    137     )
    139 if enable_progress_bar:
    140     progress_bar_callback = TQDMProgressBar()

MisconfigurationException: Trainer was configured with `enable_progress_bar=False` but found `ProgressBarCallback` in callbacks list.
```

## Suggested solution

In `CAREamist._define_callbacks` `ProgressBarCallback` should only be added to the callbacks list if `config.training_config.enable_progress_bar` is `True`. See where code should change below.
https://github.com/CAREamics/careamics/blob/adacfe63e7af978561677d69f384afaf74415ccc/src/careamics/careamist.py#L246-L257

melisande-c:

If `enable_progress_bar` was set to `False` in the `TrainingConfig` it would cause an error at training when using the `CAREamist` API. This PR solves the problem by removing the `enable_progress_bar` field from the `TrainingConfig` since it is not important for reproducibility and only a visual thing.

### Overview - what changed?

Removed `enable_progress_bar` from `TrainingConfig`. Added `enable_training_config` argument to `CAREamist.__init__`.

If `enable_training_config` is set to `True` in `CAREamist` initialisation the `ProgressBarCallback` will be added to the callback list, if it is set to `False` then it will not.

- `CAREamist.__init__`
- `CAREamist._define_callbacks`

### Removed features or files

- `enable_progress_bar` field from `TrainingConfig`.

## How has this been tested?

Added `test_enable_progress_bar` test.

## Related Issues

- Resolves #389 

## Breaking changes

Loading any old config files, but they can easily converted by removing the `enable_prpgress_bar` field.

---

- [x] Pre-commit passes


jdeschamps (approved):

melisande-c:
@jdeschamps I could keep the field in the `DataConfig` but use the pydantic `deprecated` parameter to prevent breaking changes, or do you think that is unnecessary?

jdeschamps:
> @jdeschamps I could keep the field in the `DataConfig` but use the pydantic `deprecated` parameter to prevent breaking changes, or do you think that is unnecessary?

We don't need backward compatibility right now, not until we have a stable release.

We should check whether it breaks something in https://github.com/CAREamics/MicroSplit-reproducibility.

melisande-c:
> > @jdeschamps I could keep the field in the `DataConfig` but use the pydantic `deprecated` parameter to prevent breaking changes, or do you think that is unnecessary?
> 
> We don't need backward compatibility right now, not until we have a stable release.
> 
> We should check whether it breaks something in https://github.com/CAREamics/MicroSplit-reproducibility.

Right yeah, good point, I don't think CAREamics version has been fixed in the MicroSplit-reproducibility pyproject.toml, probably should happen since #396 seems like it changes some stuff that the repo uses as well

Issue #425, opened 2025-02-26
Next-Gen Dataset: Patching Strategies


melisande-c:
In #373 an initial draft of how to switch between patching strategies was introduced, i.e. the `PatchSpecsGenerator` protocol, the whole structure of this class needs rethinking. (Also if anyone has a better name for this that would be appreciated).

We need to reconsider carefully all the required features, for example: maybe we want to be able to implement them as python generators, see https://github.com/CAREamics/careamics/pull/420#discussion_r1971287633 for motivation.

melisande-c:
Also see idea for `WholeSamplePatchSpecsGenerator` https://github.com/CAREamics/careamics/pull/420#discussion_r1971300013

melisande-c:
Also see this other comment chain on what parameters should be arguments vs attributes https://github.com/CAREamics/careamics/pull/420#discussion_r1971270099. 

It probably makes sense to just generate patch specs from the list `ImageStack`s, and then different patching strategies can use whatever metadata they need from within the `ImageStack`. This could be for example the `chunk_size` from the `ZarrImageStack`, see #410.

This would make the `PatchSpecsGenerator` more uniform – they would all only have the following as the function signature to generate patches. All additional parameters have to be attributes.

```python
def generate(self, image_stacks: list["ImageStack], patch_size: Sequence[int]) -> list[PatchSpecs]:
```

melisande-c:
My proposed idea for random patching, so that the random patches are different each epoch is make the patch specs generators have a `__getitem__` method which will be indexed in the dataset `__getitem__`. 
- For a deterministic patch specs generator, like the sequential patching strategy, this can basically just return the `PatchSpecs` from a list stored in the class instance. 
- For the random patching strategy it will return a random patch. 

The constraints for the random patching strategy is that there should still be a known number of patches produced for each sample. To do this we can create bins that determine which patch index belongs to each sample. Then, for a given patch index, we can return the `PatchSpecs` with random coords but a `data_index` and `sample_index` that belongs to the expected sample. Below is a rough sketch 

```python
class RandomPatchSpecsGenerator:

    def __init__(self, image_stacks: Sequence[ImageStack], patch_size: Sequence[int]):
        self.patch_size = patch_size
        self.data_shapes = [image_stack.data_shape for image_stack in image_stacks]
        self.n_patches_per_stack = [
            self._n_patches(image_stack.data_shape, self.patch_size)
            for image_stack in image_stacks
        ]
[...]

Issue #436, opened 2025-03-11
Next-Gen Dataset: Data configuration changes


veegalinova:
Following the comments in https://github.com/CAREamics/careamics/pull/420, here are the following points for changes in data configuration: 
- Dataset can be configured with `DataConfig` or `InferenceConfig`. These configs contain similar parameters, but the following params require extra logic and checking:
   - `patch_size` vs `tile_size` and `tile_overlap`
   - `transforms` vs `tta_transforms` 
   - `dataloader_params` can not be defined in `InferenceConfig`
- Add an optional `random_seed` parameter

jdeschamps:
It may not be the right choice or tool, but we could examine whether Pydantic `aliases` can allow us to keep the same configuration but use clear naming in the functions. 

One important point being that the data configuration used for training is saved in the configuration yam, but not the inference configuration.

jdeschamps:
Solved by https://github.com/CAREamics/careamics/pull/486

Issue #416, opened 2025-02-19
Next-Gen Dataset: Validation data splitting


melisande-c:
The current datasets have a `split_dataset` method to separate some data for validation. Now that we're moving away from sequential patching we need to make sure we partition the validation data in a way that we don't include overlapping patches between validation and training sets.

jdeschamps:
Not a mandatory requirement, but a potential good to have: a validation-specific `seed` would also be useful if it would ensure that the validation crops are always the same between repetitions.

(somewhat linked to https://github.com/CAREamics/careamics/issues/297).

melisande-c:
> we don't include overlapping patches between validation and training sets.

This is easy with multiple images, since some images can be assigned to training and some can be assigned to validation; but if there is only 1 image, then it has to be split into sections for training and validation?

melisande-c:
> > we don't include overlapping patches between validation and training sets.
> 
> This is easy with multiple images, since some images can be assigned to training and some can be assigned to validation; but if there is only 1 image, then it has to be split into sections for training and validation?

If we had stratified-random-patching this would be easier?

jdeschamps:
Random thought because I just commented on the masking of background patches: if we had a mechanism to mask part of the image for exclusion, then the patches from the fixed random validation could be masked for training. It may be convoluted, but just for the sake of argument, the `CAREamics` dataset could have an `update_mask` method that could take a list of `PatchSpecs`. We could pass the list generated by the validation dataset.

jdeschamps:
Following current developments on background masking:

For N2V, we could generate a list of coordinates internally `val_coordinates` and the sampler could reject training patches based on this list. 

For the validation dataset, the job would be easier as we would just draw from this list.

PR #439, opened 2025-03-11
Feature(Next-Gen Dataset): Improved patching strategy classes

melisande-c:
## Description

> [!NOTE]  
> **tldr**: This PR addresses the shortcomings of the original draft implementation of the `PatchSpecsGenerator` classes, that were discovered after trying to use them in the prototype dataset in #420. This includes having a random patching strategy that will change each epoch. They have also been renamed to `PatchingStrategy` classes.

The previous `PatchSpecsGenerator` classes weren't working very well when we tried to use them in the dataset prototype in #420, and the name is also hard to say 😅. The main problem was that the random patches wouldn't change each epoch.

### Overview - what changed?

Added a `RandomPatchingStrategy`, `FixedRandomPatchingStrategy` and `SequentialPatchingStrategy`.

Renamed `PatchSpecsGenerator` to `PatchingStrategy`. Now the classes should also be initialised with the datashapes and the patch size, since for a given `CAREamicsDataset` instance we don't expect these to change. The `PatchingStrategy` protocol dictates that a concrete implementation should have the methods `n_patches` that should return the total number of patches, and a `get_patch_spec` method that takes an index and will return a `PatchSpecs` dict.

EDIT: `CAREamicsDataset` now takes a `mode` argument to determine whether it is training, validating or predicting. The `mode` will determine the patching strategy: `RandomPatching` for training, `FixedRandomPatching` for validating and `TilingStrategy` for predicting. (`TilingStrategy` is implemented in an upcoming PR).

The `RandomPatchingStrategy` creates index bins to work out what image_stack and sample a given patch index maps to, and then generates a random patch. This is to ensure the random patching still produces a fixed known number of patches for each image_stack and sample within the image_stack, but is then random spatial location within the single sample.

~~In the dataset the patching strategy should be a attribute and during `__getitem__`, `get_patch_spec` will be called to index the data.~~
EDIT: In the dataset the patching strategy is an attribute and during `__getitem__`, `get_patch_spec` is called to index the data.

See sketch code below:

```python

class CAREamicsDataset:

    def __init__(self, ...):
        patch_extractor = create_patch_extractor(...)
        patching_strategy = RandomPatchingStrategy(...) #maybe we want to make this swappable

    def __len__(self):
        return patching_strategy.n_patches()

    def __getitem__(self, index: int):
        patch_spec = self.patching_strategy.get_patch_spec(index)
        patch = patch_extractor.extract_patch(**patch_spec)
        return patch
```

### New features or files

- `RandomPatchingStrategy` - patches will change each epoch
- `FixedRandomPatchingStrategy` - patches are fixed each epoch
- placeholder `SequentialPatchingStrategy`

- A couple organisational changes, like moving the `PatchSpecs` definition.
- Exported all the notebooks to scripts because mypy isn't set up to look for errors in them
- Updated `demo_patch_extractor` to use the new `RandomPatchingStrategy`
- EDIT: `CAREamicsDataset.__init__`
- EDIT: `CAREamicsDataset.__getitem__`

### Removed features or files

- remove the method `extract_patches` from `PatchExtractor` because extracting multiple patches isn't needed anymore

## How has this been tested?

Added tests

## Related Issues

- #425 mostly solved apart from `WholeSamplePatchingStrategy`

I think I should wait for #420 to be merged and I will implement integration with dataset class

I will address tiling for prediction where we need the overlap coords in an up coming PR

---

- [x] Pre-commit passes


melisande-c:
### Updates after merge with main

- patching_strategy initialised as an attribute of `CAREamicsDataset`
- `patching_strategy.get_patch_spec` called in `CAREamicsDataset.__getitem__`
- Introduced `mode` argument to `CAREamicsDataset.__init__`, and `Mode` `Enum`
    - (This will be set by the `data_module`, in `data_module.setup`)
- temporarily raise `NotImplementedError` for `mode=Mode.Predicting` because `TilingStrategy` will be introduced in an upcomming PR.  
- Updated `SequentialPatchingStrategy` to match new `PatchingStrategy` interface (but it is no longer used)
- Added `SequentialPatchingStrategy` tests
- Give `mode` param to `CAREamicsDataset` in existing dataset tests.

src/careamics/dataset_ng/patching_strategies/patching_strategy_types.py
jdeschamps, line 1: The same way that we have `image_stack_protocol.py`, I think it would make sense to name this module `patching_strategy_protocol.py`.

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+
+    However the strategy still ensures that there will be a known number of patches for
+    each sample in each image stack. This is achieved through defining a set of bins
+    that map to each sample in each image stack. Whichever bin an `index` passed to
+    `get_patch_spec` falls into, determines the `"data_idx"` and `"sample_idx"` in
+    the returned `PatchSpecs`, but the `"coords"` will be random.
+
+    The number of patches in each sample is based on the number of patches that would
+    fit if they were sampled sequentially.
+    """
```
jdeschamps, line 28: Can we mention that it follows the PatchingStategy protocol and point to the module? It's annoying if we change names, but it is useful to the understanding.

(side comment: I have avoided looking into it but at some point we will need to review (globally) docstring and see if we can link between files for the website)

src/careamics/dataset_ng/patching_strategies/__init__.py
jdeschamps, line 1: When you'll be writing the readme, you could add the relevant paragraphs in the init files.

"""Patching strategy modules

Blah blah defines protocol, implemented in blabla for this and that purpose.
"""

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+        Parameters
+        ----------
+        data_shapes : sequence of (sequence of int)
+            The shapes of the underlying data. Each element is the dimension of the
+            axes SC(Z)YX.
+        patch_size : sequence of int
+            The size of the patch. The sequence will have length 2 or 3, for 2D and 3D
+            data respectively.
+        seed : int, optional
+            An optional seed to ensure the reproducibility of the random pathces.
```
jdeschamps: Type in `pathces`

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+    twice the two outputs can be different.
+
+    However the strategy still ensures that there will be a known number of patches for
+    each sample in each image stack. This is achieved through defining a set of bins
+    that map to each sample in each image stack. Whichever bin an `index` passed to
+    `get_patch_spec` falls into, determines the `"data_idx"` and `"sample_idx"` in
+    the returned `PatchSpecs`, but the `"coords"` will be random.
+
+    The number of patches in each sample is based on the number of patches that would
+    fit if they were sampled sequentially.
```
jdeschamps: Not important: "if they were sampled sequentially, non-overlapping, and covering the entire array."

jdeschamps: there's also the same sentence somewhere else

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+        data respectively.
+
+    Returns
+    -------
+    int
+        The number of patches.
+    """
+    if len(patch_size) != len(spatial_shape):
+        raise ValueError(
+            "Number of patch dimension do not match the number of spatial "
```
jdeschamps: Can we make this an fstring and, in parenthesis, add the shapes?

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # these bins will determine which image stack and sample a patch comes from
+        # the image_stack_index_bins map to each image stack
+        # the sample_index_bins map to each sample within each image stack
+        self.image_stack_index_bins, self.sample_index_bins = self._calc_patch_bins(
+            self.data_shapes, self.patch_size
+        )
+
+        # this is needed to calculate the sample_idx relative to the image_stack
+        samples_per_image_stack = [data_shape[0] for data_shape in self.data_shapes]
```
jdeschamps: This could be moved to `_calc_patch_bins` and done in the same loop as the other two bin lists

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+        Returns
+        -------
+        PatchSpecs
+            A dictionary that specifies a single patch in a series of `ImageStacks`.
+        """
+        # TODO: break into smaller testable functions?
+        if index >= self.n_patches:
+            raise IndexError(
+                f"Index {index} out of bounds for RandomPatchingStrategy with number "
+                f"of patches, {self.n_patches}"
```
jdeschamps: again super minor, I'd write `f"of patches {self.n_patches}"`. But feel free to ignore.

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+        # digitize returns the bin that `index` belongs to
```
jdeschamps, line 93: Ahahah thanks for the explanation, I was wondering what it was doing!

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+
+        data_shape = self.data_shapes[data_index.item()]
+        spatial_shape = data_shape[2:]
+
+        # calculate sample index relative to image stack:
+        #   subtract the total number of samples in the previous image stacks
+        if data_index == 0:
+            n_previous_samples = 0
+        else:
+            n_previous_samples = self.sample_bins[data_index - 1]
```
jdeschamps: on Line 98 you call `data_index.item()`, but here you use the `np.ndarray` directly

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+            A dictionary that specifies a single patch in a series of `ImageStacks`.
+        """
+        # TODO: break into smaller testable functions?
+        if index >= self.n_patches:
+            raise IndexError(
+                f"Index {index} out of bounds for RandomPatchingStrategy with number "
+                f"of patches, {self.n_patches}"
+            )
+        # digitize returns the bin that `index` belongs to
+        data_index = np.digitize(index, bins=self.image_stack_index_bins)
```
jdeschamps: Why not call `.item()` here already?

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+        Parameters
+        ----------
+        data_shapes : sequence of (sequence of int)
+            The shapes of the underlying data. Each element is the dimension of the
+            axes SC(Z)YX.
+        patch_size : sequence of int
+            The size of the patch. The sequence will have length 2 or 3, for 2D and 3D
+            data respectively.
+        seed : int, optional
+            An optional seed to ensure the reproducibility of the random pathces.
```
jdeschamps: same typo

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        self.rng = np.random.default_rng(seed=seed)
+        self.patch_size = patch_size
+        self.data_shapes = data_shapes
+
+        # simply generate all the patches at initialisation, so they will be fixed
+        self.fixed_patch_specs: list[PatchSpecs] = []
+        for data_idx, data_shape in enumerate(self.data_shapes):
+            spatial_shape = data_shape[2:]
+            for sample_idx in range(data_shape[0]):
```
jdeschamps, line 220: This is only returning a single random patch per sample? Is it not missing a loop over `_n_patches`?

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,308 @@
...
+
+    Raises
+    ------
+    ValueError
+        Raises if the number of spatial dimensions do not match the number of patch
+        dimensions.
+    """
+    if len(patch_size) != len(spatial_shape):
+        raise ValueError(
+            "Number of patch dimension do not match the number of spatial "
```
jdeschamps: Same here, can we add the shapes in the error message?

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -71,32 +80,41 @@ def __init__(
...
-            patch_generator: PatchSpecsGenerator = RandomPatchSpecsGenerator(
+    def _initialize_patching_strategy(self) -> PatchingStrategy:
+        patching_strategy: PatchingStrategy
+        if self.mode == Mode.TRAINING:
+            if isinstance(self.config, InferenceConfig):
+                raise ValueError("Inference config cannot be used for training.")
+            patching_strategy = RandomPatchingStrategy(
+                data_shapes=self.input_extractor.shape,
+                patch_size=self.config.patch_size,
+                # TODO: Add random seed to dataconfig
```
jdeschamps, line 97: I might have asked before, but can't remember. Do we have an issue for that?

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -71,32 +80,41 @@ def __init__(
...
+                seed=getattr(self.config, "random_seed", None),
+            )
+        elif self.mode == Mode.VALIDATING:
+            if isinstance(self.config, InferenceConfig):
+                raise ValueError("Inference config cannot be used for validating.")
+            patching_strategy = FixedRandomPatchingStrategy(
```
jdeschamps, line 103: Not to be addressed in this PR, but for validation we need to think a little bit what to do. Currently, we have the option to split a percentage of the dataset into validation when there is not validation explicitly given. It may also make sense to allow limiting the number of patches if we pass entire samples to the validation dataset.

jdeschamps (changes requested):
No comment on the structure, I like it very much. A lot of very minor points, the only important comment being on the fixed random patching returning a single patch per sample (or I misunderstood the code).

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: Oops! I think you are right! Thanks for spotting!!!

src/careamics/dataset_ng/patching_strategies/patching_strategy_types.py
melisande-c: I called it `patching_strategy_types` because it also includes the `PatchSpecs` type, but I can change it

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: yeah true, this was an after thought, I will move it

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: makes sense

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: `self.sample_bins` is a numpy array and `self.data_shape` is a list, which is why it works for one and not the other, but yes if I just called `.item` at the start that would simplify things. 

But also I should be more consistent with whether things are lists or arrays.

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: It is a point in #436

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: As I mention in #416, the random patching is going to make validation spitting very annoying if you only have 1 image. Previously the in-memory dataset used fixed sequential patching so some patches could be separated for validation, but now the patches are random each epoch.

So we either go back to sequential patching for this case, or we find a way to split the image into validation and training sections.

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: I'll also add a test to `test_all_strategies` that ensures some percentage of input is sampled by the patching.

src/careamics/dataset_ng/patching_strategies/patching_strategy_types.py
jdeschamps: Also makes sense, I don't have a strong opinion. It can stay like this.

src/careamics/dataset_ng/patching_strategies/random_patching.py
jdeschamps: Got it thanks!

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -5,12 +5,13 @@
...
 import numpy as np
 
-from .patching_strategy_types import PatchSpecs
+from .patching_strategy_protocol import PatchSpecs
 
 
 class RandomPatchingStrategy:
     """
-    A patching strategy for sampling random patches.
+    A patching strategy for sampling random patches, it implements the `PatchinStrategy`
```
jdeschamps: Sorry small typo, missing a "g"

jdeschamps (approved):

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,325 @@
...
+        if index >= self.n_patches:
+            raise IndexError(
+                f"Index {index} out of bounds for FixedRandomPatchingStrategy with "
+                f"number of patches, {self.n_patches}"
+            )
+        # simply index the pre-generated patches to get the correct patch
+        return self.fixed_patch_specs[index]
+
+
+def _random_coords(
```
veegalinova: Maybe names of this and `_n_patches` functions could be a bit more explicit? 
Like `_generate_random_coords` and `_calculate_n_patches`

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,325 @@
...
+    spatial_shape: Sequence[int], patch_size: Sequence[int], rng: np.random.Generator
+) -> tuple[int, ...]:
+    """Generate random patch coordinates for a given `spatial_shape` and `patch_size`.
```
veegalinova, line 267: Question: is there a need for detailed docstrings for the private functions that are only used internally? (and do not contain complicated logic)
Do we need the docstrings in these cases at all? To me it seems like unnecessary work, but happy to hear other opinions.

tests/dataset_ng/patching_strategies/test_sequential_patching_ng.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        # set to true where the patches would be sampled from
+        tracking_array[(patch_spec["sample_idx"], slice(None), *spatial_slice)] = True
+
+    for tracking_array in tracking_arrays:
+        # if the patch specs covered all the image all the values should be true
+        assert tracking_array.all()
+
+
+# TODO: what else to test?
```
veegalinova, line 58: Might also test that each location is covered only once? I'll need to revisit the implementation anyways, will take a mental note

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -0,0 +1,325 @@
...
+        """
+        self.rng = np.random.default_rng(seed=seed)
+        self.patch_size = patch_size
+        self.data_shapes = data_shapes
+
+        # these bins will determine which image stack and sample a patch comes from
+        # the image_stack_index_bins map to each image stack
+        # the sample_index_bins map to each sample within each image stack
+        # sample_bins is needed to calculate the sample_idx relative to the image_stack
+        self.image_stack_index_bins, self.sample_index_bins, self.sample_bins = (
```
veegalinova: For some reason these variable names are hard to read for me 
`index_bins_per_image_stack` `index_bins_per_image_sample` like in the function below seem marginally better...

veegalinova (approved):
Looks good, nothing substantial to add!

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: renamed 👍

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: When we turn numpy docsting validation back on in the pre-commit it will complain yeah, there probably are settings you can tweak. (I am planning to turn it back on probably next week)

But tbh I don't mind enforcing docstrings for all functions if the trade-off is sometimes having to write dostrings for fairly self explanatory functions, sometimes private functions can have confusing logic and future-us will probably be grateful for some docs.

src/careamics/dataset_ng/patching_strategies/random_patching.py
melisande-c: Yeah I agree the names are not the best. However I find "per" here a little misleading here because there is one set of bins that are used to map a patch-index to an image-stack or a sample... Not sure how to resolve, I could try to improve the comments?

src/careamics/dataset_ng/patching_strategies/random_patching.py
jdeschamps: I do agree on the last point, I know it is a pain, but it will alleviate pain for others in the future (if the doc is written well enough...). We can keep it minimum for self-explanatory functions.

src/careamics/dataset_ng/patching_strategies/random_patching.py
veegalinova: I think comments are fine as they are! 

I asked gpt for variable name ideas, got: 
`dataset_bins` (instead of image_stack_index_bins): These bins map to each separate dataset/volume in the collection
`global_sample_bins` (instead of sample_index_bins): These bins map to individual samples across all datasets
`samples_per_dataset` (instead of sample_bins): Keeps track of how many samples are in each dataset cumulatively

Still not great 😆

melisande-c: `samples_per_dataset` is just wrong, the bins are the cumulative version, (and also the others don't make sense to me)

I could do 
`cumulative_image_stack_indices`
`cumulative_sample_indices`
`cumulative_samples`
?

or 
`cumulative_patches_per_image_stack`
`cumulative_patches_per_sample`
`cumulative_samples_per_image_stack`

melisande-c: I went for
```
image_stack_cumulative_patches,
sample_cumulative_patches,
image_stack_cumulative_samples
```
hopefully that hasn't made it worse

PR #444, opened 2025-03-24
Feature(Next-Gen Dataset): Prediction patching (tiling)

### LINKED ISSUE #425: Next-Gen Dataset: Patching Strategies

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Introducing `PatchingStrategies` for prediction. These are a `TilingStrategy` that will output a `TileSpecs` object, which inherits from the `PatchSpecs` type but includes the additional fields `crop_coords`, `crop_size` and `stitch_coords`; the second patching strategy is the `WholeSampleStrategy`, this is for prediction without tiling and will select a whole sample.

To be able to stitch predicted tiles back together we need some additional information that is not included in the current `PatchSpecs` type. These are the: 
- `crop_coords`: where will the tile be cropped from to deal with overlaps, this is the top-left corner, relative to the tile.
- `crop_size`: the size of the crop. So the tile will be cropped to `crop_coords: crop_coords + crop_size`
- `stitch_coords`: where the cropped tiles will be placed in the stitched image, top-left corner, relative to the image.
So this PR introduces the `TileSpecs` `TypedDict` to store this information, and a `TilingStrategy` class to produce it.

### Overview - what changed?

In the `CAREamicsDataset` if the `mode` is set to predicting, it will chose the `TilingStrategy` if both the `tile_size` and `tile_overlaps` are not `None` in the `InferenceConfig`, otherwise it will choose the `WholeSampleStrategy`.

Also a `Imageregions_to_tileinfos` function has been introduced to convert the new `ImageRegionData` to `TileInformation` objects so the existing `stitch_prediction` functions can be used as they are (but they should be updated in the future, once the dataset is complete).

The implementation of the `TilingStrategy` is largely adapted from the existing `extract_tiles` function.

### New features or files

- `TileSpecs` typed dict
- `TilingStrategy` patching strategy
- `WholeSamplePatchingStrategy` patching strategy
- `imageregions_to_tileinfos` function

- `CAREamicsDataset` patching strategy initialisation

## How has this been tested?

- Added test to make sure the stitched array is equal to the original array, when using tiles generated from `TilingStrategy` and the existing CAREamics `stitch_prediction` functions, in both 2D and 3D.
- Added `TilingStrategy` and `WholeSamplePatchingStrategy` to the `test_all_strategies.py` tests.

## Related Issues

- Resolves #425 

---

- [x] Pre-commit passes


src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -39,7 +42,9 @@ class ImageRegionData(NamedTuple):
     data_shape: Sequence[int]
     dtype: str  # dtype should be str for collate
     axes: str
-    region_spec: PatchSpecs
+    # Union is kinda ugly, could be solved with generics but generic NamedTuples are not
+    #   supported in python 3.9
+    region_spec: Union[PatchSpecs, TileSpecs]
```
jdeschamps: can you leave a TODO so that it is easier to find again once we remove py39 support?

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -157,10 +172,20 @@ def __getitem__(
         self, index: int
     ) -> tuple[ImageRegionData, Optional[ImageRegionData]]:
         patch_spec = self.patching_strategy.get_patch_spec(index)
-        input_patch = self.input_extractor.extract_patch(**patch_spec)
+        input_patch = self.input_extractor.extract_patch(
```
jdeschamps, line 145: Any reason for that change? just to be explicit?

(I just want to make sure I didn't miss something)

src/careamics/dataset_ng/patching_strategies/tiling_strategy.py
```diff
@@ -0,0 +1,172 @@
...
+        tile_size : sequence of int
+            The size of the tile. The sequence will have length 2 or 3, for 2D and 3D
+            data respectively.
+        overlaps : sequence of int
+            How much a tile will overlap with adjacent tiles in each spatial dimension.
+        """
+        self.data_shapes = data_shapes
+        self.tile_size = tile_size
+        self.overlaps = overlaps
+        # TODO: check tile_size and overlap have same length
```
jdeschamps: I'd say this should probably be done in the configuration (see the other PR)

(meaning. validating that size and overlaps have the same length)

src/careamics/dataset_ng/patching_strategies/whole_sample.py
```diff
@@ -0,0 +1,35 @@
+from collections.abc import Sequence
+
+from .patching_strategy_protocol import PatchSpecs
+
+
+class WholeSamplePatchingStrategy:
+    # Warning: can only be used with batch size = 1
```
jdeschamps: Why is that? Wondering if I am missing something here!

(thinking of a "SYX" case where the stack has dimensions (100, 64, 64) for instance -> edge case I agree)

jdeschamps (approved):
Looks good!

Would it be worth refactoring the choice of patching strategy out of the dataset into a patching factory?

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: If the patch_spec is a `TileSpec` it has extra fields now (`crop_coords`, `crop_size`, and `stitch_coords`) so unpacking won't work

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: I think I might not actually need this union anymore because I just use `typing.cast` to cast the `PatchSpec` to a `TileSpec`. Which is also ugly and I forgot to put a TODO there, so I'll add it. (Also what I was trying to do with generics is not supported until 3.11, so yeah, I guess it will probably stay like this for a while)

src/careamics/dataset_ng/patching_strategies/tiling_strategy.py
melisande-c: Ah yeah good point, I'll remove the TODO

src/careamics/dataset_ng/patching_strategies/whole_sample.py
melisande-c: So yes it would work in this case, but it won't work in the case of multiple image stacks with different dimensions

melisande-c:
> Looks good!
> 
> Would it be worth refactoring the choice of patching strategy out of the dataset into a patching factory?

I was thinking this, this would actually remove the need for the dataset to know whether it is in "training", "validation" or "prediction" mode. In relation to your PR #445, this would mean there could be a nested `PatchingStrategyConfig`, that would have fields relevant to each patching strategy. It would mean that there wouldn't need to be a separate `PredictionDataConfig` any more. Instead a function `create_prediction_dataconfig` would just create a `DataConfig` with a `TilingStrategy` instance.

melisande-c:
> > Looks good!
> > Would it be worth refactoring the choice of patching strategy out of the dataset into a patching factory?
> 
> I was thinking this, this would actually remove the need for the dataset to know whether it is in "training", "validation" or "prediction" mode. In relation to your PR #445, this would mean there could be a nested `PatchingStrategyConfig`, that would have fields relevant to each patching strategy. It would mean that there wouldn't need to be a separate `PredictionDataConfig` any more. Instead a function `create_prediction_dataconfig` would just create a `DataConfig` with a `TilingStrategy` instance.

But maybe this can be an additional PR 😅

src/careamics/dataset_ng/dataset/dataset.py
jdeschamps: Got it!

src/careamics/dataset_ng/patching_strategies/whole_sample.py
jdeschamps: Can we catch that and warn users wherever the patching strategy is instantiated?

jdeschamps:
> > Looks good!
> > Would it be worth refactoring the choice of patching strategy out of the dataset into a patching factory?
> 
> I was thinking this, this would actually remove the need for the dataset to know whether it is in "training", "validation" or "prediction" mode. In relation to your PR #445, this would mean there could be a nested `PatchingStrategyConfig`, that would have fields relevant to each patching strategy. It would mean that there wouldn't need to be a separate `PredictionDataConfig` any more. Instead a function `create_prediction_dataconfig` would just create a `DataConfig` with a `TilingStrategy` instance.

I like that!

> > > Looks good!
> > > Would it be worth refactoring the choice of patching strategy out of the dataset into a patching factory?
> > 
> > 
> > I was thinking this, this would actually remove the need for the dataset to know whether it is in "training", "validation" or "prediction" mode. In relation to your PR #445, this would mean there could be a nested `PatchingStrategyConfig`, that would have fields relevant to each patching strategy. It would mean that there wouldn't need to be a separate `PredictionDataConfig` any more. Instead a function `create_prediction_dataconfig` would just create a `DataConfig` with a `TilingStrategy` instance.
> 
> But maybe this can be an additional PR 😅


yes :)

src/careamics/dataset_ng/patching_strategies/whole_sample.py
melisande-c: with a log message?

jdeschamps: Yes, but for now I guess your warning can just be a todo. we should just not forget that if something can break, we should catch it early!

src/careamics/dataset_ng/patching_strategies/patching_strategy_protocol.py
```diff
@@ -28,6 +28,37 @@ class PatchSpecs(TypedDict):
...
+        The top-left side of where the tile will be cropped, in coordinates relative
+        to the tile.
+    crop_size: sequence of int
+        The size of the cropped tile.
+    stitch_coords: sequence of int
+        Where the tile will be stitched back into an image, taking into account
+        that the tile will be cropped, in coords relative to the image.
+    """
+
+    crop_coords: Sequence[int]
```
veegalinova, line 57: I have a small suggestion for namings here since there is no need to keep the previous ones for compatibility: `crop_coords` -> `tile_start`, `crop_size` -> `tile_size`, `stitch_coords` -> `stitch_start`
Coords means both start and end to me, and I think `tile` is better for consistency

veegalinova (approved):
Everything is great, just worked out of the box when I tried it in the pipeline!

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: I removed the union and the comment

src/careamics/dataset_ng/patching_strategies/patching_strategy_protocol.py
melisande-c: Do you think this should change in the `PatchSpecs` as well for consistency? It has `coords` which are only the start of the patch.

The patch vs tile distinction is confusing but we should probably figure it out - does tile refer to all the data before and after cropping, or only after cropping? Since `TileSpecs` inherits from `PatchSpecs` it also has `patch_size`, so `crop_size` is the size after it has been cropped to remove the the overlaps. So if the name "tile" refers to only after cropping then this name change would make sense

jdeschamps: Interestingly enough I think of tiles as the pre-cropping. But @veegalinova's suggestion and redefinition (at least for me) would clarify the parameters and maybe raise clarity in the code.

PR #445, opened 2025-03-25
feat: Next-Gen Dataset - Uniform data configuration

### LINKED ISSUE #436: Next-Gen Dataset: Data configuration changes

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Have Training and Prediction data configurations share parameters.

This PR is an attempt at exploring how to improve the Pydantic data models for the next generation dataset.

In the current code base, `DataConfig` and `InferenceConfig` share a lot of parameters, or have similar parameters that are differently named (e.g. `patch_size` vs `tile_size`, `transforms` vs `tta_transform`). This adds unnecessary complexity in checking and calling these parameters.

See https://github.com/CAREamics/careamics/issues/436.

### Overview - what changed?

In order to simplify handling the data configs, this PR does the following:
- Introduce a parent Pydantic model that defines parameters (taking the namespace `DataConfig`).
- Define two child Pydantic models that add additional constraints and validations to the parameters defined in the parent class.

- `DataConfig` has now extended parameters with fewer constraints (e.g. optional patches, optional new parameter called `patch_overlap`), including a new `random_seed` parameter, and some of the validations
- The training configuration is now called `TrainingDataConfig` and adds constraints that are necessary during training: non-null patches, defaults for transforms and dataloader etc.
- The inference configuration is `PredictionDataConfig` and adds constraints that are necessary during prediction.

Importantly, `TrainingDataConfig` and `PredictionDataConfig` share all parameters, including `patch_size`, `patch_overlap`, `transforms` etc. In `PredictionDataConfig`, `patch_size`, `patch_overlap` and `train_dataloader_params` have aliases, which means that the configurations have to be instantiated as follow:

``` python
t = TrainingDataConfig(
    data_type="array",
    axes="YX",
    patch_size=[64, 64]
)
print(t.patch_size)

p = PredictionDataConfig(
    data_type="array",
    axes="YX",
    tile_size=[64, 64], # optional, could be omitted
)
print(p.patch_size) # the parameter is still accessed as patch_size
```

As far as the NG Dataset is concerned, we only need to declare the `DataConfig` type:

``` python
class CareamicsDataset(Dataset):
    def __init__(
        self,
        data_config: DataConfig,
        ...
```

Then, whenever parameters are passed to another function, to avoid `mypy` errors, we still need to have a check:
```python
if not isinstance(self.config, TrainingDataConfig):
    raise ValueError("Prediction config cannot be used for validating.")

...
```

But the need to specify different parameters is now alleviated, example:

``` python
        ...
        return Compose(
            transform_list=list(self.config.transforms), # works for both training and prediction (tta)
        )
```

### New features or files

- `general_data_model.py` contains now `DataConfig`, which is a general parent (Pydantic) model
- New file `prediction_data_model.py` contains the prediction data model
- New file `training_data_model.py` contains the training data model

- `CAREamicsDataset` was modified to take a `DataConfig` and to simplify where the parameters are now identical between the two child models.

### Removed features or files

-

## How has this been tested?

Using the jupyter notebook provided in the `dataset_ng` module.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/436

### Should we stick to aliases

Aliases bring a strange behaviour:
```python
p = PredictionDataConfig(
    data_type="array",
    axes="YX",
    tile_size=[64, 64], # optional, could be omitted
)
print(p.patch_size) # the parameter is still accessed as patch_size
```

This means that `PredictionDataConfig` is closer upon instantiation to the current API of the CAREamist and the `create_predict_datamodule`, and internally not different from the training data config.

But passing a parameter named differently to the member parameter is strange and we might just as well stick to `patch_shape` and `patch_overlap`. We can still have the `tiling` vocabulary elsewhere.

### Unused parameters

Now both the training and prediction data config have unused/unrelated parameters (e.g. overlaps for training, validation dataloader for prediction). I don't think that this is a problem as long as the default values and the descriptions are clear (e.g. "unused"). Two ways to go about it:

- Ignore them, don't validate them, they will/should not be called anyway
- Validate against them, enforce for instance that `patch_overlaps` is `None` in training data 

I think we could try the first approach, if our code is well compartmentalized that should not be an issue. Now it may introduce strange behaviour the day we pass the wrong config to the wrong function... So solution 2 would guard against it.

### Training dataloader parameters

Currently both models have `training_dataloader_params`, but this is quite awkward for prediction. We should probably rename that to `dataloader_params` and in the description explicitly say that it is either training or prediction.

### TODOs

Left to do or consider:

- [ ] Naming change `TrainingDataConfig` -> `TrainDataConfig`
- [ ] Replacing `InferenceConfig` by the `PredictionDataConfig`
- [x] Implementing validators in `PredictionDataConfig`
- [ ] Should we keep `train_dataloader_params` name even though it will be called for prediction as well?
- [ ] Decide on whether to keep aliases
- [ ] Discuss constraints on unused parameters
- [ ] Clean up jupyter notebook (used for testing)
- [ ] Add tests

### Merging order

This may impact https://github.com/CAREamics/careamics/pull/444 and should be merged afterwards (if approved).

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/config/data/prediction_data_model.py
+++ b/src/careamics/config/data/prediction_data_model.py
@@ -1,8 +1,11 @@
 """Prediction data configuration."""
 
+from __future__ import annotations
+
 from typing import Any, Optional, Union
 
-from pydantic import Field
+from pydantic import Field, model_validator
+from typing_extensions import Self
 
 from .general_data_model import DataConfig
 
@@ -55,4 +58,37 @@ class PredictionDataConfig(DataConfig):
     val_dataloader_params: Optional[dict[str, Any]] = Field(default=None)
     """This parameter is unused during prediction."""
 
-    # TODO: validate length patch size and patch overlap vs axes
+    @model_validator(mode="after")
+    def validate_dimensions(self: Self) -> Self:
+        """
+        Validate 2D/3D dimensions between axes, patch size and patch overlaps.
+
+        Returns
+        -------
+        Self
+            Validated data model.
+
+        Raises
+        ------
+        ValueError
+            If the patch size and overlaps do not match axes.
+        """
+        if self.is_tiled():
+            assert self.patch_size is not None
+            assert self.patch_overlaps is not None
+
+            if "Z" in self.axes:
+                if len(self.patch_size) != 3 or len(self.patch_overlaps) != 3:
+                    raise ValueError(
+                        f"Patch size and overlaps must have 3 dimensions if the data is"
+                        f" 3D (got axes {self.axes})."
+                    )
+
+            else:
+                if len(self.patch_size) != 2 or len(self.patch_overlaps) != 2:
+                    raise ValueError(
+                        f"Patch size and overlaps must have 2 dimensions if the data is"
+                        f" 2D (got axes {self.axes})."
+                    )
+
+        return self
--- a/src/careamics/config/data/prediction_data_model.py
+++ b/src/careamics/config/data/prediction_data_model.py
@@ -1,0 +1,55 @@
+"""Prediction data configuration."""
+
+from typing import Any, Optional, Union
+
+from pydantic import Field
+
+from .general_data_model import DataConfig
+
+
+class PredictionDataConfig(DataConfig):
+    """Prediction data configuration.
+
+    The following parameters defined in the parent `DataConfig` are aliased:
+    - `path_size` -> `tile_size`
+    - `patch_overlap` -> `tile_overlap`
+    - `train_dataloader_params` -> `dataloader_params`
+
+    The aliases should be used to pass the parameters to the model.
+
+    As opposed to `DataConfig`, the `val_dataloader_params` is set to `None` by default.
+    Likewise, `patche_size` (`tile_size`) and `patch_overlap` (`tile_overlap`) can be
+    `None` to predict on the entire image at once.
+
+    Examples
+    --------
+    Minimum example:
+
+    >>> data = PredictionDataConfig(
+    ...     data_type="array", # defined in SupportedData
+    ...     batch_size=4,
+    ...     axes="YX"
+    ... )
+
+    """
+
+    patch_size: Optional[Union[list[int]]] = Field(
+        default=None, min_length=2, max_length=3, alias="tile_size"
+    )
+    """Tile size used during prediction. If `None`, the entire image is predicted at
+    once."""
+
+    patch_overlap: Optional[Union[list[int]]] = Field(
+        default=[48, 48], min_length=2, max_length=3, alias="tile_overlap"
+    )
+    """Overlap between tiles during prediction."""
+
+    train_dataloader_params: dict[str, Any] = Field(
+        default={}, validate_default=True, alias="dataloader_params"
+    )
+    """Dictionary of PyTorch prediction dataloader parameters."""
+
+    val_dataloader_params: Optional[dict[str, Any]] = Field(default=None)
+    """This parameter is unused during prediction."""
+
+    # TODO: validate length patch size and patch overlap vs axes
--- a/src/careamics/config/data/prediction_data_model.py
+++ b/src/careamics/config/data/prediction_data_model.py
@@ -46,6 +46,7 @@ class PredictionDataConfig(DataConfig):
     )
     """Overlap between tiles during prediction."""
 
+    # TODO: should we enforce `suffle=False` for prediction? does it matter?
     train_dataloader_params: dict[str, Any] = Field(
         default={}, validate_default=True, alias="dataloader_params"
     )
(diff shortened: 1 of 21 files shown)


src/careamics/config/data/prediction_data_model.py
```diff
@@ -0,0 +1,94 @@
...
+
+    patch_overlaps: Optional[Union[list[int]]] = Field(
+        default=[48, 48], min_length=2, max_length=3, alias="tile_overlaps"
+    )
+    """Overlap between tiles during prediction."""
+
+    # TODO: should we enforce `suffle=False` for prediction? does it matter?
+    train_dataloader_params: dict[str, Any] = Field(
+        default={}, validate_default=True, alias="dataloader_params"
+    )
```
melisande-c, line 55: Yes shuffle does need to be False because in the current implementation the stitch prediction will not work if they are not in order. Also I think users would expect to get their predictions in the same order that they put the inputs.

In my new tiling PR #444, the predictions are reordered to use the original stitch prediction functions.

src/careamics/config/data/prediction_data_model.py
```diff
@@ -0,0 +1,94 @@
...
+    """Dictionary of PyTorch prediction dataloader parameters."""
+
+    val_dataloader_params: Optional[dict[str, Any]] = Field(default=None)
+    """This parameter is unused during prediction."""
```
melisande-c, line 59: Could this field just only be included in the `TrainingDataConfig` and not the underlying `DataConfig`? Maybe this would add more annoying checks later though.

src/careamics/config/data/prediction_data_model.py
```diff
@@ -0,0 +1,94 @@
...
+
+from pydantic import Field, model_validator
+from typing_extensions import Self
+
+from .general_data_model import DataConfig
+
+
+class PredictionDataConfig(DataConfig):
+    """Prediction data configuration.
+
```
melisande-c, line 15: Where are tta_transforms added? sorry if I missed it

melisande-c:
With the aliases, which name is exported on `model_dump`?

src/careamics/config/data/prediction_data_model.py
jdeschamps: Makes sense!

src/careamics/config/data/prediction_data_model.py
jdeschamps: Yes, it would mean that every time we call `config.val_dataloader_params` (arguably this would happen only once), we would need to check that it is a `TrainingDataConfig`. Although, we may only pass specifically `TrainingDataConfig` where this happens!

I don't have a strong opinion here, a bit more clarity vs potentially fewer checks in the code. Maybe clarity trumps?

jdeschamps: Actually:

`CAREamist.train` would access specifically a `TrainingDataConfig`, so no check needed. Same in prediction since we instantiate ourselves the dataconfig.

src/careamics/config/data/prediction_data_model.py
jdeschamps: `tta_transform` -> `transforms`.

But this needs a bit of thinking because TTA transforms are applied very differently (not via `Compose`).

jdeschamps:
> With the aliases, which name is exported on `model_dump`?

The aliases! But the prediction config is never (currently at least) exported so the benefit here is very minimal if not non-existant.

src/careamics/config/data/prediction_data_model.py
melisande-c: I mean where is the actual instance added? since the default is empty list

jdeschamps: I am not sure I understand, by default we don't use TTA (unless I am mistaken currently CAREamics.predict has default tta=False) so the default is empty.

melisande-c: Ok sorry I was getting confused because I didn't see changes in `PredictDataModule` which uses the `InferenceConfig`. But this PR keeps the existing functionality to work with the existing datasets but introduces an alternative `PredictDataConfig` that the new dataset can use! got it

melisande-c (approved):

src/careamics/config/data/prediction_data_model.py
jdeschamps: Ah yes that was not clear from the description!

veegalinova: In my mind, `tta_transforms` and `transforms` are two different things, and there are use cases where both could be used together, so there should be two distinct keys

veegalinova (approved):

jdeschamps:
Superseded by https://github.com/CAREamics/careamics/pull/486

PR #448, opened 2025-03-27
Feature(Next-Gen Dataset): Limit number of files loaded at a time + refac of dataset initialisation

### LINKED ISSUE #402: Next-Gen Dataset: PyTorch `Sampler` for multiple files
If we want to reimplement the case where a user may have many tiff files that cannot be all loaded into memory, we can implement a custom PyTorch `Sampler`, [see docs](https://pytorch.org/docs/stable/data.html#torch.utils.data.Sampler). 

This way we can ensure that patches are selected from only a few files at a time. Once one set of files have been sampled from, we can deallocate the memory and move onto the next batch of files.

### LINKED ISSUE #406: Next-Gen Dataset: Lazy file loading
To have a PyTorch `Sampler` to limit memory usage as proposed in #402, there should be an object that will load a Tiff file only when it is ready to be sampled and then deallocated the memory when moving onto the next file, or batch of files.

This will probably happen at the `ImageStack` and `PatchExtractor` level, see #373 where these classes were introduced.

melisande-c:
## Description

> [!NOTE]  
> **tldr**: If not all of the files (input + target) fit into the available memory then we want to only load a batch of files at a time. This emulates the functionality of the previous `IterableDataset`, and also extends it since the patches from multiple files can be shuffled together. Also included is a refactoring of how the dataset is initialised, including the patch extractor factory functions.

Note: Sorry this PR got kinda big 😅. I can split the refactoring to a different PR if it's confusing.

If someone has many files that won't all fit into memory we want to load batches of the files at a time, this will be more relevant for MicroSplit and possibly CARE rather than N2V.

The need for the refactoring came to make it easier to keep track of what underlying `ImageStack` are stored in the CAREamicsDataset.

#### FIFO lazy file loading

Created a first-in-first-out (FIFO) PyTorch `BatchSampler`, a `ManagedLazyImageStack` and a `FifoImageStackManager`. The idea is that the batch indices are output in an order that means only the indices from `n` image stacks are output at a time, where `n` is determined by the available memory. The `ManagedLazyImageStack` will only load its data the first time `extract_patch` is called, and the `FifoImageStackManager` will send a signal to close the least recently loaded image stack - following the FIFO strategy.

#### Dataset initialisation refactor

The `CAREamicsDataset` initialisation was refactored so that it no longer controls the construction of the `PatchExtractors` but gets given them as arguments - following the dependency injection pattern. 

Additionally, I removed all the `ImageStackLoader` implementations and left `ImageStackLoader` to only be used for the custom image stack loading. Instead I created separate patch extractor factory functions instead of having overloads. I did this because having a really long chain of function calls seemed annoying.

Now there is also a `create_dataset` convenience function, and a `create_dataloader` convenience function. The latter is needed because for the lazy loading the custom batch sampler needs to be set.

The `FifoImageStackManager` and the `ManagedLazyImageStack` work together using a callback system, to keep coupling between the classes low. The `ManagedLazyImageStack` can be given callback functions that are methods of the `FifoImageStackManager` that means the manager can be notified every time a registered image stack is loaded or closed.

If a new image stack want to load, the manager is notified, if there are already `n` image stacks loaded the manager will close the image stack that was loaded least recently.

The `FifoBatchSampler` needs to produce the file indices in an order that means each image stack will only be loaded and closed once (to reduce the time spent loading the data). It does this by producing indices that map to only a group of image stacks at a time.

Currently the `FifoBatchSampler` also has the responsibility to register the image stacks to the manager and close the remaining open image stacks at the end of iteration. But maybe this can be refactored somewhere else in future.

### New features or files

- `FifoBatchSampler`
- `FifoImageStackManager`
- `ManagedLazyImageStack`
- `create_dataloader`
- `create_dataset` + more specific factory functions
- patch extractor factory functions
- `PatchingStrategy.get_patch_indices` (needed in the FifoBatchSampler)

- `CareamicsDataset.__init__`

### Removed features or files

- `ImageStackLoader` implementations - the image stacks are now directly initialised in the patch extractor factory functions. The `ImageStackLoader` is kept for custom loading.

## How has this been tested?

Added tests (that can probably use some refactoring)

- `test_fifo_batch_sampler` - should test most functionality (not a unit test)
- `test_fifo_image_stack_manager` (also not a unit test)
- test extract patch for the `ManagedLazyImageStack`
- additional `PatchingStrategy` test for `get_patch_indices`

## Related Issues

- Resolves #406 
- Resolves #402

As currently implemented, this system will probably not work properly for `num_workers>0`. This is because the `BatchSampler` and by extension the `FifoImageStackManager` run in main, but if different processes are editing the `loaded_image_stacks` list in the manager at the same time this will cause problems. The safest thing to do would be to create a separate manager for each worker, but I already wasted a lot of time in a rabbit hole on how the PyTorch multiprocess dataloading works so I think I will leave this for now... A warning is raised if lazy loading is selected with `num_workers>0`

Also sorry if this PR is confusing, I let it get a little bit out of hand....

---

- [x] Pre-commit passes


melisande-c:
Closing to split into two PRs.

PR #449, opened 2025-03-28
Refac(Next-Gen Dataset): Dataset initialization

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Refactor of the dataset initialisation. This PR changes the dataset initialisation so that it is given already initialised `PatchExtractor` objects instead having the responsibility to create them itself. Then to make initialisation more convenient, dataset factory functions have been added.

Note: this PR has been created by cherry-picking relevant commits from #448, as it was too big and combined two fairly separate concepts.

The CAREamicsDataset initialisation was refactored so that it no longer controls the construction of the `PatchExtractors` but gets given them as arguments - following the dependency injection pattern.

The advantage of this pattern is that it is easier to know how the dataset has been constructed and what it's underlying types are. This means it is easier to make clearer code paths that avoid `if isinstance` checks to find out what the underlying image stacks are.

Taking away the responsibility of creating the patch extractors from the dataset also means we have more flexibility to change or add construction logic without having to change the class itself.

### Overview - what changed?

Additionally to the dataset class initialisation, I removed all the `ImageStackLoader` implementations and left `ImageStackLoader` to only be used for the custom image stack loading. I created separate patch extractor factory functions instead of having overloads. I did this because having a really long chain of function calls seemed annoying, i.e. `create_dataset` calls `create_patch_extractor` which calls `load_image_stacks`. It also made type hinting easier.

Now there is also a `create_dataset` convenience function, along with all the more specific factory functions. And a `determine_dataset_type` function to switch between them. These types are determined by both how the data is loaded and what the underlying `ImageStack` will be.

To keep track of the underlying types of the dataset I made both the `PatchExtractor` class and the `CAREamicsDataset` into python `Generic` types.

Example:

```python
# we know that the underlying image stack will be the in-memory kind
patch_extractor: PatchExtractor[InMemoryImageStack]

# we know that the underlying image stack will be the Zarr kind
dataset: CAREamicsDataset[ZarrImageStack]
```

### New features or files

- `create_dataset` + more specific factory functions
- patch extractor factory functions

- `CareamicsDataset.__init__`

### Removed features or files

- `ImageStackLoader` implementations - the image stacks are now directly initialised in the patch extractor factory functions. The `ImageStackLoader` is kept for custom loading.

## How has this been tested?

Updated existing tests to use the factory functions (currently only the array factory function and the tiff factory function). Also updated the notebook

---

- [x] Pre-commit passes


src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -181,4 +163,7 @@ def __getitem__(
         else:
             target_data = None
 
-        return input_data, target_data
+        if target_data is not None:
+            return input_data, target_data
+        else:
+            return (input_data,)  # Default collate_fn doesn't work with None
```
jdeschamps: Would it be worth in your opinion having our own collate function or is it better to just have a few `if else` statements?

src/careamics/dataset_ng/dataset/factory.py
```diff
@@ -0,0 +1,404 @@
...
+    create_array_extractor,
+    create_custom_file_extractor,
+    create_custom_image_stack_extractor,
+    create_ome_zarr_extractor,
+    create_tiff_extractor,
+)
+from .dataset import CareamicsDataset, Mode
+
+P = ParamSpec("P")
+GenericImageStack = TypeVar("GenericImageStack", bound=ImageStack, covariant=True)
```
jdeschamps: Shouldn't we define the `GenericImageStack` in one module and import it from there to avoid proliferation? (see the defintion also in the dataset)

src/careamics/dataset_ng/dataset/dataset.py
```diff
@@ -25,6 +21,7 @@
 from careamics.transforms import Compose
 
 P = ParamSpec("P")
```
jdeschamps: Type variable `P` is not used here anymore if I am not mistaken

src/careamics/dataset_ng/dataset/factory.py
```diff
@@ -0,0 +1,404 @@
...
+    NotImplementedError
+        For lazy-loading (`in_memory=False`) of a custom file type.
+    ValueError
+        If the `data_type` is "custom" but both `read_func` and `image_stack_loader` are
+        None.
+    ValueError
+        If the `data_type` is unrecognized.
+    """
+    if data_type == SupportedData.ARRAY:
+        # TODO: ignoring in_memory arg, error if False?
```
jdeschamps, line 83: We can probably just throw a warning `You set in_memory=False, but in_memory is not relevant when loading arrays`, and mention it in the docs.

src/careamics/dataset_ng/dataset/factory.py
```diff
@@ -0,0 +1,404 @@
...
+    CareamicsDataset[InMemoryImageStack]
+        A CAREamicsDataset
+    """
+    input_extractor = create_array_extractor(source=inputs, axes=config.axes)
+    target_extractor: Optional[PatchExtractor[InMemoryImageStack]]
+    if targets is not None:
+        target_extractor = create_array_extractor(source=targets, axes=config.axes)
+    else:
+        target_extractor = None
+    dataset = CareamicsDataset(config, mode, input_extractor, target_extractor)
```
jdeschamps: Not important, but could return directly `CAREamicsDataset` without creating a `dataset` local variable

src/careamics/dataset_ng/patch_extractor/image_stack_loader.py
```diff
@@ -1,29 +1,21 @@
...
 
-from careamics.config.support import SupportedData
-from careamics.file_io.read import ReadFunc
 from careamics.utils import BaseEnum
 
-from .image_stack import ImageStack, InMemoryImageStack, ZarrImageStack
+from .image_stack import ImageStack
 
 P = ParamSpec("P")
+GenericImageStack = TypeVar("GenericImageStack", bound=ImageStack, covariant=True)
```
jdeschamps: Same comment here

src/careamics/dataset_ng/dataset/factory.py
jdeschamps: Actually it is in many more places

jdeschamps (approved):
Looks good! I like the use of generics, I don't find it too bad to have a function with a giant `if else` statement (soon a `match case` I guess) and it elegantly hides all complexity from users.

src/careamics/dataset_ng/dataset/dataset.py
melisande-c: Yeah maybe it is better to keep the `None` for clarity

src/careamics/dataset_ng/dataset/factory.py
```diff
@@ -0,0 +1,404 @@
...
+    data_type = SupportedData(config.data_type)
+    dataset_type = determine_dataset_type(
+        data_type, in_memory, read_func, image_stack_loader
+    )
+    if dataset_type == DatasetType.ARRAY:
+        return create_array_dataset(config, mode, inputs, targets)
+    elif dataset_type == DatasetType.IN_MEM_TIFF:
+        return create_tiff_dataset(config, mode, inputs, targets)
+    # TODO: Lazy tiff
+    elif data_type == DatasetType.IN_MEM_CUSTOM_FILE:
```
veegalinova: Should it be `dataset_type`?

src/careamics/dataset_ng/dataset/factory.py
```diff
@@ -0,0 +1,404 @@
...
+    ARRAY = "array"
+    IN_MEM_TIFF = "in_mem_tiff"
+    LAZY_TIFF = "lazy_tiff"
+    IN_MEM_CUSTOM_FILE = "in_mem_custom_file"
+    OME_ZARR = "ome_zarr"
+    CUSTOM_IMAGE_STACK = "custom_image_stack"
+
+
+# bit of a mess of if-else statements
+def determine_dataset_type(
```
veegalinova, line 45: I think that the `DatasetType` determining logic is a bit bulky and not really necessary. 
For now, I think doing if / else directly on the `data_type` in the `create_dataset` function will be more elegant, as there are not that many options.

veegalinova (approved):
I really like this!!

src/careamics/dataset_ng/dataset/factory.py
melisande-c: Oops! yes it should be, thanks

src/careamics/dataset_ng/dataset/factory.py
melisande-c: So this comes from splitting my PR in two, but I do end up reusing this function, so I'll leave it separate

src/careamics/dataset_ng/dataset/factory.py
melisande-c: ideally users wouldn't ever encounter this and it should all be handled by internal logic (in the lightning modules probably). But good to have a warning anyway

jdeschamps: Within CAREamics, I agree. In the case where we offer the dataset as an independent package (still have this in the back of my head), this may not be so unlikely.

PR #453, opened 2025-04-03
Mean/Std calculation 

CatEek:
## Description

> [!NOTE]  
> **tldr**: 

Changed the way mean/std are calculated to avoid incorrect result in some corner cases

We found that if some cases numpy can incorrectly interpret axes order which leads to incorrect results when mean/std are calculated in a vectorized way

- for loop  added to `src/careamics/dataset/dataset_utils/running_stats.py`

- [ ] Pre-commit passes

### DIFF
--- a/src/careamics/dataset/dataset_utils/running_stats.py
+++ b/src/careamics/dataset/dataset_utils/running_stats.py
@@ -21,9 +21,13 @@ def compute_normalization_stats(image: NDArray) -> tuple[NDArray, NDArray]:
     tuple of (list of floats, list of floats)
         Lists of mean and standard deviation values per channel.
     """
-    # Define the list of axes excluding the channel axis
-    axes = tuple(np.delete(np.arange(image.ndim), 1))
-    return np.mean(image, axis=axes), np.std(image, axis=axes)
+    # Define the lists for storing mean and std values
+    means, stds = [], []
+    # Iterate over the channels dimension and compute mean and std
+    for ax in range(image.shape[1]):
+        means.append(image[:, ax, ...].mean())
+        stds.append(image[:, ax, ...].std())
+    return np.stack(means), np.stack(stds)
 
 
 def update_iterative_stats(


melisande-c:
So it turns out this is a floating point precision problem? @veegalinova is this what was found out in old numpy issues? 

Should we also do `array.mean(dtype=np.float64)` to be safe? In the [numpy docs](https://numpy.org/doc/2.2/reference/generated/numpy.mean.html) it says for int arrays it is by default float64 but for float arrays the default is the same as the original array type. 

Quote from the notes section:
"_Note that for floating-point input, the mean is computed using the same precision the input has. Depending on the input data, this can cause the results to be inaccurate, especially for [float32](https://numpy.org/doc/2.2/reference/arrays.scalars.html#numpy.float32) (see example below). Specifying a higher-precision accumulator using the [dtype](https://numpy.org/doc/2.2/reference/generated/numpy.dtype.html#numpy.dtype) keyword can alleviate this issue._"

CatEek:
`%timeit -n 100 np.mean(arr, axis=axes)
`
67.5 ms ± 1.45 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

`%timeit -n 100 np.mean(arr, axis=axes, dtype=np.float64)
`
68.5 ms ± 971 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

`%timeit -n 100 calc_m(arr)`
66.8 ms ± 1.21 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

same with float64
68.1 ms ± 317 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

We can add float64, but let's not overcomplicate things

melisande-c (approved):
Yeah I've just done a bit of experimenting and this is such a strange problem, it seems to be fine with float32 when not using the `axis` argument,  even for very large arrays, and it even seems to be faster. But yeah I say we can leave it without float64.

but just an example of how weird this is:
Takes ~0.3s
```python
>>> array = np.ones((80000000, 2), dtype=np.float32)
>>> (array[:,0].mean(), array[:,1].mean())
(1.0, 1.0)
```
Takes ~1s, but gives totally wrong answer??
```python
>>> np.ones((80000000, 2), dtype=np.float32).mean(axis=0)
[0.2097152 0.2097152]
```
Takes ~1.4s
```python
>>> np.ones((80000000, 2), dtype=np.float32).mean(axis=0, dtype=np.float64)
[1. 1.]
```

PR #461, opened 2025-04-18
Next-Gen dataset: lightning module

### LINKED ISSUE #438: Next-Gen Dataset: Lightning module
Create a lightning modules that will work with the new dataset. Important for testing the full training pipeline. There will probably have to be changes to both:
-  the datamodule, which will handle creating the correct set up for the dataset according to the config; 
- and, the main lighting module, since outputs of the dataset are changing.

veegalinova:
Note: I'm still not 100% happy with the code quality and have not written any unit tests yet, but the PR is getting quite large and urgent, so here it is! 😅 

## Description

> [!NOTE]  
> **tldr**: This PR introduces a lightning model and datamodule classes to train algorithms with the new NG dataset 

To train with the new dataset implementation, several key changes were necessary in the current implementation of the lightning data module and model. We decided to implement them as separate new files. 
This PR introduces:
 - new `data_module.py` and lightning modules for care and n2v for the NG dataset
 - usage examples (notebooks)
 - configuration updates and minor bugfixes

### Overview - what changed?

The main implementation was copied from the existing modules, which were then refactored and some new features added (detailed below). Several changes were also made to the `data_model` and `optimizer_models` configuration files, as detailed below. The new features and changes to current files organically emerged during the implementation and testing of the notebooks. 

The main bulk of the additions comes from the example notebooks explained in the testing section. They were created to test that the metrics did not change after the dataset's re-implementation. They should eventually be deleted or transferred to `ceramics_examples`. 

### New features or files
- added `dataset_ng` folder to the `lightning` folder
- added `data_module.py` containing a combined data module for training and prediction
- added lightning modules split by algorithm - separate classes for n2v and care, inherited from a base `UnetModule` class with the common functionality. Open to discussion about this decision, of course, but I kinda like how it turned out. 
- `lightning/dataset_ng/unet_module.py`:
   Base lighting module `UNetModule` has a number of new features:
  - metric calculation (currently hardcoded to only PSNR, but in the future we can make it configurable and extendable)
  - saving hyperparameters (useful to compare experiment setups in wandb)
  - a parameter for loading the best trained model checkpoint before running the prediction (can be important for the correct working of CARE)
  - logging learning rate

- `lightning/dataset_ng/datamodel.py`:
  - both training and inference datamodule are now implemented in the same class 
  - revised lightning datamodule hooks usage - the datasets are now created in the setup hook as per [datamodule documentation](https://lightning.ai/docs/pytorch/stable/data/datamodule.html)

- `config/data/data_model.py`
   to increase the training performance, added a new validator for the dataloader parameters, which:
   - automatically determines CPU count and increases `num_workers`
   - sets `pin_memory` to `True` if training on CUDA (speeds up the device transfer)
   
- `config/optimizer_models.py`
  Removed the default `lr` parameter. Reasoning - currently, a convenience function [`create_careamics_module`](https://github.com/CAREamics/careamics/blob/062c6ba517a9bec59928c6b307414fff82ae79ee/src/careamics/lightning/lightning_module.py#L634C5-L634C28) sets the scheduler parameters dict to empty if they are not passed. On the contrary, when instantiating the configuration by itself, the lr parameter will be set to the default. This created an annoying lr discrepancy between the old and the new implementation that was hard to spot. 

- `dataset_ng/legacy_interoperability.py` 
  Changed the return data to match the expected input by the stitching [`convert_outputs`](https://github.com/CAREamics/careamics/blob/488255829a3e95acde4252752b2f793ef950cf71/src/careamics/prediction_utils/prediction_outputs.py#L12) function

- `patching_strategies/random_patching.py`
  Changed patch number calculation - it was calculated based on the total image size, instead of on each dimension separately (how it is in the current patching implementation. This resulted in the number of training and validation patches not matching between the old and new notebooks. 

- `lightning/lightning_module.py` - the current lightning module
  - added learning rate logging

- moved all the demos to a separate folder
- moved `dataset.py` and `factory.py` to the root of `dataset_ng`

## How has this been tested?

I created 3 notebooks to cover some major use-cases:
- [N2V on BSD64](https://github.com/CAREamics/careamics/blob/61cef46fe093707ad4541ec2661d57527dc237d7/src/careamics/dataset_ng/demos/bsd68_demo.ipynb) dataset
- [N2V on mouse nuclei](https://github.com/CAREamics/careamics/blob/61cef46fe093707ad4541ec2661d57527dc237d7/src/careamics/dataset_ng/demos/mouse_nuclei_demo.ipynb) dataset
- [CARE on U2OS](https://github.com/CAREamics/careamics/blob/61cef46fe093707ad4541ec2661d57527dc237d7/src/careamics/dataset_ng/demos/care_U2OS_demo.ipynb) dataset

The resulting metrics match the ones written in the notebooks on the documentation website. 

## Related Issues

- Resolves [#438](https://github.com/CAREamics/careamics/issues/438)

## Additional Notes
"Future work":
- adding the new configuration parameters once the related PR is ready
- fixing several non-critical ruff and mypy errors
- adding the missing documentation
- adding test to critical parts 

Two crucial parts remain to implement to smooth the notebooks' rough edges: automatic train / val splitting and data statistics calculation. 

---

- [ ] Pre-commit passes


src/careamics/config/optimizer_models.py
```diff
@@ -51,9 +51,7 @@ class OptimizerModel(BaseModel):
 
     # Optional parameters, empty dict default value to allow filtering dictionary
     parameters: dict = Field(
-        default={
-            "lr": 1e-4,
-        },
+        default={},
```
melisande-c, line 54: I think I encountered problems with this before as well. Personally I am leaning towards having no defaults in the pydantic classes and setting them all in the convenience functions, obviously not in this PR, idk if it will have any unforeseen problems tho...

src/careamics/dataset_ng/dataset.py
```diff
@@ -102,22 +102,46 @@ def _initialize_patching_strategy(self) -> PatchingStrategy:
...
+        # TODO: if stats are not set then target is present
+        #  the transforms break silently
         if isinstance(self.config, DataConfig):
             target_means = self.config.target_means
             target_stds = self.config.target_stds
-            if target_means is not None and target_stds is not None:
-                target_stats = Stats(target_means, target_stds)
+            target_stats = Stats(target_means, target_stds)
+        else:
+            target_stats = Stats((), ())
```
melisande-c, line 143: could you briefly explain what the problem is here? Is the problem no error is raised if `target` is not `None` but `target_means` is `None`?

src/careamics/dataset_ng/demos/bsd68_demo.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "ax[0].set_title(\"Validation Batch\")\n",
+    "for i in range(8):\n",
+    "    ax[i].imshow(val_batch[0].data[i][0].numpy(), cmap=\"gray\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Traint the model"
```
melisande-c: Typo: Traint -> Train

src/careamics/dataset_ng/patching_strategies/random_patching.py
```diff
@@ -335,4 +335,6 @@ def _calc_n_patches(spatial_shape: Sequence[int], patch_size: Sequence[int]) ->
             f"spatial dimensions {len(spatial_shape)}, for `patch_size={patch_size}` "
             f"and `spatial_shape={spatial_shape}`."
         )
-    return int(np.ceil(np.prod(spatial_shape) / np.prod(patch_size)))
```
melisande-c, line 338: Oops sorry! I guess I got `prod` and `ceil` in the wrong order, hope it didn't take too long to spot

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -0,0 +1,349 @@
...
+class CareamicsDataModule(L.LightningDataModule):
+    def __init__(
+        self,
+        data_config: DataConfig,
+        train_data: Optional[SupportedDataType] = None,
+        train_data_target: Optional[SupportedDataType] = None,
+        val_data: Optional[SupportedDataType] = None,
+        val_data_target: Optional[SupportedDataType] = None,
+        pred_data: Optional[SupportedDataType] = None,
+        pred_data_target: Optional[SupportedDataType] = None,
```
melisande-c: I guess we never need `pred_data_target`? Apart from maybe calculating metrics, but not sure if that should happen in the data module or not

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -0,0 +1,349 @@
...
+                )
+            elif isinstance(input_data, list):
+                input_list, target_list = self._convert_paths_to_pathlib(
+                    input_data, target_data
+                )
+            else:
+                raise ValueError(
+                    f"Unsupported input type for {self.data_type}: {type(input_data)}"
+                )
+            return input_list, target_list
```
melisande-c: There is actually another case for `CUSTOM`, when using a custom `image_stack_loader` the input can actually be `Any` since we don't want to put unnecessary constraints on the user, for example the source could be some sort of reader class like the `pyczi.CziReader`

This is the function that creates the dataset in that case
https://github.com/CAREamics/careamics/blob/488255829a3e95acde4252752b2f793ef950cf71/src/careamics/dataset_ng/dataset/factory.py#L347-L355

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
```diff
@@ -0,0 +1,141 @@
...
+        self._best_checkpoint_loaded = False
+
+        # TODO: how to support metric evaluation better
+        self.metrics = MetricCollection(PeakSignalNoiseRatio())
+
+    def forward(self, x: Any) -> Any:
+        return self.model(x)
+
+    def _log_training_stats(self, loss: Any) -> None:
+        batch_size = self._trainer.datamodule.config.batch_size
```
melisande-c: Using `batch_size` from the config here could potentially be wrong for the last batch. I propose having `batch_size` as an argument to the function and calculating it in `training_step` from the length of the `batch`? Similar for `_log_validation_stats`.

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
```diff
@@ -0,0 +1,141 @@
...
+            loss,
+            on_step=True,
+            on_epoch=True,
+            prog_bar=True,
+            logger=True,
+            batch_size=batch_size,
+        )
+
+        optimizer = self.optimizers()
+        current_lr = optimizer.param_groups[0]["lr"]
```
melisande-c: pylance is complaining to me in my IDE that technically `self.optimizers` can also return a list of optimizers, not v important tho since we will never have more than one optimizer

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if isinstance(algorithm_config, dict):
+            algorithm_config = algorithm_factory(algorithm_config)
+
+        self.config = algorithm_config
+        self.model: nn.Module = model_factory(algorithm_config.model)
+        self.loss_func = loss_factory(algorithm_config.loss)
+
+        self.save_hyperparameters({"algorithm_config": algorithm_config.model_dump()})
+
```
melisande-c: We do have this [`HyperParametersCallback`](https://github.com/CAREamics/careamics/blob/main/src/careamics/lightning/callbacks/hyperparameters_callback.py) which saves the config in the checkpoints and is applied within the `CAREamist`, but if this `save_parameters` method makes it easier to use with wandb maybe we should switch but it doesn't log the full config only the algorithm.

Note can also do `WandbLogger(config=config)` to log the config on wandb

src/careamics/utils/lightning_utils.py
```diff
@@ -35,7 +35,7 @@ def read_csv_logger(experiment_name: str, log_folder: Union[str, Path]) -> dict:
         lines = f.readlines()
 
         for single_line in lines[1:]:
-            epoch, _, train_loss, _, val_loss = single_line.strip().split(",")
+            epoch, _, _, train_loss, _, val_loss = single_line.strip().split(",")
```
melisande-c: Don't have to address now but it would be good to make this more robust to adding more metrics. Using `pandas` would be easiest but if we don't want to add an additional dependency we can simply log the metrics in a dict using the (currently skipped) header as the keys.

i.e.
```python
metrics={}
keys = lines[0].strip().split(",")
for key in keys:
    metrics[key] = []
for single_line in lines[1:]:
    values = single_line.strip().split(",")
    for key, value in zip(keys, values):
        metrics[key].append(value)
```

src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def training_step(
+        self,
+        batch: Union[tuple[ImageRegionData], tuple[ImageRegionData, ImageRegionData]],
+        batch_idx: Any,
+    ) -> Any:
+        x = batch[0]
+        x_masked, x_original, mask = self.n2v_manipulate(x.data)
+        prediction = self.model(x_masked)
+        loss = self.loss_func(prediction, x_original, mask)
```
melisande-c, line 48: Splitting lightning modules per algorithm makes sense to me since for each algorithm functionality is slightly extended or modified. 

However one small thing is that I feel the function signature of the `loss_func` is slightly obscured, (it already was but it feels even more removed here), maybe the initialisation of the loss could be moved to the child classes? and we probably won't need the `loss_factory` function in that case. Don't worry for now if it seems like too much work

melisande-c (approved):
Nice! Notebooks worked for me!

I guess also a fairly important next step is getting the automatic stats calculation to work as before and then we will have pretty much replicated all the previous functionality

src/careamics/dataset_ng/dataset.py
veegalinova: Exactly, no error is raised. Moreover, in this case, after the transforms are called, the target silently becomes an empty array.

src/careamics/lightning/dataset_ng/data_module.py
veegalinova: Yeah, I left it for the case where you'd count metrics. If that happens, the logic should be part of the lightning module, but you'd need targets. I'm not sure if it's an actual use case right now, though.

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
veegalinova: Done! 
Note - we did not have this before because the batch size is calculated automatically. But now the data structure is more complicated, so we need to pass it manually.

src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py
veegalinova: Good idea! I removed the `loss_factory` and the `model_factory` also.

src/careamics/dataset_ng/patching_strategies/random_patching.py
veegalinova: It did not influence the metrics that much, even. I just noticed because the wandb epoch plots were not matching up with the old runs 😆

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: yeah fair enough we can just keep it I guess

src/careamics/dataset_ng/dataset.py
melisande-c: I guess this should be dealt with in a PR to handle the stats calculation?

src/careamics/lightning/dataset_ng/data_module.py
jdeschamps: can we flag it (with a comment) so that it is does not weird us out if we forget about that?

src/careamics/lightning/dataset_ng/data_module.py
veegalinova: I ended up rewriting this file quite a bit to comply with mypy, added an if statement for this case

src/careamics/lightning/dataset_ng/lightning_module.py
```diff
@@ -0,0 +1,207 @@
...
+from careamics.models.model_factory import model_factory
+from careamics.transforms import N2VManipulateTorch
+from careamics.transforms.normalize import Denormalize
+from careamics.utils.logging import get_logger
+from careamics.utils.torch_utils import get_optimizer, get_scheduler
+
+logger = get_logger(__name__)
+
+
+class UNetModule(L.LightningModule):
```
jdeschamps: I probably missed it, but what is this module for? Is it just a lightning module that were meant for all algorithms (superseded by the `lightning_modules` module?), that is just kept for reference?

src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py
```diff
@@ -0,0 +1,54 @@
+from typing import Any, Union
+
+from careamics.config import (
+    N2VAlgorithm,
+)
+from careamics.dataset_ng.dataset import ImageRegionData
+from careamics.lightning.dataset_ng.lightning_modules.unet_module import UnetModule
```
jdeschamps: Not important, but here the import is absolute, while it is relative in the `care_module`.

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
jdeschamps: > We do have this [`HyperParametersCallback`](https://github.com/CAREamics/careamics/blob/main/src/careamics/lightning/callbacks/hyperparameters_callback.py) which saves the config in the checkpoints and is applied within the `CAREamist`, but if this `save_parameters` method makes it easier to use with wandb maybe we should switch but it doesn't log the full config only the algorithm.
>

Yeah it is a bit annoying to have two similar mechanisms. The problem is the same we face a long time ago: the lightning module does not have access to the full configuration.

Please correct me, but if I remember well, these two mechanisms are doing very similar things. The `save_hyperparameters` save the dict in `hparams`, which is exactly what our callback is doing.

> 
> Note can also do `WandbLogger(config=config)` to log the config on wandb

We should figure out when it is saving the configuration exactly. The issue that may arise is that the statistics may be updated after saving the configuration (if it is saved at instantiation of the logger).

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
```diff
@@ -0,0 +1,137 @@
...
+            loss,
+            on_step=False,
+            on_epoch=True,
+            prog_bar=True,
+            logger=True,
+            batch_size=batch_size,
+        )
+        self.log_dict(self.metrics, on_step=False, on_epoch=True, batch_size=batch_size)
+
+    def _load_best_checkpoint(self) -> None:
```
jdeschamps, line 80: Could we override that one in n2v to add a warning to users saying that the "best checkpoint in terms of metrics" might not be the best performing model?

src/careamics/lightning/dataset_ng/lightning_modules/care_module.py
```diff
@@ -0,0 +1,53 @@
...
+from .unet_module import UnetModule
+
+logger = get_logger(__name__)
+
+
+class CAREModule(UnetModule):
+    def __init__(self, algorithm_config: Union[CAREAlgorithm, dict]) -> None:
+        super().__init__(algorithm_config)
+        assert isinstance(
+            algorithm_config, CAREAlgorithm
```
jdeschamps: This should also be compatible with N2N

jdeschamps (approved):

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
melisande-c: So the datamodule can also save hyperparameters in the checkpoint, according to these [docs for `load_from_checkpoint`](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.core.LightningDataModule.html#lightning.pytorch.core.LightningDataModule.load_from_checkpoint) they will be under the key `"datamodule_hyper_parameters"`. This would mean that the config file isn't all saved together in a single file though, and I think we have a few more parameters.

> The issue that may arise is that the statistics may be updated after saving the configuration (if it is saved at instantiation of the logger).

Ah I didn't think of that, maybe the wandb config can also be updated when the stats are updated

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -17,21 +17,22 @@
 
 logger = get_logger(__name__)
 
-
-InputDataType = Union[Path, str, NDArray]
-SupportedDataType = Union[InputDataType, list[InputDataType]]
+ItemType = Union[Path, str, NDArray[Any]]
+InputType = Union[ItemType, list[ItemType], None]
```
melisande-c: This typing still wouldn't allow arbitrary types for when using a custom `image_stack_loader`, like the example I gave before where the source could be a `pyczi.CziReader`.

Maybe to make the typing clear to users we can overload the `__init__` method so that `image_stack_loader` is not optional if the data sources are not `Path`, `str` or `NDArray`.

i.e.
```python

@overload
def __init__(
    self,
    data_config: DataConfig,
    train_data: Optional[Any],
    # etc. ...
    image_stack: ImageStackLoader,
): ...

@overload
def __init__(
    self,
    data_config: DataConfig,
    train_data: Optional[SupportedDataType],
    # etc. ...
    image_stack: Optional[ImageStackLoader],
): ...
```

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -174,64 +135,130 @@ def _convert_paths_to_pathlib(
...
+        elif isinstance(input_data, list):
+            if isinstance(input_data[0], (str, Path)):
+                if target_data is not None:
+                    assert isinstance(target_data, list)
+                input_list, target_list = self._convert_paths_to_pathlib(
+                    input_data, target_data
+                )
+                return input_list, target_list
+        elif self.image_stack_loader is not None:
+            return input_data, target_data
```
melisande-c: The check if a custom `image_stack_loader` should probably be an early return at the start of this method, because a custom input could be a list of arbitrary types and then it would get caught in the "is instance list" if-statement and cause an error.

melisande-c (commented):
Just a couple things that I think should still change for the custom image stack loader option

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
veegalinova: Added a check 👍

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
veegalinova: Removed this call and added `HyperParametersCallback` to the notebooks instead

src/careamics/utils/lightning_utils.py
veegalinova: Good idea, implemented

src/careamics/lightning/dataset_ng/lightning_module.py
veegalinova: Good catch, smh this file escaped commit

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
veegalinova: Added a warning for this case

src/careamics/lightning/dataset_ng/data_module.py
veegalinova: Good suggestion! I rewrote it to 3 overloads - standard init, init with a custom `read_source_func` and init with a custom `image_stack_loader`

src/careamics/lightning/dataset_ng/data_module.py
veegalinova: Moved the check to the top

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -23,16 +23,77 @@
...
+    def __init__(
+        self,
+        data_config: DataConfig,
+        *,
+        train_data: Optional[Any] = None,
+        train_data_target: Optional[Any] = None,
+        val_data: Optional[Any] = None,
+        val_data_target: Optional[Any] = None,
+        pred_data: Optional[Any] = None,
+        pred_data_target: Optional[Any] = None,
```
melisande-c: For the custom read function the sources should be a Path, or str

melisande-c:
Looks good to merge! :)

PR #463, opened 2025-04-29
Docs(Next-Gen Dataset): Add notebook to demonstrate create a custom image stack

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Adding a notebook to demonstrate creating a custom image stack.

This is the first step to creating some documentation for users on how to use the new dataset.

### Overview - what changed?

Added a notebook and also fixed a bug that I discovered when `create_dataset` is called with a custom image stack loader.

The notebook creates an ImageStack for loading from a HDF5 file, the HDF5 api allows you to access slices stored on disk without having to load the whole dataset so I thought it was an appropriate use case.

The image stack itself is extremely similar to the `ZarrImageStack`, with almost the exact same `extract_patch` method.

I provide two example image stack loader functions, for slightly different use cases, to demonstrate the flexibility.

## How has this been tested?

Running the notebook.

- [x] Pre-commit passes


veegalinova:
I rebased the branch to main, removed stats calculation (to be merged in https://github.com/CAREamics/careamics/pull/466), and added training to the notebook

jdeschamps (approved):
Looks great, thanks! Love the idea of making an HD5 example.

I must say I also love the ImageStack/ImageStackLoader design, it's very elegant and powerful.

Here are small typos/grammar things I could pick up, or suggestions:
- In the intro, the links don't work on the Github rendering of the notebook (the root folder is `demo`, it looks there for the relative paths).
- In the intro, at the end, we could add something like this (feel free to improve/correct):
    ```
    <div class="alert alert-block alert-info">
    - `ImageStack` protocol: defines a class responsible for loading data from a specific format.
    - `ImageStackLoader` protocol: defines the signature of a function responsible for resolving how to load the data from a specific format (e.g. find arrays within Zarr groups, or even multiple arrays in multiple Zarr files).
    </div>
    ```
- "The data_shape attribute should be THE shape"
- "The data_dtype attribute is the data type of the underlying array." -> can we give an example in the text (e.g. np.float)? Only if you feel that's useful.
- In `extract_patch`: I find the command bock very difficult to read, it is not clear where sentences end. Maybe something like:
    ```
    
        # Original axes assumed to be any subset of STCZYX (containing YX), in any order.
        # Arguments must be transformed to index data in its original axes order.
        # To do this, we loop through original axes and append correct index/slice, for
        # each axis (STCZYX).
        #  Note: if any axis is not present in original_axes it is skipped.
        #  Note: S and T axes in original data are multiplexed in a single axis in CAREamics.
    ```
- "Now define the image loader": can we specify that "target" only matters if you use a target in the instantiation of the `CAREamicsDataset`? (so as to not be confusing for people that are only thinking N2V)
- "# A image stack loader" -> "# An image stack loader".
- "Now let's run N2V training pipeline and see how it performs" -> it should be CARE
- "# TODO: replace with N2V!!!" -> there's a todo to remove

melisande-c:
After taking a step back and looking at this again, I've realised that we don't need use an image stack loader... When using a custom image stack, users could initialise the image stack instances themselves and pass them to the factory function. This would remove some complexity with type checking in the lightning module as well. The image stack loader protocol was left over from a slightly different approach. I could do this refactor directly in this branch or make a new one

veegalinova:
> After taking a step back and looking at this again, I've realised that we don't need use an image stack loader... When using a custom image stack, users could initialise the image stack instances themselves and pass them to the factory function. This would remove some complexity with type checking in the lightning module as well. The image stack loader protocol was left over from a slightly different approach. I could do this refactor directly in this branch or make a new one

I think it's ok to do it directly here. I'm not quite sure I get your idea, but it's always great to remove complexity where possible, looking forward 😄

melisande-c:
> > After taking a step back and looking at this again, I've realised that we don't need use an image stack loader... When using a custom image stack, users could initialise the image stack instances themselves and pass them to the factory function. This would remove some complexity with type checking in the lightning module as well. The image stack loader protocol was left over from a slightly different approach. I could do this refactor directly in this branch or make a new one
> 
> I think it's ok to do it directly here. I'm not quite sure I get your idea, but it's always great to remove complexity where possible, looking forward 😄

I didn't have time to do this today, I briefly tried but it involves changing some stuff in a few places, so maybe we can leave it as it is for now.

jdeschamps:
Can you add a note in the notebook and open an issue?

If I understood you correctly: rather than having an image stack loader, users can instantiate their image stacks (e.g. different arrays in various zarr files), add them to a list and pass them as is.

But from the point of view of a format (e.g. czi views), the image stack loader still makes sense because the views will always be organized the same way, so an image stack loader can reduce overhead for users, or does it not? If the difference is a few simple lines, it may not indeed.

edit: Oh I guess indeed the image stack loader mechanism can be removed, and then convenience functions (e.g. czi views) can be used externally if they can help users.

melisande-c:
> If I understood you correctly: rather than having an image stack loader, users can instantiate their image stacks (e.g. different arrays in various zarr files), add them to a list and pass them as is.
> 
> edit: Oh I guess indeed the image stack loader mechanism can be removed, and then convenience functions (e.g. czi views) can be used externally if they can help users.

Yeah exactly, I realised if they are writing the loading function anyway, it doesn't matter if it called externally to initialise the image stacks or internally. The image stack loader was left over from when we were trying to unify all the internal loading with the custom loading but we ended up changing that

PR #472, opened 2025-05-22
Refactor(Next-Gen Dataset): Remove ImageStackLoader system

melisande-c:
## Description

> [!NOTE]  
> **tldr**: This PR removes the `ImageStackLoader` protocol in favour of passing custom `ImageStack` classes directly.

The `ImageStackLoader` was first created at the start of the next-gen dataset dataset development when there was an attempt to have a unified loading system for all the image stack types that has now been abandoned. Removing the `ImageStackLoader` gives more flexibility to custom image stack users and reduces the amount of code in CAREamics that needs to be maintained.

This was decided following discussion in #463 .

### Overview - what changed?

Instead of passing an image stack loader function to the lighting datamodule (and dataset factory) a list of image stacks can be passed instead. Users will instead initialise their custom ImageStacks in their own code rather than have it happen within CAREamics code.

Refactored the `create_dataset` factory function. `inputs` and `targets` are type hinted as a union of `Sequence[NDArray]`, `Sequence[Path]` and `Sequence[ImageStack]`. If the input is not a sequence of arrays or paths then it is assumed to be image stacks (one cannot call `isinstance` on python protocols because the objects do not inherit from the protocol class). 

The `_determine_dataset_type` function was removed, this was originally implemented for use with the upcoming lazy tiff but it was complicating things and I will find a different solution.

The `_validate_custom_input` method in the datamodule was modified to handle `ImageStacks` as input.

- `create_dataset`
- `CareamicsDataModule`
- `demo_custom_image_stack` notebook

### Removed features or files

- `src/careamics/dataset_ng/patch_extractor/image_stack_loader.py`
- `demo_custom_image_stack_loader` notebook

## How has this been tested?

Updated and ran the `demo_custom_image_stack` notebook, but something probably still need to be tested.

Some checks and guards in the `create_dataset` and `CAREamicsDataModule` are duplicated, such as checking data type in config matches the input type and checking there is a `read_func` for custom file types. If we remove some of these checks from the `CAREamicsDataModule` we can simplify some of it's methods. I think the main thing that needs to be in the datamodule is to process path inputs to list directory contents and to convert single arrays and `ImageStacks` to lists. And to check that the lists are the same length.

---

- [x] Pre-commit passes


src/careamics/dataset_ng/factory.py
```diff
@@ -26,168 +26,125 @@
...
-        Additional key-word arguments to pass to the `read_func`.
-    image_stack_loader : ImageStackLoader, optional
-        A function for custom image stack loading. This argument is ignored unless the
-        `data_type` in the `config` is "custom".
-    image_stack_loader_kwargs : {str, Any}, optional
-        Additional key-word arguments to pass to the `image_stack_loader`.
+) -> CareamicsDataset:
+    if _is_source_item_type(inputs, np.ndarray):
+        if config.data_type != SupportedData.ARRAY:
+            raise ValueError
```
jdeschamps: Can we add an error message?

src/careamics/dataset_ng/factory.py
```diff
@@ -26,168 +26,125 @@
...
+        if (targets is not None) and (not _is_source_item_type(targets, np.ndarray)):
+            raise TypeError(
+                "Input and target types do not match, inputs are arrays but found "
+                f"target type '{type(targets[0])}'."
+            )
+        return create_array_dataset(config, mode, inputs, targets)
+    elif _is_source_item_type(inputs, Path):
+        if (targets is not None) and (not _is_source_item_type(targets, Path)):
```
jdeschamps, line 101: Shouldn't this also raise an error if the config stipulates type `array`, but we pass on paths?

src/careamics/dataset_ng/factory.py
```diff
@@ -26,168 +26,125 @@
...
+            raise TypeError(
+                "Input and target types do not match, inputs are paths but found "
+                f"target type '{type(targets[0])}'."
+            )
+        return _create_from_paths(
+            config, mode, inputs, targets, in_memory, read_func, read_kwargs
+        )
+    # assume remaining input type option is Sequence[ImageStack]
+    else:
+        if (targets is not None) and (
```
jdeschamps, line 117: and check that data type is custom in configuration?

src/careamics/dataset_ng/factory.py
```diff
@@ -386,23 +340,40 @@ def create_custom_image_stack_dataset(
...
-            targets,
-            config.axes,
-            image_stack_loader,
-            *args,
-            **kwargs,
-        )
+        target_extractor = create_custom_image_stack_extractor(targets, config.axes)
     else:
         target_extractor = None
     dataset = CareamicsDataset(config, mode, input_extractor, target_extractor)
```
jdeschamps: Can return `CareamicsDataset` directly.


Makes me also wonder whether we should not name it `CAREamicsDataset` for coherence or stick to conventions. Not relevant in this PR.

jdeschamps (approved):
It's nice to see the code base shrink a little!

Now, I need to put some thoughts down for myself mainly. The `image_stack_loaders` are now the responsibility of the users who need to provide a list of `ImageStacks` if they want custom loading. There is still the alternative of providing a `read_func`.

For the future integration into the CAREamist, that means that it should also accept `Sequence[ImageStacks]` as an input. We might want to simplify it a little for users by allowing passing a `folder_path` and `image_stack_loader` (any callable that takes a path and returns a `Sequence[ImageStacks]`) and then internally doing the list comprehension.

Regarding configuration, currently the `custom` data type covers both custom cases (`Sequence[ImageStacks`] or `read_func`). We probably don't need to differentiate them since the `input` type allows discriminating.

src/careamics/dataset_ng/factory.py
melisande-c: Whoops I forgot to go back to that

src/careamics/dataset_ng/factory.py
melisande-c: happens is the `_create_from_paths` function, in the else block

melisande-c:
> Now, I need to put some thoughts down for myself mainly. The `image_stack_loaders` are now the responsibility of the users who need to provide a list of `ImageStacks` if they want custom loading. There is still the alternative of providing a `read_func`.

Yes, the way we had it set up meant there was the constraint that additional kwargs had to be shared between train, target, and maybe val and predict images, which might be a bit annoying. There were workarounds, like making a TypedDict source type but it adds complication. This way users can instantiate all image stacks as they please, outside of CAREamics.

> For the future integration into the CAREamist, that means that it should also accept `Sequence[ImageStacks]` as an input. We might want to simplify it a little for users by allowing passing a `folder_path` and `image_stack_loader` (any callable that takes a path and returns a `Sequence[ImageStacks]`) and then internally doing the list comprehension.

Yes correct the CAREamist will have to be able to accept a sequence of image stacks. Not sure about adding back in the `image_stack_loader`. Using a custom `ImageStack` is highest level of customisation and therefore it sort of follows that it has the least level of convenience. In any case, writing an `image_stack_loader` function that has to adhere to our constraints is maybe less convenient than being able to load the image stacks how they please. But we will still keep the `read_func` which is for file formats that don't have chunked/sub file reading functionality.

> Regarding configuration, currently the `custom` data type covers both custom cases (`Sequence[ImageStacks`] or `read_func`). We probably don't need to differentiate them since the `input` type allows discriminating.

We don't need to change it yes since providing a `read_func` with `Path` inputs is how we distiguish the difference, but if we differentiated in the config it might remove some `if-else` blocks in the code

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -71,8 +71,6 @@ def __init__(
         val_data_target: Optional[Any] = None,
         pred_data: Optional[Any] = None,
         pred_data_target: Optional[Any] = None,
-        image_stack_loader: ImageStackLoader,
```
jdeschamps, line 74: Is this overload still necessary? I think it is now the same as the first one.

In my docstring PR, I removed it so we can also ignore it here...

jdeschamps (approved):

melisande-c:
Closing because after some thought the `ImageStackLoader` is a nice feature, I think some of the other factory methods (mostly the for the `PatchExtractor`) can be refactored instead to simplify the code a bit.

PR #474, opened 2025-05-26
Fix: wrong axes in BMZ specs

jdeschamps:
## Description

Following https://github.com/bioimage-io/collection-bioimage-io/issues/750, https://github.com/bioimage-io/collection-bioimage-io/issues/747 and https://github.com/bioimage-io/collection-bioimage-io/issues/749, it appears that we had a bug (or rather a misunderstanding) in the BMZ specs export.

This fixes this issue by correctly specifying `xy` as the normalizing axes. Not sure why the BMZ validation would work with the normalization being wrong...

### DIFF
--- a/src/careamics/model_io/bioimage/model_description.py
+++ b/src/careamics/model_io/bioimage/model_description.py
@@ -160,7 +160,7 @@ def _create_inputs_ouputs(
             preprocessing=[
                 FixedZeroMeanUnitVarianceDescr(
                     kwargs=FixedZeroMeanUnitVarianceAlongAxisKwargs(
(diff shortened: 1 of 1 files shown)


jdeschamps:
Need to debug what's going on with the tests.

[EDIT] Now all our BMZ export fail.

jdeschamps:
The code was actually correct, and there was simply a misinterpretation of the docs!

https://bioimage-io.github.io/spec-bioimage-io/bioimageio/spec/model/v0_5.html#FixedZeroMeanUnitVarianceDescr

PR #476, opened 2025-05-27
Support CZI file format with NextGen datasets

Callidior:
## Description

Adds support for the CZI file format within the NextGen datasets framework.

CZI is the image format used by ZEISS microscopes. Processing CZI files directly without having to convert them to a different format would streamline the denoising workflow of ZEISS microscope users significantly easier.

### Overview - what changed?

- Implement `CziImageStack` for reading patches from an individual scene of a single CZI file.
- Add a `create_czi_extractor` patch extractor factory function for creating a patch extractor for all scenes contained in a list of CZI files.
- Add a `create_czi_dataset` factory (similar to the functions present for other file formats).
- Add `czi` as supported format in enums and docstrings.

Added `CziImageStack` and `create_czi_extractor` factory function.

Regarding the user-facing API, one just needs to specify `data_type="czi"` in the config factory functions, e.g.:

```python
config_2d = create_n2v_configuration(
    experiment_name="n2v_experiment",
    data_type="czi",
    axes="SCYX",
    patch_size=[64, 64],
    batch_size=32,
    num_epochs=100,
    independent_channels=True,
    n_channels=1,
)
config_3d = create_n2v_configuration(
    experiment_name="n2v_experiment",
    data_type="czi",
    axes="SCZYX",
    patch_size=[16, 64, 64],
    batch_size=16,
    num_epochs=100,
    independent_channels=True,
    n_channels=1,
)
```

The `axes`  argument can be one of the following:

- `SCYX`: perform 2-D denoising (also works on time-series and Z-stacks).
- `SCZYX`: perform 3-D denoising. If no Z axis is present, the T (time) axis will be used as 3rd dimension instead.
- `SCTYX`: perform 3-D denoising, enforcing the T (time) axis to be used as 3rd dimension.

### New features or files

- `CziImageStack` added to new file `src/careamics/dataset_ng/image_stack/czi_image_stack.py`
- `create_czi_extractor` added to existing file `src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py`

- Support `"czi"` as data type in `src/careamics/dataset_ng/factory.py`

## How has this been tested?

- New unit test in `tests/dataset_ng/patch_extractor/image_stack/test_czi_image_stack.py` for testing `CziImageStack`. Random data is written to a new CZI file and then read back using `CziImageStack` with different `axes` modes.
- Training and inference with a N2V model on a CZI file ran successfully.

---

- [x] Pre-commit passes

on-behalf-of: @zeiss <bjoern.barz@zeiss.com>

### DIFF
--- a/src/careamics/config/configuration_factories.py
+++ b/src/careamics/config/configuration_factories.py
@@ -199,7 +199,7 @@ def _create_algorithm_configuration(
 
 
 def _create_data_configuration(
-    data_type: Literal["array", "tiff", "custom"],
+    data_type: Literal["array", "tiff", "czi", "custom"],
     axes: str,
     patch_size: list[int],
     batch_size: int,
@@ -212,7 +212,7 @@ def _create_data_configuration(
 
     Parameters
     ----------
-    data_type : {"array", "tiff", "custom"}
+    data_type : {"array", "tiff", "czi", "custom"}
         Type of the data.
     axes : str
         Axes of the data.
@@ -282,7 +282,7 @@ def _create_training_configuration(
 def _create_supervised_config_dict(
     algorithm: Literal["care", "n2n"],
     experiment_name: str,
-    data_type: Literal["array", "tiff", "custom"],
+    data_type: Literal["array", "tiff", "czi", "custom"],
     axes: str,
     patch_size: list[int],
     batch_size: int,
@@ -306,7 +306,7 @@ def _create_supervised_config_dict(
         Algorithm to use.
     experiment_name : str
         Name of the experiment.
-    data_type : Literal["array", "tiff", "custom"]
+    data_type : Literal["array", "tiff", "czi", "custom"]
         Type of the data.
     axes : str
         Axes of the data (e.g. SYX).
@@ -405,7 +405,7 @@ def _create_supervised_config_dict(
 
 def create_care_configuration(
     experiment_name: str,
-    data_type: Literal["array", "tiff", "custom"],
+    data_type: Literal["array", "tiff", "czi", "custom"],
     axes: str,
     patch_size: list[int],
     batch_size: int,
@@ -445,7 +445,7 @@ def create_care_configuration(
     ----------
     experiment_name : str
         Name of the experiment.
-    data_type : Literal["array", "tiff", "custom"]
+    data_type : Literal["array", "tiff", "czi", "custom"]
         Type of the data.
     axes : str
         Axes of the data (e.g. SYX).
@@ -551,6 +551,33 @@ def create_care_configuration(
     ...     n_channels_in=3,
     ...     n_channels_out=1 # if applicable
     ... )
+
+    If you would like to train on CZI files, use `"czi"` as `data_type` and `"SCYX"` as
+    `axes` for 2-D or `"SCZYX"` for 3-D denoising. Note that `"SCYX"` can also be used
+    for 3-D data but spatial context along the Z dimension will then not be taken into
+    account.
+    >>> config_2d = create_care_configuration(
+    ...     experiment_name="care_experiment",
+    ...     data_type="czi",
+    ...     axes="SCYX",
+    ...     patch_size=[64, 64],
+    ...     batch_size=32,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels_in=1,
+    ...     n_channels_out=1,
+    ... )
+    ... config_3d = create_care_configuration(
+    ...     experiment_name="care_experiment",
+    ...     data_type="czi",
+    ...     axes="SCZYX",
+    ...     patch_size=[16, 64, 64],
+    ...     batch_size=16,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels_in=1,
+    ...     n_channels_out=1,
+    ... )
     """
     return Configuration(
         **_create_supervised_config_dict(
@@ -576,7 +603,7 @@ def create_care_configuration(
 
 def create_n2n_configuration(
     experiment_name: str,
-    data_type: Literal["array", "tiff", "custom"],
+    data_type: Literal["array", "tiff", "czi", "custom"],
     axes: str,
     patch_size: list[int],
     batch_size: int,
@@ -616,7 +643,7 @@ def create_n2n_configuration(
     ----------
     experiment_name : str
         Name of the experiment.
-    data_type : Literal["array", "tiff", "custom"]
+    data_type : Literal["array", "tiff", "czi", "custom"]
         Type of the data.
     axes : str
         Axes of the data (e.g. SYX).
@@ -722,6 +749,33 @@ def create_n2n_configuration(
     ...     n_channels_in=3,
     ...     n_channels_out=1 # if applicable
     ... )
+
+    If you would like to train on CZI files, use `"czi"` as `data_type` and `"SCYX"` as
+    `axes` for 2-D or `"SCZYX"` for 3-D denoising. Note that `"SCYX"` can also be used
+    for 3-D data but spatial context along the Z dimension will then not be taken into
+    account.
+    >>> config_2d = create_n2n_configuration(
+    ...     experiment_name="n2n_experiment",
+    ...     data_type="czi",
+    ...     axes="SCYX",
+    ...     patch_size=[64, 64],
+    ...     batch_size=32,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels_in=1,
+    ...     n_channels_out=1,
+    ... )
+    ... config_3d = create_n2n_configuration(
+    ...     experiment_name="n2n_experiment",
+    ...     data_type="czi",
+    ...     axes="SCZYX",
+    ...     patch_size=[16, 64, 64],
+    ...     batch_size=16,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels_in=1,
+    ...     n_channels_out=1,
+    ... )
     """
     return Configuration(
         **_create_supervised_config_dict(
@@ -747,7 +801,7 @@ def create_n2n_configuration(
 
 def create_n2v_configuration(
     experiment_name: str,
-    data_type: Literal["array", "tiff", "custom"],
+    data_type: Literal["array", "tiff", "czi", "custom"],
     axes: str,
     patch_size: list[int],
     batch_size: int,
@@ -810,7 +864,7 @@ def create_n2v_configuration(
     ----------
     experiment_name : str
         Name of the experiment.
-    data_type : Literal["array", "tiff", "custom"]
+    data_type : Literal["array", "tiff", "czi", "custom"]
         Type of the data.
     axes : str
         Axes of the data (e.g. SYX).
@@ -941,6 +995,31 @@ def create_n2v_configuration(
     ...     num_epochs=100,
     ...     independent_channels=False,
     ...     n_channels=3
+    ... )
+
+    If you would like to train on CZI files, use `"czi"` as `data_type` and `"SCYX"` as
+    `axes` for 2-D or `"SCZYX"` for 3-D denoising. Note that `"SCYX"` can also be used
+    for 3-D data but spatial context along the Z dimension will then not be taken into
+    account.
+    >>> config_2d = create_n2v_configuration(
+    ...     experiment_name="n2v_experiment",
+    ...     data_type="czi",
+    ...     axes="SCYX",
+    ...     patch_size=[64, 64],
+    ...     batch_size=32,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels=1,
+    ... )
+    ... config_3d = create_n2v_configuration(
+    ...     experiment_name="n2v_experiment",
+    ...     data_type="czi",
+    ...     axes="SCZYX",
+    ...     patch_size=[16, 64, 64],
+    ...     batch_size=16,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels=1,
     ... )
     """
     # if there are channels, we need to specify their number
--- a/src/careamics/config/configuration_factories.py
+++ b/src/careamics/config/configuration_factories.py
@@ -563,20 +563,16 @@ def create_care_configuration(
     ...     patch_size=[64, 64],
     ...     batch_size=32,
     ...     num_epochs=100,
-    ...     independent_channels=True,
     ...     n_channels_in=1,
-    ...     n_channels_out=1,
     ... )
-    ... config_3d = create_care_configuration(
+    >>> config_3d = create_care_configuration(
     ...     experiment_name="care_experiment",
     ...     data_type="czi",
     ...     axes="SCZYX",
     ...     patch_size=[16, 64, 64],
     ...     batch_size=16,
     ...     num_epochs=100,
-    ...     independent_channels=True,
     ...     n_channels_in=1,
-    ...     n_channels_out=1,
     ... )
     """
     return Configuration(
@@ -761,20 +757,16 @@ def create_n2n_configuration(
     ...     patch_size=[64, 64],
     ...     batch_size=32,
     ...     num_epochs=100,
-    ...     independent_channels=True,
     ...     n_channels_in=1,
-    ...     n_channels_out=1,
     ... )
-    ... config_3d = create_n2n_configuration(
+    >>> config_3d = create_n2n_configuration(
     ...     experiment_name="n2n_experiment",
     ...     data_type="czi",
     ...     axes="SCZYX",
     ...     patch_size=[16, 64, 64],
     ...     batch_size=16,
     ...     num_epochs=100,
-    ...     independent_channels=True,
     ...     n_channels_in=1,
-    ...     n_channels_out=1,
     ... )
     """
     return Configuration(
@@ -1008,17 +1000,15 @@ def create_n2v_configuration(
     ...     patch_size=[64, 64],
     ...     batch_size=32,
     ...     num_epochs=100,
-    ...     independent_channels=True,
     ...     n_channels=1,
     ... )
-    ... config_3d = create_n2v_configuration(
+    >>> config_3d = create_n2v_configuration(
     ...     experiment_name="n2v_experiment",
     ...     data_type="czi",
     ...     axes="SCZYX",
     ...     patch_size=[16, 64, 64],
     ...     batch_size=16,
     ...     num_epochs=100,
-    ...     independent_channels=True,
     ...     n_channels=1,
     ... )
     """
--- a/src/careamics/config/data/data_model.py
+++ b/src/careamics/config/data/data_model.py
@@ -95,9 +95,9 @@ class DataConfig(BaseModel):
     )
 
     # Dataset configuration
-    data_type: Literal["array", "tiff", "custom"]
-    """Type of input data, numpy.ndarray (array) or paths (tiff and custom), as defined
-    in SupportedData."""
+    data_type: Literal["array", "tiff", "czi", "custom"]
+    """Type of input data, numpy.ndarray (array) or paths (tiff, czi, and custom), as
+    defined in SupportedData."""
 
     axes: str
     """Axes of the data, as defined in SupportedAxes."""
--- a/src/careamics/dataset_ng/factory.py
+++ b/src/careamics/dataset_ng/factory.py
@@ -156,7 +156,7 @@ def create_dataset(
     Returns
     -------
     CareamicsDataset[ImageStack]
-        The CAREamicsDataset
+        The CAREamicsDataset.
 
     Raises
     ------
@@ -215,7 +215,7 @@ def create_array_dataset(
     Returns
     -------
     CareamicsDataset[InMemoryImageStack]
-        A CAREamicsDataset
+        A CAREamicsDataset.
     """
     input_extractor = create_array_extractor(source=inputs, axes=config.axes)
     target_extractor: Optional[PatchExtractor[InMemoryImageStack]]
@@ -249,7 +249,7 @@ def create_tiff_dataset(
     Returns
     -------
     CareamicsDataset[InMemoryImageStack]
-        A CAREamicsDataset
+        A CAREamicsDataset.
     """
     input_extractor = create_tiff_extractor(
         source=inputs,
@@ -287,7 +287,7 @@ def create_czi_dataset(
     Returns
     -------
     CareamicsDataset[CziImageStack]
-        A CAREamicsDataset
+        A CAREamicsDataset.
     """
 
     input_extractor = create_czi_extractor(source=inputs, axes=config.axes)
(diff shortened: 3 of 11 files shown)


src/careamics/config/configuration_factories.py
```diff
@@ -551,6 +551,33 @@ def create_care_configuration(
...
+    >>> config_2d = create_care_configuration(
+    ...     experiment_name="care_experiment",
+    ...     data_type="czi",
+    ...     axes="SCYX",
+    ...     patch_size=[64, 64],
+    ...     batch_size=32,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels_in=1,
+    ...     n_channels_out=1,
```
jdeschamps: Are `independent_channels` and `n_channels_out` necessary for all CZI files? Otherwise I would just keep the minimum number of parameters for the example.

src/careamics/config/configuration_factories.py
```diff
@@ -551,6 +551,33 @@ def create_care_configuration(
...
+    ... )
+    ... config_3d = create_care_configuration(
```
jdeschamps: The doctests are failing:

```
FAILED src/careamics/config/configuration_factories.py::line:559,column:1
FAILED src/careamics/config/configuration_factories.py::line:757,column:1
FAILED src/careamics/config/configuration_factories.py::line:1004,column:1
```

I can't test it right now, but I surmise it is the missing `>>>` before the second function call!

src/careamics/config/configuration_factories.py
```diff
@@ -722,6 +749,33 @@ def create_n2n_configuration(
...
+    ...     data_type="czi",
+    ...     axes="SCYX",
+    ...     patch_size=[64, 64],
+    ...     batch_size=32,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels_in=1,
+    ...     n_channels_out=1,
+    ... )
+    ... config_3d = create_n2n_configuration(
```
jdeschamps: Same missing `>>>` here probably causing doctest failure.

src/careamics/config/configuration_factories.py
```diff
@@ -942,6 +996,31 @@ def create_n2v_configuration(
...
+    ...     experiment_name="n2v_experiment",
+    ...     data_type="czi",
+    ...     axes="SCYX",
+    ...     patch_size=[64, 64],
+    ...     batch_size=32,
+    ...     num_epochs=100,
+    ...     independent_channels=True,
+    ...     n_channels=1,
+    ... )
+    ... config_3d = create_n2v_configuration(
```
jdeschamps: Same missing `>>>` here probably causing doctest failure.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
```diff
@@ -0,0 +1,297 @@
...
+        Path to the CZI file without scene index.
+    scene : int or None
+        Index of the scene to extract, or None if not specified.
+    data_shape : Sequence[int]
+        The shape of the data in the order `(SC(Z)YX)`.
+    axes : str
+        The axes in the CZI file corresponding to the dimensions in `data_shape`.
+        Possible axes are `C`, `T`, `Z`, `Y`, `X`, and `S`. The axis `S` (sample)
+        is special and combines all remaining axes present in the data.
+    """
```
jdeschamps, line 92: Could you provide examples of the use of the ImageStack, especially regarding to the `path@scene` mechanism?

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
```diff
@@ -0,0 +1,297 @@
...
+
+        # Check for scene encoded in filename.
+        # Normally, file path and scene should be provided as separate arguments but
+        # we would also like to support using the `source` property to re-create the
+        # CZI image stack. In this case, the scene index is encoded in the file path.
+        scene_matches = re.match(r"^(.*)@(\d+)$", _data_path.name)
+        if scene_matches:
+            if scene is not None:
+                raise ValueError(
+                    "Scene index is specified in the filename and as an argument. "
```
jdeschamps:
```suggestion
                    f"Scene index is specified in the filename ({_data_path.name}) and as an argument ({scene}). "
```

src/careamics/dataset_ng/factory.py
```diff
@@ -257,6 +264,42 @@ def create_tiff_dataset(
...
+        Whether to create the dataset in "training", "validation" or "predicting" mode.
+    inputs : Any
+        The input sources to the dataset.
+    targets : Any, optional
+        The target sources to the dataset.
+
+    Returns
+    -------
+    CareamicsDataset[CziImageStack]
+        A CAREamicsDataset
```
jdeschamps:
```suggestion
        A CAREamicsDataset.
```

jdeschamps: (yeah that's a bit ridiculous, sorry about that)

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
```diff
@@ -0,0 +1,297 @@
...
+
+import numpy as np
+from numpy.typing import NDArray
+from pylibCZIrw.czi import CziReader, Rectangle, open_czi
+
+
+class CziImageStack:
+    """
+    A class for extracting patches from an image stack that is stored as a CZI file.
+
```
jdeschamps, line 14: Could you add a few words on the use of the bounding rectangle? It may not makes sense, but if it is defined then it will be used, otherwise the whole range is used?

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
```diff
@@ -0,0 +1,297 @@
...
+        Path to the CZI file, including the scene index if specified.
+    data_path : Path
+        Path to the CZI file without scene index.
+    scene : int or None
+        Index of the scene to extract, or None if not specified.
+    data_shape : Sequence[int]
+        The shape of the data in the order `(SC(Z)YX)`.
+    axes : str
+        The axes in the CZI file corresponding to the dimensions in `data_shape`.
+        Possible axes are `C`, `T`, `Z`, `Y`, `X`, and `S`. The axis `S` (sample)
```
jdeschamps: Can "C" be absent? The example in the configuration factory stated otherwise.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
```diff
@@ -0,0 +1,297 @@
...
+        all scenes present in the CZI file will be extracted.
+        The scene can also be provided as part of `data_path` by appending an `"@"`
+        followed by the scene index to the filename.
+
+    depth_axis : {"none", "Z", "T", "auto"}, default: "auto"
+        Which axis to use as depth-axis for providing 3-D patches.
+
+        - `"none"`: Only provide 2-D patches. If a Z or T dimension is present in the
+          data, they will be combined into the sample dimension `S`.
+        - `"Z"`: Use the Z-axis as depth-axis. If a T axis is present as well, it will
```
jdeschamps, line 46: Are both `auto` and `Z` necessary? `auto` would always select `Z` if it is present, so it would be equivalent.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
```diff
@@ -95,6 +96,48 @@ def create_ome_zarr_extractor(
...
+
+    Returns
+    -------
+    PatchExtractor
+    """
+    depth_axis: Literal["none", "Z", "T", "auto"] = "none"
+    if axes.endswith("TYX"):
+        depth_axis = "T"
+    elif "Z" in axes:
+        depth_axis = "auto"
```
jdeschamps: Joining the previous comment about `depth_axis`, "Z" is not used here, so it may not be necessary?

src/careamics/config/data/data_model.py
```diff
@@ -95,9 +95,9 @@ class DataConfig(BaseModel):
...
 
     # Dataset configuration
-    data_type: Literal["array", "tiff", "custom"]
-    """Type of input data, numpy.ndarray (array) or paths (tiff and custom), as defined
-    in SupportedData."""
+    data_type: Literal["array", "tiff", "czi", "custom"]
+    """Type of input data, numpy.ndarray (array) or paths (tiff, czi, and custom), as
+    defined in SupportedData."""
 
     axes: str
```
jdeschamps, line 102: If "SC" axes are mandatory, then we probably need a model validation here to check that the axes are compatible if data type is `czi`.

jdeschamps (changes requested):
Looks very good, thanks!!

Very minor changes requested, mostly to fix the examples in the docstring. Since I am really not familiar with CZI, there are some question that arose (e.g., axes).

Not a comment, but just a warning that some refactoring to the code in this PR will happen in https://github.com/CAREamics/careamics/pull/472 (after merging in `main` obviously).

src/careamics/config/configuration_factories.py
Callidior: They aren't necessary. I removed them from the example.

src/careamics/config/configuration_factories.py
Callidior: Added it, thanks for the hint!

src/careamics/config/configuration_factories.py
Callidior: Fixed

src/careamics/config/configuration_factories.py
Callidior: Fixed

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: I added examples.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: Done

src/careamics/dataset_ng/factory.py
Callidior: I also fixed it in the docstrings of all the other factory functions in this file, where this was copy-pasted from :)

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: I'm not sure what you mean since bounding rectangle is not mentioned in this docstring. I elaborated the documentation of the `scene`  argument and hope that clarifies it.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: It cannot. I added a list of all possible combinations. However, the user does not have to specify this axis order directly. It is constructed from the value of `depth_axis`.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: `auto` can also select `T` if `T` is present but `Z` is not. It can also select `None` if the data is 2-D. Setting `Z` explicitly as `depth_axis` would raise an error in these cases.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
Callidior: Here I wanted to clarify with you: Does CAREamics know the concept of a T axis and can use it in-place of Z?
`Z` is missing here because I thought `T` is not used elsewhere in CAREamics and wouldn't be allowed in the `axes`  string, so I used `Z` as an alias:

- If the data has a Z axis and the user specifies `axes="SCZYX"`, the Z axis will be used as 3rd dimension.
- If the data does *not* have a Z axis and the user specifies `axes="SCZYX"`, we assume they want to perform 3-D denoising using the time axis as 3rd dimension and translate from `SCZYX` to `SCTYX` when instantiating the image stack.

However, if it's okay that the user simply specifies `axes="SCTYX"`, we can get rid of all this ambiguity and the `auto` value for `depth_axis` - the user could always make it explicit then.

src/careamics/config/data/data_model.py
Callidior: The `CziImageExtractor` will always have a data shape of `(samples, channels, (Z), Y, X)`, which I thought is fixed by the protocol.
However, it is not mandatory for the user to specify `S` and `C` in the `axes`  string. The `create_czi_extractor` factory will only check for the presence of `T` or `Z` to set the `depth_axis` argument accordingly.
Thus, `axes="ZYX"` would have the same effect as `axes="CZYX"` and `axes="SCZYX"`. Alas, I was a bit uncertain whether this user-specified value would also be used at other places in CAREamics besides for creating the image stack, so I kept the `axes` specification in the example in line with that the data shape will actually be.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
jdeschamps: I was mostly curious about the metadata extracted from the CZI file, and whether its use was something that is relevant for users to know. Do CZI users sometimes define it or is it automated metadata generated by the format.

I should have been a bit more explicit, this was really just lack of knowledge from my side and curiosity about the implications for users.

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
jdeschamps: You are correct, we don't have any use of "T" axis. It is up to the image stack to decide what to do if there is such an axis, typically mix it with "S" dimension. The way we asked users to use another axis as "Z" is to simply label it as "Z" (since it is used that way, it is philosophically a Z axis ^^). 

My comment was linked to my other comment regarding `depth_axis` values: is "depth_axis=Z" useful here since:

- Value "depth_axis=Z" is not reachable via `patch_extractor`
- Value `depth_axis=auto` and `depth_axis=Z` have the same consequence in the presence of `Z` axis (with `auto` having the additional benefit of selecting `T` if `T` is present)

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
jdeschamps: See the other discussion, as the two comments were linked!

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
jdeschamps: (this discussion actually relates mainly to the parameters in `czi_image_stack`)

src/careamics/config/data/data_model.py
jdeschamps: Let me just describe a little bit the logic in CAREamics, it is probably clear to you already but that might still make the discussion easier!

The axes specified by users should be the real axes present in the data (but a subset of `STCZYX`, in whichever order, except if I remember correctly that `Y` and `X` need be contiguous) and are used in different ways:
- configuration: 
    - check compatibility with network for 2D/3D (so `Z` must be specified to get 3D)
    - check compatibility with requested patch size (same, `Z` must be specified for 3D patches)
- convenience functions (e.g., `create_n2v_configuration`):
    - force users to specify input channels if `C` is present
- `ImageStacks`: 
    - allow the `ImageStacks` to understand how to index the patches in the data

So, if users have for instance a numpy array (or any format) with axes order `YXCZ`, then they should set `axes=YXCZ` in the configuration. In turn, `axes` will be propagated all the way to the `ImageStack` representing the array.

CAREamics patches are always `SC(Z)YX` (`SCZYX` in our example) with singleton dimensions potentially for `S` (batch) and `C`. The dataset requests from the `ImageStack` patches at coordinates index in the space `SCZYX`. It is the responsibility of the `ImageStack` to translate the coordinates from `SCZYX` to `CYXZ`, and to return the correct patch.

In the end, if `S` and `C` are not mandatory due to the format, then they should not be assumed to be there by default by the `ImageStack`. If it is a CZI specificity (that there is always `S` and `C`), then for coherence with the other formats and the logic of the codebase, I would also enforce that users have to specify it in the configuration (since it is always there).

What type of axes can we expect in a CZI file? e.g., the axes that would end up in `S`.

And what does a scene represent again?

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: For the average user, it is not relevant to know about the bounding rectangles.
CZI stores multiple scenes alongside each other in one large image plane, each in its own bounding rectangle. If someone would use pylibCZIrw to write their own CZI files, they could define these bounding rectangles themselves. But most commonly, these files are written by microscopy software and the bounding rectangles are not user-defined. One example is scanning several wells on a multi-well plate. The entire image would cover the whole plate with all the wells on it. Each scene would correspond to one well and its bounding rectangle would be the region of that well in the image of the plate.
But as I said, the user typically doesn't need to care about that. In the microscopy software, they simply select the scene they want to see and the corresponding region is then read from the file.

In `CziImageStack`, we do need to know the bounding rectangles internally, since we need to extract sub-patches of the scenes and hence need to add the left and top side of the scene's bounding rectangle as an offset to the region of interest we are extracting from the full plane. But the user would not need to deal with it. They only need to specify the index of the scene to extract from the file.

src/careamics/config/data/data_model.py
Callidior: The CZI format doesn't have a specific storage order for axes except X and Y. Think of it as a loose collection of XY planes, each tagged with additional axes coordinates as metadata.
The additional axes `C`, `Z`, and `T` will always be there, even if they are singleton axes.
There can be more axes, e.g., a phase axis (called `H`, I think) for images captured with structured illumination patterns with different phases. Here, we are combining all additional axes into the sample dimension `S`, including `T` or `Z` if they are not mentioned in the `axes`  string.

Scenes, on the other hand, are independent acquisitions stored in the same file. They can have different spatial shapes, which makes them different from just another axis. That's why we create one `CziImageStack` per scene instead of merging them into the sample dimension.

Because there is no inherent "order" of axes and the CZI metadata clearly defines which axes are present, I don't see why the user should be need to specify exactly that again as an argument. That would require the user to first determine for each CZI file, what kind of axes are present, just to provide some information to CAREamics that CAREamics already determines internally. Additionally, it is possible to train on multiple CZI files with different sets of axes. Some could have a `T` axes, some not. We would just all consider them as "samples".
Thus, the only thing we are currently using the `axes`  string for is detecting whether the user wants to train on 2-D or 3-D patches and, in the latter case, whether `Z`  or `T` should be used as 3rd dimension.
However, we do require that `YX`, `ZYX`, or `TYX` are at the end of the string. That is something that could be validated.

Would you prefer if the user just had to specify the spatial axes, i.e., `"YX"`, `"ZYX"`, or `"TYX"`, or should `"SC"` also always be part of the `axes`  string?

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
Callidior: I agree the current handling was confusing and to some degree nontransparent.

I now removed the `"auto"` option for `"depth_axis"`. The new default is `"none"`, yielding 2-D patches. 3-D patches must be requested explicitly by either setting `depth_axis="Z"` or `depth_axis="T"`.

In terms of the `create_czi_extractor` factory, that means that `axes="ZYX"` and `axes="TYX"` will yield 3-D patches, everything else 2-D patches.

I think this is cleaner.

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
Callidior: @jdeschamps Are there some open points left here?

src/careamics/dataset_ng/patch_extractor/image_stack/czi_image_stack.py
jdeschamps: Very clear, thanks!

src/careamics/config/data/data_model.py
jdeschamps: Alright, so basically right now there is a bit of an incompatibility between the CZI format, which knows where axes are and does not need to transform between dimension bases, and the way we thought about CAREamics, with T not being used and axes being potentially randomly ordered.

While the Lightning API allows users to have full control, we still need to maintain access via the main API (`CAREamist` class), and that one is regulated by our Pydantic configuration. Currently, we feel that would create incoherence in the way CZI would be used and the way other supported formats would be.

We discussed it and a potential solution would be:
- adding `depth_axis` to the configuration (where `depth_axis: Literal["T", "Z"] | None`)
- adding a configuration validation that `depth_axis` can only be `T` or `Z` if these axes are in `axes`
- `depth_axis` would be used to validate patching dimensions
- users would then pass in `axes` the actual axis in their data as is currently done, but in the case of CZI there is no use of the axes

Notes:
- if there is `C` (not singleton) in CZI, users need to specify it in `axes` as it used to check for the UNet configuration

What does that mean for this PR?

- Probably you just need to receive `depth_axis` in the `patch_extractor`
- I think it is probably easier to leave the PR as is, and then we try on our side to fix the coherence in the configuration and refactor the `patch_extractor` if necessary.

@melisande-c any additional opinion after our discussion?

melisande-c: I'll just clarify that this would mean a change in behaviour in all of CAREamics, whether a 3D model should be used would no longer be inferred by `Z` being present in the axes but by a new `depth_axis` parameter (which can be set to `None` for 2D). For example, one could have the axes `SZYX` but set the `depth_axis` to `None` and this would mean that a 2D model would be trained, with both `S` and `Z` folded into the sample dimension.

For the CZI case, the `axes` parameter probably won't carry very much information anymore, but maybe we can discuss this further incase specifying a subset of the axes should result in different behaviour than the full `STC(Z)YX`.

Also, at the end of Joran's comment he means the `create_czi_patch_extractor` function not `patch_extractor`.

jdeschamps:
Additional note: the py39 CI are failing due to the use of `str | Path`. We wanted to drop py39 soon anyway, so there is now a PR for that and we can ignore the failing CIs here.

src/careamics/config/data/data_model.py
Callidior: If the solution via the `depth_axis` argument doesn't fit well into CAREamics, I am also open to other ways of solving this. All we need is a way to specify which axis should be used as 3rd one, if any.
Introducing `depth_axis`  into the CAREamics configuration in addition to `axes` would probably make the user-facing API more complex and less intuitive. I would also be fine with simply using the `axes` argument, that already exists, but if would not be allowed to take arbitrary values for CZI. The only valid options would be `SCYX`, `SCZYX`, and `SCTYX`. I think that should be acceptable for users if it is clearly documented and validated.
What do you think?

jdeschamps (approved):

src/careamics/config/data/data_model.py
jdeschamps: Let's keep your PR like this for now and we'll play a little bit with the parameters in the configuration to see what impact it has on the rest of the code.

PR #479, opened 2025-05-27
MS rebuttal changes

CatEek:
Same as the previous one


jdeschamps:
There is conflict with main and I am under the impression that some of the commits I have seen before in the previous PR. Is that the case? Can you rebase to main?

CatEek:
Because

PR #484, opened 2025-05-30
fix: Fix pre-commit by adding missing docstrings

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Pre-commit currently fails because of missing dosctrings, this PR adds the missing docs.

> [!IMPORTANT]  
> - pre-commit will still fail because of a `mypy` error that will be fixed in https://github.com/CAREamics/careamics/pull/472.
> - I excluded `data_module.py` from numpydoc because it tries to enforce documenting the overloads, which is incompatible with ruff.
> - I might not have adhered strictly at our conventions and the docstrings will have to be reviewed once we document the dataset on the website. We will anyway need to do a complete review of the docstrings because the formatting is failing at times on the website.

Missing docstrings NG dataset files are causing pre-commit to fail on all PRs.

### Overview - what changed?

Added docstrings.

- `src.careamics.lightning.dataset_ng.*`

## How has this been tested?

N/A

## Breaking changes

Potential conflicts with https://github.com/CAREamics/careamics/pull/472.

---

- [x] Pre-commit passes


melisande-c (approved):
Thanks for doing this! But yeah we should do a full review of all the docs soon

PR #485, opened 2025-06-03
chore: Drop support for 3.9

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Update requirements and python syntax to >= py3.10, dropping py3.9.

[Python 3.9](https://devguide.python.org/versions/) is being dropped this year, and it is time to drop 3.9. This PR updates the syntax of the code base via `ruff`, and remove the 3.9 from various places (e.g., BMZ, CI, metadata).

3.9 is being dropped and won't be supported in the future.

### Overview - what changed?

Mostly requirements in CI, metadata, and code syntax throughout the codebase.

I updated the requirements in the pre-commit configuration and pyproject, then ran `ruff` to automatically update the syntax in the code. I then fixed manually the various `(str, Path)` -> `str | Path` occurrences. 

Finally, I removed 3.9 CIs.

### New features or files

Changes have happened throughout the code thanks to `ruff`.

## How has this been tested?

Running tests via CI here.

## Related Issues

Note that it will allow https://github.com/CAREamics/careamics/pull/476 to have the CI pass, and will interact with all other open PRs because the changes are many.

- [x] Pre-commit passes


src/careamics/careamist.py
```diff
@@ -827,7 +828,7 @@ def predict_to_disk(
             source_path = source.pred_data
             source_data_type = source.data_type
             extension_filter = source.extension_filter
-        elif isinstance(source, (str, Path)):
+        elif isinstance(source, str | Path):
```
melisande-c, line 831: I didn't realise you could pass union types to `isinstance`, but I guess you can

src/careamics/dataset/patching/patching.py
```diff
@@ -89,7 +90,7 @@ def prepare_patches_supervised(
     """
     means, stds, num_samples = 0, 0, 0
     all_patches, all_targets = [], []
-    for train_filename, target_filename in zip(train_files, target_files):
+    for train_filename, target_filename in zip(train_files, target_files, strict=False):
```
melisande-c, line 93: The default for the new `zip` argument `strict` is already `False`

melisande-c (approved):
Looks good, my comments are unimportant, feel free to ignore. I guess we can slowly replace all the cases of `Union[A, B]` with `A | B` as we go along.

src/careamics/careamist.py
jdeschamps: I was also surprised! I think ruff was enforcing that.

src/careamics/dataset/patching/patching.py
jdeschamps: I don't remember changing that, must have come from the `pre-commit`. Will remove and check if it reappears.

src/careamics/careamist.py
melisande-c: hmm interesting

src/careamics/dataset/patching/patching.py
melisande-c: not important anyway

jdeschamps: Screenshot 2025-06-17 at 00 06 01

Indeed, it is `ruff`.

jdeschamps:
So here's what's new:

- Torch `>2.4` does not support intel macOS, so we are probably going to ditch it soon as well (for isntance that is currently preventing us from upgrading to numpy 2)
- Github now supports silicon runners (`macos-latest`)
- Although the runner are silicon macOS, `MPS` will fail

I have tried:
- Forcing installation of torch CPU versions (but failed)
- Disabling MPS by enforcing `device=cpu`, but I get segmentation faults

Therefore I settled for:
- Splitting CI between macOS (silicon) and linux/windows
- `pytest.mark` all tests that fail on MPS
- Only running the non MPS-related failing tests in the macOS CI

In the future, MPS might be supported and the `pytest.mark` will be removed and the CI fused again. In the mean time, some tests are not run on the CI for macos, but I think we can live with that.

melisande-c:
> I have tried:
> 
> * Forcing installation of torch CPU versions (but failed)
> * Disabling MPS by enforcing `device=cpu`, but I get segmentation faults

Segfaults when setting device to cpu is really weird but splitting the CI makes sense

jdeschamps:
> > I have tried:
> > 
> > * Forcing installation of torch CPU versions (but failed)
> > * Disabling MPS by enforcing `device=cpu`, but I get segmentation faults
> 
> Segfaults when setting device to cpu is really weird but splitting the CI makes sense

Yes, I should probably have investigated more... But it required making `device` a `CAREamist` parameter and using this throughout 50-ish tests. Which is a lot of changes to reverse once MPS works. Maybe we can revisit later.

For completeness, here is the error:

```
/Users/runner/work/_temp/808ee2d2-db9b-4766-94bc-cce739964dee.sh: line 1: 11606 Segmentation fault: 11  pytest --color=yes -m "not lvae"
```

And it was generated [here](https://github.com/CAREamics/careamics/actions/runs/15849066666/job/44677852514).

PR #486, opened 2025-06-04
Refac(Next-Gen): Fold patching strategies into a subconfiguration of DataConfig

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Unify `InferenceConfig` and `DataConfig` into a single configuration for the Next-Gen Dataset, by moving all patching strategy related parameters in the configuration into a sub-configuration.

CAREamics rely on two different data configuration: one for training/validation, the other one for inference. This was not an issue when both steps were handled by different datasets. The new Next-Gen Dataset will handle all datasets in CAREamics, and the code would be greatly simplified if both steps were also handled by a single data configuration.

This PR offers a potential solution to handling both train/val and test in a single data configuration.

### Overview - what changed?

A new `NGDataConfig` class has been added that handles all the parameters necessary for the various patching strategy and dataloader parameters. The subconfiguration `PatchingModel` holds the parameters related to the patching strategies.

The `dataset` and `data_module` have been updated to only take a single data config as input.

Here, the previous `patch_size` parameter has been folded into a subconfiguration for the strategies, and also corresponds to the `tile_size`. The patching strategy is chosen by setting the `name` parameter used as discriminator.

Each patching strategy has a Pydantic model, with some validation if applicable. I chose inheritance to avoid duplicating code (e.g. validation of parameters). 

### New features or files

- New `NGDataConfig` class, with `seed`, `patching_strategy` and `test_dataloader_params` new parameters
- Various patching strategy have their own Pydantic model

- Dataset and data module now take `NGDataConfig` as sole configuration.

## How has this been tested?

Tests locally, ran the local NG Dataset notebooks and NG Dataset BSD68.

## Related Issues

- Resolves #

Note that an additional PR would be necessary to integrate this `NGDatasetConfig` into the main configuration. I made the choice to not introduce breaking change for the rest of the code.

---

- [x] Pre-commit passes


src/careamics/config/data/ng_data_model.py
```diff
@@ -0,0 +1,373 @@
...
+
+    # Dataset configuration
+    # TODO what becomes of custom? is custom passing imagestacks?
+    data_type: Literal["array", "tiff", "zarr", "custom"]
+    """Type of input data."""
+
+    axes: str
+    """Axes of the data, as defined in SupportedAxes."""
+
+    patching: Union[PatchingStrategies] = Field(..., discriminator="name")
```
melisande-c: `PatchingStrategies` is already a Union

```suggestion
    patching: PatchingStrategies = Field(..., discriminator="name")
```

src/careamics/dataset_ng/dataset.py
```diff
@@ -65,33 +68,43 @@ def __init__(
...
-                patch_size=self.config.patch_size,
-                # TODO: Add random seed to dataconfig
-                seed=getattr(self.config, "random_seed", 42),
+                patch_size=self.config.patching.patch_size,
+                seed=self.config.seed,
             )
         elif self.mode == Mode.VALIDATING:
-            if isinstance(self.config, InferenceConfig):
-                raise ValueError("Inference config cannot be used for validating.")
+            if self.config.patching.name != SupportedPatchingStrategy.RANDOM:
```
melisande-c, line 83: Should this be `FIXED_RANDOM`

src/careamics/config/data/patching_strategies/overlapping_patched_model.py
```diff
@@ -0,0 +1,103 @@
...
+from .patched_model import PatchedModel
+
+
+class OverlappingPatchedModel(PatchedModel):
+    """Overlapping patching Pydantic model.
+
+    Attributes
+    ----------
+    name : "overlapping"
+        The name of the patching strategy.
```
melisande-c: There is actually no name attribute (it is inherited but not set which makes sense since this class is just a class to inherit from).

melisande-c: Maybe there could be a comment somewhere mentioning this class is meant to be inherited from

src/careamics/config/data/ng_data_model.py
```diff
@@ -0,0 +1,373 @@
...
+
+# TODO: Validate the specific sizes of tiles and overlaps given UNet constraints
+#   - needs to be done in the Configuration
+#   - patches and overlaps sizes must also be checked against dimensionality
+
+
+# TODO: validation axes vs patch sizes
+# TODO: is 3D updated anywhere in the code?
+# TODO: can we tell when it is made for training but has the wrong patching?
+#       - or is the responsibility of the creator?
```
melisande-c: I guess we can't know unless there is a `mode` parameter. Which could work since the `CAREamicsDataset` takes a mode argument

melisande-c: In which case, I guess we wouldn't need the duplication of the checks in the `__init__` of the CAREamicsDataset since it should be already validated in the config. Maybe it would clean things up a bit

src/careamics/config/data/ng_data_model.py
```diff
@@ -0,0 +1,373 @@
...
+    WholePatchingModel,
+)
+
+# TODO: Validate the specific sizes of tiles and overlaps given UNet constraints
+#   - needs to be done in the Configuration
+#   - patches and overlaps sizes must also be checked against dimensionality
+
+
+# TODO: validation axes vs patch sizes
+# TODO: is 3D updated anywhere in the code?
```
melisande-c: Have you tested training with 3D?

melisande-c (approved):
Looks good! Some small comments :)

src/careamics/config/data/ng_data_model.py
jdeschamps: Good catch, thanks!

src/careamics/dataset_ng/dataset.py
jdeschamps: oh right!

jdeschamps: Ah wait, I know why it is like that. it is not super logical but:

- Only one patching strategy is defined for training currently (in the config) and it is `random`.
- Validation patching strategy cannot be chosen.
- Since validation only happens during training, only `random` should be accepted there and it ensures that we get the correct `patch_size` parameter.

I need to have a closer look to see if there's an alternative to this counter-intuitive piece of code.

src/careamics/config/data/patching_strategies/overlapping_patched_model.py
jdeschamps: I changed the classes that were inherited to private classes (`_OverlappingPatchedModel`). Let me know what you think!

src/careamics/config/data/ng_data_model.py
jdeschamps: No, but I left this TODO for when we will make the NG dataset the default in CAREamics, since this method could be called.

Currently since it is separate and there is no call to the `set_3D` method, I preferred not setting it.

src/careamics/config/data/ng_data_model.py
jdeschamps: The problem is that the `mode` parameter does not really belong in the configuration and it would be odd to create a separate `training` and `validation` configurations. These are consequences of calls from Lightning ([see here](https://github.com/CAREamics/careamics/blob/0cfbc6fc0bc90c7cccd37c8a4a96b088559d85d0/src/careamics/lightning/dataset_ng/data_module.py#L552)).

For now I'd prefer to leave it as it is. The convenience functions for the n2v training configuration or for inference should make sure to only set coherent patching, and otherwise we get an error in the dataset.

melisande-c: ah that's true, one other option could be to have `training_patching_strategy`, `prediction_patching_strategy` (and maybe also for val) but then the config is getting pretty bloated

src/careamics/config/data/patching_strategies/overlapping_patched_model.py
melisande-c: Ah nice, yeah I think that makes it clear!

src/careamics/config/data/ng_data_model.py
jdeschamps: Yeah I'd rather keep like this for now and see whether we run into issues or not.

src/careamics/dataset_ng/dataset.py
jdeschamps: For now, I'd leave like this.

melisande-c (approved):

PR #491, opened 2025-06-16
Fix: N2V masking bugs

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Some bugs were discovered in the n2v masking functions. This PR fixes said bugs. Additionally validation has been added to the configuration to make sure that every patch will have at least one blind-spot pixel.

It seems the bugs were affecting training stability in some cases, hopefully these fixes will solve the problem.

### Overview - what changed?

- Replaced the `_get_stratified_coords_torch` function with a new simpler (still vectorised) implementation, that I think is actually closer to the original tensorflow implementation. 
- Modified how the replacement pixels are selected in `uniform_manipulate_torch`.
- Added new validation to the `Configuration`, to make sure there is always at least one masked pixel in each patch.
- Added optional `device` argument to `N2VManipulateTorch` to allow for local testing using the cpu when cuda is available.

The stratified coords func defines a grid of coordinates on the spatial dimensions which is the same for each sample in the batch. Then a random offset is added only in the spatial dimensions. Finally coordinates that lie outside of the patch dimensions are filtered, not clipped, this ensures that the average masked pixel percentage will converge to the chosen masked pixel percentage. See docs for a better explanation.

Made sure selected replacement pixels in `uniform_manipulate_torch` do not get mixed over a batch.

### New features or files

- `Configuration.validate_n2v_mask_pixel_perc`

- `_get_stratified_coords_torch`
- `uniform_manipulate_torch`
- `N2VManipulateTorch.__init__`

### Removed features or files

- `_odd_jitter_func_torch`

## How has this been tested?

Tested by observing results in notebooks, additional tests should be added. I will open an issue shortly

---

- [ ] Pre-commit passes

### DIFF
--- a/src/careamics/config/configuration.py
+++ b/src/careamics/config/configuration.py
@@ -6,6 +6,7 @@ import re
 from pprint import pformat
 from typing import Any, Callable, Literal, Union
 
+import numpy as np
 from bioimageio.spec.generic.v0_3 import CiteEntry
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 from pydantic.main import IncEx
@@ -182,6 +183,49 @@ class Configuration(BaseModel):
             )
 
         return name
+
+    @model_validator(mode="after")
+    def validate_n2v_mask_pixel_perc(self: Self) -> Self:
+        """
+        Validate that there will always be at least one blind-spot pixel in every patch.
+
+        The probability of creating a blind-spot pixel is a function of the chosen
+        masked pixel percentage and patch size.
+
+        Returns
+        -------
+        Self
+            Validated configuration.
+
+        Raises
+        ------
+        ValueError
+            If the probability of masking a pixel within a patch is less than 1 for the
+            chosen masked pixel percentage and patch size.
+        """
+        # No validation needed for non n2v algorithms
+        if not isinstance(self.algorithm_config, N2VAlgorithm):
+            return self
+
+        mask_pixel_perc = self.algorithm_config.n2v_config.masked_pixel_percentage
+        patch_size = self.data_config.patch_size
+        expected_area_per_pixel = 1 / (mask_pixel_perc / 100)
+
+        n_dims = 3 if self.algorithm_config.model.is_3D() else 2
+        required_patch_size = tuple(
+            int(np.ceil(expected_area_per_pixel ** (1 / n_dims))) for _ in range(n_dims)
+        )
+        required_mask_pixel_perc = (1 / np.prod(patch_size)) * 100
+        if expected_area_per_pixel > np.prod(patch_size):
+            raise ValueError(
+                "The probability of creating a blind-spot pixel is "
+                f"below 1, for a patch size of {patch_size} with a masked pixel "
+                f"percentage of {mask_pixel_perc}%. Either increase the patch size to "
+                f"{required_patch_size} or increase the masked pixel percentage to "
+                f"at least {required_mask_pixel_perc}."
+            )
+
+        return self
 
     @model_validator(mode="after")
     def validate_3D(self: Self) -> Self:
--- a/src/careamics/config/configuration.py
+++ b/src/careamics/config/configuration.py
@@ -212,17 +212,18 @@ class Configuration(BaseModel):
         expected_area_per_pixel = 1 / (mask_pixel_perc / 100)
 
         n_dims = 3 if self.algorithm_config.model.is_3D() else 2
+        patch_size_lower_bound = int(np.ceil(expected_area_per_pixel ** (1 / n_dims)))
         required_patch_size = tuple(
-            int(np.ceil(expected_area_per_pixel ** (1 / n_dims))) for _ in range(n_dims)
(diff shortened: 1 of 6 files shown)


src/careamics/config/configuration.py
```diff
@@ -183,6 +184,49 @@ def no_symbol(cls, name: str) -> str:
...
+        if not isinstance(self.algorithm_config, N2VAlgorithm):
+            return self
+
+        mask_pixel_perc = self.algorithm_config.n2v_config.masked_pixel_percentage
+        patch_size = self.data_config.patch_size
+        expected_area_per_pixel = 1 / (mask_pixel_perc / 100)
+
+        n_dims = 3 if self.algorithm_config.model.is_3D() else 2
+        required_patch_size = tuple(
+            int(np.ceil(expected_area_per_pixel ** (1 / n_dims))) for _ in range(n_dims)
```
jdeschamps: There is a constraint on the patch size: `all_elements_power_of_2_minimum_8`, so in order for the error message to be consistent with the constraint you'd need to compute the closest (larger) power of 2.

src/careamics/transforms/n2v_manipulate_torch.py
```diff
@@ -57,6 +60,8 @@ def __init__(
             N2V manipulation configuration.
         seed : Optional[int], optional
             Random seed, by default None.
+        device : str
```
jdeschamps, line 63: Should it be a literal?

src/careamics/transforms/pixel_manipulation_torch.py
```diff
@@ -75,50 +75,24 @@ def _apply_struct_mask_torch(
...
+    defining a grid and sampling a pixel in each grid square. The grid is defined such
+    that the resulting density of masked pixels is the desired masked pixel percentage.
 
     Parameters
     ----------
     mask_pixel_perc : float
-        Actual (quasi) percentage of masked pixels across the whole image. Used in
-        calculating the distance between masked pixels across each axis.
+        Actual (quasi) percentage of masked pixels across the whole image. Over many
+        iterations the percentage of masked pixels will converge to this value.
```
jdeschamps: To me the phrasing implies that the more you run it, the more the function will return a number of masked pixels closer to the percentage.

I think we can just stick to "Percentage of masked pixels across the whole image."


Not that important since it is a private function...

src/careamics/transforms/pixel_manipulation_torch.py
```diff
@@ -129,60 +103,48 @@ def _get_stratified_coords_torch(
...
+    # subset of this area will have the same expected masked pixel percentage, therefore
+    # we can filter the masked pixels that lie outside the patch size to have a patch
+    # with the desired masked pixel percentage.
+
+    batch_size = shape[0]
+    spatial_shape = shape[1:]
+
+    n_dims = len(spatial_shape)
+    expected_area_per_pixel = 1 / (mask_pixel_perc / 100)
+    # TODO: validate in config that expected_area_per_pixel < prod(patch size)
```
jdeschamps: this TODO can go!

tests/conftest.py
```diff
@@ -200,8 +200,17 @@ def minimum_n2v_configuration(
     Returns
     -------
     dict
-        A minumum configuration example.
+        A minimum configuration example.
     """
+    # update masked pixel percentage so config validation will pass
+    if "n2v_config" not in minimum_algorithm_n2v:
+        minimum_algorithm_n2v["n2v_config"] = {}
```
jdeschamps, line 207: can we also have a test for the validation in the `test_configuration.py`?

jdeschamps (approved):

src/careamics/transforms/pixel_manipulation_torch.py
melisande-c: Ok sure, maybe I can change it to "the expected value of the percentage of masked pixels" which implies in each individual image there can be slightly more or less, but the average will be the chosen masked pixel percentage.

src/careamics/transforms/n2v_manipulate_torch.py
melisande-c: You can also do things like "cuda:0" so I left as a string

src/careamics/transforms/pixel_manipulation_torch.py
melisande-c: done

src/careamics/config/configuration.py
melisande-c: done

tests/conftest.py
melisande-c: done

src/careamics/transforms/pixel_manipulation_torch.py
melisande-c: done

src/careamics/transforms/n2v_manipulate_torch.py
jdeschamps: Makes sense!

PR #505, opened 2025-07-02
Expose lightning trainer params 

### LINKED ISSUE #480: [Feature] Fixing number of steps per epoch
## Problem

Currently, the total number of training steps depends on the number of patches in the training set, which again depends on the patch size and the size of the images. Thus, when fixing the number of epochs, the number of total training steps can still vary significantly depending on the image size.

## Potential solution

It would be helpful to have an option for fixing the number of steps per epoch to ensure consistent training behavior regardless of the image size.

Independent of this setting, the computation of training data statistics should still be done over a representative number of patches.

CatEek:
## Description

> [!NOTE]  
> **tldr**: 

Allow configuring careamist with a native lightning trainer params

Mainly, allow limiting max train steps

### Overview - what changed?

Configuration helper functions now accept a dict with all possible lightning trainer params, all validation of which happen on lightning side

trainer_params dict allows to use any lightning configuration parameters. Num_epoch and num_steps are explicitly exposed 

## How has this been tested?

There's a relevant test and on the real application

## Related Issues

- Resolves #480 

## Breaking changes

Notebooks in Examples would have to be changed

---

- [ ] Pre-commit passes


src/careamics/config/callback_model.py
```diff
@@ -36,7 +36,7 @@ class CheckpointModel(BaseModel):
     save_last: Optional[Literal[True, False, "link"]] = Field(default=True)
     """When `True`, saves a last.ckpt copy whenever a checkpoint file gets saved."""
 
-    save_top_k: int = Field(default=3, ge=-1, le=100)
+    save_top_k: int = Field(default=1, ge=-1, le=100)
```
jdeschamps: I actually set to top 3 on purpose, since that is something people enjoy. I'd keep it and write a note stating that it is different from the defaults.

src/careamics/config/training_model.py
```diff
@@ -29,26 +29,15 @@ class TrainingConfig(BaseModel):
     model_config = ConfigDict(
         validate_assignment=True,
     )
+    lightning_trainer_config: Optional[dict] = None
+    """Configuration for the PyTorch Lightning Trainer, following PyTorch Lightning
+    Trainer class"""
 
-    num_epochs: int = Field(default=20, ge=1)
```
jdeschamps, line 33: So now we start having parts of the configuration that are just dict for Lightning, and some that are explicitly all the lightning parameters...

Having them explicit is helpful because we can catch divergences through the tests, but it requires more work and lines of code (as well as the validation on our side).

Having the dict is simple, but anytime we need to enforce something, we need to implement rather convoluted validation (e.g. `shuffler=True` in the data configuration).

Personally, I am rather in favor of making the parameters explicit and not giving access to all of them through our configuration. People using the Lightning API can do all the complicated stuff they want. 

1) Does Lightning actually perform validation? What happens when these are misconfigured?
2) I think `num_epochs` and `num_steps` should be in the configuration as they are expected and known to most users

jdeschamps: An intermediate solution is to have some parameters explicit, and others folded into a `dict`. That way, we have a clean API for users doing the normal stuff, and the dict for people wanted advanced stuff.

Let's debate what the best solution is here.

src/careamics/config/configuration_factories.py
```diff
@@ -539,13 +539,13 @@ def create_care_configuration(
...
     patch_size: list[int],
     batch_size: int,
-    num_epochs: int,
     augmentations: Optional[list[Union[XYFlipModel, XYRandomRotate90Model]]] = None,
     independent_channels: bool = True,
     loss: Literal["mae", "mse"] = "mae",
     n_channels_in: Optional[int] = None,
     n_channels_out: Optional[int] = None,
     logger: Literal["wandb", "tensorboard", "none"] = "none",
+    trainer_params: Optional[dict] = None,
```
jdeschamps: The point of the convenience functions for the configurations are to give direct access to users to the most important parameters, these parameters should then be easy to understand for most users (with the help of the documentation).

Folding important parameters such as `num_epochs` (or `max_`) and `num_steps` in a dict is not a good idea in my opinion.

While I find justifications for the optimizer and lr params, because these will be seldom used by most users, it is not the case of the number of epochs and steps, which will need to be set every single time.

I would put them as explicit parameters. And if the configuration takes a dict (which I think I am also not in favor), then the dictionary could be created/expanded in this function and passed to the internal ones for the creation of the configuration.

I would also make them mandatory.

jdeschamps: Also as a side note, you don't provide any example on how to pass these important parameters. (which is not necessary if they are mandatory)

jdeschamps (changes requested):

src/careamics/config/callback_model.py
CatEek: It will require loss monitor which we don't want to set

src/careamics/config/training_model.py
CatEek: Don't have to strong of an opinion but don'tsee immediate value in duplication Lighning functionality. It does it;s own validation and throws more user-friendly errors than pydantic

src/careamics/config/configuration_factories.py
CatEek: It's very straightforward `trainer_params={"max_epochs": 10}` However, i see the point. We can add a simple logic with just 2 mutually exclusive parameters, max_epochs or mas_steps. If they're not provided explicitly we ask for the dict according to lightnig docs

src/careamics/config/callback_model.py
jdeschamps: Can you give more details? Because it was not described in the PR description and I missed the modification when doing the PR review.

Why don't we want to have a loss monitor?
Does this impact the loss plotting at the end? (probably not?)

I do think that saving multiple "top-N" checkpoints is a useful features for users.

src/careamics/config/training_model.py
jdeschamps: If our users wanted to use Lightning, they'd use Lightning. What the `CAREamist` offer is a simplified API where we make an intermediate decision on what are important parameters likely to be used by users.

I'm leaning towards the hybrid version, which is:
> Only expose useful parameters, hide the others in a `dict`.

Extra code for us to maintain, but at least it clarifies things for users. We can't expect them to go in the Lightning docs to understand what they need to do.

jdeschamps: Following this, because people might create a configuration in other ways that with the convenience functions, my opinion is to have something like that:

```python

class CheckpointModel(BaseModel):

    num_epochs: int
    num_step: int
    params: dict

   @field_validator(params)
   def does_not_contain_other_params():
        ....


# careamist.py
    self.trainer = Trainer(
        ...
        max_epochs=self.cfg.train_config.num_epochs,
        max_steps=self.cfg.train_config.num_steps,
        **self.cfg.train_config.params
   )
```

Yes it is more cumbersome to maintain and requires additional validation, but:
1. Clean API for users, regardless of how they create their config
2. `config.yml` is logical with respect to the API 
3. Clarifies our expectations (max_epoch and max_steps need to be set), avoiding behaviour coming from not setting those

An alternative, is to pack everything in the configuration is a dict and have the validation throw an error when steps and epochs are not set. But then the Pydantic configuration and its schema are not explicit about these and I don't like that.

src/careamics/config/configuration_factories.py
jdeschamps: See answers in the other thread.

jdeschamps:
There are some minor conflicts due to recent changes in `main`

gatoniel:
It would be awesome if that PR could also be used to set other trainer params, e.g., [`val_check_interval`](https://lightning.ai/docs/pytorch/stable/common/trainer.html#val-check-interval). This would solve my question raised in this issue: https://github.com/CAREamics/careamics/issues/510

jdeschamps:
@CatEek You have merge conflicts with main!

src/careamics/config/callback_model.py
CatEek: Yes, it might be useful. But It conflicts with num_steps parameter. Maybe later we can combine thew 2 in a simple wrapper to simplify the API

src/careamics/careamist.py
```diff
@@ -828,7 +822,7 @@ def predict_to_disk(
             source_path = source.pred_data
             source_data_type = source.data_type
             extension_filter = source.extension_filter
-        elif isinstance(source, str | Path):
+        elif isinstance(source, (str, Path)):
```
jdeschamps: Since https://github.com/CAREamics/careamics/commit/888799380ed2fb1c606d6cb3c2b02fe66dbc64e7 pre-commit hooks (ruff presumably) should correct this to `str | Path` (python 3.10).

So these changes should be reverted.

src/careamics/careamist.py
```diff
@@ -841,7 +835,7 @@ def predict_to_disk(
             raise ValueError(
                 "Predicting to disk is not supported for input type 'array'."
             )
-        assert isinstance(source_path, str | Path)  # because data_type != "array"
+        assert isinstance(source_path, (Path, str))  # because data_type != "array"
```
jdeschamps: Same here!

src/careamics/careamist.py
jdeschamps: It probably comes from merging?

src/careamics/config/training_model.py
jdeschamps: When you close a discussion, please indicate what you changed to resolve it... Otherwise it;s extra work for us to figure out what happened.

(I'll mark it as unresolved so you can see the message)

jdeschamps (changes requested):
So I am still not convinced by the PR, but the reason is mostly because of Lightning and a fact that was not clear to me: setting a maximum number of **global steps** is not really what we want and what our users want (see https://github.com/CAREamics/careamics/issues/480 https://github.com/CAREamics/careamics/issues/510, and the [lightning API](https://lightning.ai/docs/pytorch/stable/common/trainer.html#max-steps)). Typically, this fact should have been mentioned in the PR description...

The problem often raised is that experiments may vary tremendously in the number of steps per epochs due to data size. The "classical" solution that users expect is to be able to limit the number of steps within each epoch.

Here, this `max_step` parameter is not doing that. So why should we explicit it? It will rather confuse everyone (starting by me during my first review).

That brings me to the following questions:
- Is it possible at all with Lightning to set the number of steps per epoch?
- If not what solution can we provide?
    - Limit the number of patches generated per epoch? (`max_steps / batch_size`)
    - Sample more often the validation set? (to match a user-defined set of steps). That seems too messy

CatEek:
> So I am still not convinced by the PR, but the reason is mostly because of Lightning and a fact that was not clear to me: setting a maximum number of **global steps** is not really what we want and what our users want (see #480 #510, and the [lightning API](https://lightning.ai/docs/pytorch/stable/common/trainer.html#max-steps)). Typically, this fact should have been mentioned in the PR description...
> 
> The problem often raised is that experiments may vary tremendously in the number of steps per epochs due to data size. The "classical" solution that users expect is to be able to limit the number of steps within each epoch.
> 
> Here, this `max_step` parameter is not doing that. So why should we explicit it? It will rather confuse everyone (starting by me during my first review).
> 
> That brings me to the following questions:
> 
> * Is it possible at all with Lightning to set the number of steps per epoch?
> * If not what solution can we provide?
>   
>   * Limit the number of patches generated per epoch? (`max_steps / batch_size`)
>   * Sample more often the validation set? (to match a user-defined set of steps). That seems too messy

Yes, it's called limit_train_batches. And what i proposed a user can just pass any params from the lightning trainer as a dict.

src/careamics/careamist.py
CatEek: I don't know because it's been a while

src/careamics/config/training_model.py
CatEek: I'm confused about the meaning of this discussion. num_steps is explicit now

src/careamics/careamist.py
CatEek: I think we'd have to go through typing again in a separate PR following the transition to newer python version

jdeschamps:
> .



> > So I am still not convinced by the PR, but the reason is mostly because of Lightning and a fact that was not clear to me: setting a maximum number of **global steps** is not really what we want and what our users want (see #480 #510, and the [lightning API](https://lightning.ai/docs/pytorch/stable/common/trainer.html#max-steps)). Typically, this fact should have been mentioned in the PR description...
> > The problem often raised is that experiments may vary tremendously in the number of steps per epochs due to data size. The "classical" solution that users expect is to be able to limit the number of steps within each epoch.
> > Here, this `max_step` parameter is not doing that. So why should we explicit it? It will rather confuse everyone (starting by me during my first review).
> > That brings me to the following questions:
> > 
> > * Is it possible at all with Lightning to set the number of steps per epoch?
> > * If not what solution can we provide?
> >   
> >   * Limit the number of patches generated per epoch? (`max_steps / batch_size`)
> >   * Sample more often the validation set? (to match a user-defined set of steps). That seems too messy
> 
> Yes, it's called limit_train_batches. And what i proposed a user can just pass any params from the lightning trainer as a dict.

Ok great, but then here what we should do to make it clear to our users:

- `max_epochs` and `num_steps` parameters in the convenience functions, where `num_steps` is translated to `limit_train_batches`
- Place these parameters higher in the parameters list of the various functions
- Remove `max_steps` from the explicit list of parameters because it is confusing and we don't actively use it. Advanced Lightning users can set it through the dictionary.
- Add examples of setting number of epochs and steps in the docstring of the convenience functions.
- Document that `max_epochs` is by default 1000 and point to the Lightning docs.

Note that the `limit_train_batches` has a weird behaviour when passing values in the `]0,1[`, we should document that in the docstring as well and point to the Lightning docs. People will be surprised, and it is our responsibility to make transparent the parameters that we explicitly expose!

I like `limit_val_batches` as well, but out of scope here.

src/careamics/config/callback_model.py
jdeschamps: Do you mean that it conflicts with `max_steps` or `limit_train_batches`? Because in truth, we don't care about `max_steps` and we care about saving multiple "best" checkpoints.

src/careamics/config/training_model.py
jdeschamps: Well, most of the discussion was before it was made explicit again (last comment on July 8, commit https://github.com/CAREamics/careamics/pull/505/commits/5405d43d41531a2ab2bd9ba3d7083710bdc1515f on July 22). As for my [comment 3 days ago](https://github.com/CAREamics/careamics/pull/505#discussion_r2312366994), it is to ask you to answer comments on PRs so that we know what you did with the comment:

- "I included the suggestion"
- "I disagree with the suggestion because of this and that"

src/careamics/config/callback_model.py
CatEek: ok, left 3

src/careamics/cli/conf.py
jdeschamps, line 1: Here we should add a TODO for the number of steps.

src/careamics/config/callback_model.py
```diff
@@ -24,7 +24,7 @@ class CheckpointModel(BaseModel):
 
     model_config = ConfigDict(validate_assignment=True, validate_default=True)
 
-    monitor: Literal["val_loss"] = Field(default="val_loss")
+    monitor: str | None = Field(default="val_loss")
```
jdeschamps: What else but `val_loss` can be put here? Why leave it to any string? 

Since this is user facing, we can put CAREamics constraints on top of the Lightning ones.

src/careamics/config/configuration_factories.py
```diff
@@ -539,13 +557,15 @@ def create_care_configuration(
...
     patch_size: list[int],
     batch_size: int,
-    num_epochs: int,
     augmentations: list[Union[XYFlipModel, XYRandomRotate90Model]] | None = None,
     independent_channels: bool = True,
     loss: Literal["mae", "mse"] = "mae",
     n_channels_in: int | None = None,
     n_channels_out: int | None = None,
     logger: Literal["wandb", "tensorboard", "none"] = "none",
+    num_epochs: int | None = None,
```
jdeschamps: Can we put `num_epochs` and `num_steps` higher in the list of parameters? They are important.

In addition, currently we offload the default value for `num_epochs` to Lightning. But that means whenever they will change something, we will not know about it. I am strongly in favor of not leaving `None` -> `100`, but rather forcing users to set the number of `epochs` themselves.

jdeschamps: This comment can be duplicated in the other algorithms convenience functions.

src/careamics/model_io/bmz_io.py
```diff
@@ -186,11 +186,15 @@ def export_to_bmz(
         )
 
         # test model description
-        test_kwargs = (
-            model_description.config.bioimageio.model_dump()
-            .get("test_kwargs", {})
-            .get("pytorch_state_dict", {})
-        )
+        test_kwargs = {}
+        if hasattr(model_description, "config") and isinstance(
```
jdeschamps, line 190: Is this covering the case when the Lightning API did not export the config in state dict?

What does this do? Could you comment on it?

I am wondering whether we need to write a warning to Lightning users when they are exporting to BMZ that something is missing (but rn I don't understand this piece of code).

jdeschamps (changes requested):

src/careamics/config/callback_model.py
CatEek: train_loss, other metrics

src/careamics/model_io/bmz_io.py
CatEek: I honestly don't know. We sure can create an issue and investigate

jdeschamps: @mese79 do you know what this does?

jdeschamps (approved):

CatEek:
Updated convenience functions for new algorithms, wip, to be addressed in a follow-up PR

PR #532, opened 2025-09-09
Feat: Exclude background patches during training

### LINKED ISSUE #293: Next-Gen Dataset: Exclude background patches during training
## Problem

In microscopy, images often contain large background areas. During training, this will cause the network to spend substantial amount of time learning "nothing". Being able to exclude background regions would be a useful feature that many users would probably enjoy.

### Solution 1: CSBDeep-style

There is no reason not to copy what has been developed before in CSBDeep:

```python
# parameters:
# threshold
# percentile
from scipy.ndimage import maximum_filter
def _filter(datas, patch_size, dtype=np.float32):
    image = datas[0]
    if dtype is not None:
        image = image.astype(dtype)
    # make max filter patch_size smaller to avoid only few non-bg pixel close to image border
    patch_size = [(p//2 if p>1 else p) for p in patch_size]
    filtered = maximum_filter(image, patch_size, mode='constant')
    return filtered > threshold * np.percentile(image,percentile)
return _filter
```
[Source](https://github.com/CSBDeep/CSBDeep/blob/282664fc294e8ba5d00a6ea82fcadcf9198a24b9/csbdeep/data/generate.py#L20)

### Solution 2: use `mean` and `std`

Exclude patches whose mean is `N` standard deviation away from the dataset mean.

## Relevant other issues

https://github.com/CAREamics/careamics/issues/292

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Add coordinate and patch filtering in the NG Dataset via configurable modules.

Background patch filtering is an important feature important to users. 

### Overview - what changed?

This PR adds two different mechanisms to the NG Dataset: coordinate and patch filtering. The corresponding modules filter either coordinates or patches as they are requested, until either a foreground patch is found or patience has been exhausted.

The filters are based on either a binary mask of the data, or statistical criteria (max, mean, std, Shannon entropy).

Filters are their own modules and implement two protocols: `CoordinateFilterProtocol` or `PatchFilterProtocol`. They each have their own set of parameters and can be declared in the `NGDataConfig`. They are instantiated by the `CAREamicsDataset` and applied in its `__get_item__` method.

Because training is done on random coordinates/patches and these are not computed beforehand, the `__get_item__` method includes now a `while` loop to pull coordinates/patches until the conditions are met. The `while` loop can be escaped when a foreground patch is found, a background patch is selected on purpose or when `patience` is exhausted (default value is `10`).

The implemented filters are the following:
- Mask coordinate filtering: by providing a (positive) mask of the data, a masked patch corresponding to the requested coordinates can be extracted and accepted based on the percentage of masked pixels.
- Max filtering: following CSBDeep, the max filtering allow filtering patches by thresholding the minimum value of the maximum filtered patch.
- MeanSTD filtering allows filtering on the mean of the patch, and optionally on the std as well.
- Shannon entropy filtering: allows filtering on the Shannon entropy of a patch, which is sensitive to statistical distributions.

Both coordinate and patch filtering are compatible and work together, although this is probably not a realistic use case. Finally, I also added static methods to each filter module to allow users to assess what parameters they should use depending on their data.

### New features or files

- `dataset_ng.patch_filter` module with the filter submodules and classes, including factories.
- `config.data.patch_filter` module with the corresponding Pydantic modules

- `NGDatasetConfig` with the new parameters `coordinate_filter`, `patch_filter` and `patch_filter_patience`
- `CAREamicsDataset` now instantiates the filters and applies them in a `while` loop in `__get_item__`
- Convenience functions that create `CAREamicsDataset` now also take a `mask` parameter for use with the coordinate filter.

## How has this been tested?

- unit tests
- toy example
- CZI dataset provided by collaborator

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/293

### Usage

Here is an example of a benchmark with a toy example

```python
import numpy as np
import tifffile
from skimage.measure import shannon_entropy
from skimage.filters import gaussian
from scipy.ndimage import maximum_filter

from careamics.config.configuration_factories import (
    _create_ng_data_configuration,
    _list_spatial_augmentations,
)
from careamics.dataset_ng.dataset import Mode, CareamicsDataset
from careamics.dataset_ng.factory import create_array_dataset, create_tiff_dataset
from careamics.dataset_ng.patch_extractor.patch_extractor_factory import create_array_extractor
from careamics.dataset_ng.patch_filter import (
    MeanStdPatchFilter,
    ShannonPatchFilter,
    MaxPatchFilter,
    MaskCoordFilter,
)
from careamics.transforms.normalize import Denormalize

def plot(dataset: CareamicsDataset):
    stats = dataset.input_stats
    
    n_fg = 0
    n_bg = 0

    # show several rows of patches
    fig, ax = plt.subplots(4, 5, figsize=(14, 14))
    for j in range(4):    
        for i in range(5):
            sample = dataset[i+4*j][0]
            # denormalize
            data = np.array(Denormalize(
                image_means=stats.means, image_stds=stats.stds
            )(sample.data[np.newaxis, ...])).squeeze()

            ax[j, i].imshow(data)
            ax[j, i].axis("off")
            ax[j, i].set_title(
                f"Max: {data.max():.1f}\n"
                f"Mean: {data.mean():.1f}\n"
                f"Entropy: {shannon_entropy(data):.1f}"
            )

            if data.mean() > 10:
                n_fg += 1
            else:
                n_bg += 1

    print(f"Foreground patches: {n_fg}, background patches: {n_bg}")
    print(f"Foreground ratio: {n_fg / (n_fg + n_bg):.2f}")

# create images with square signal
obj_size = 32
border = 300
base_val = 75
img = np.zeros((1025, 1025))

# fill border
img[border:-border, border:-border] = base_val

for i in range(300):

    # draw coordinates
    x = rng.integers(border, img.shape[0] - border - obj_size)
    y = rng.integers(border, img.shape[1] - border - obj_size)

    img[x : x + obj_size, y : y + obj_size] = rng.choice([255, 220, 170]) 

print(f"Image mean: {img.mean():.2f}, std: {img.std():.2f}")
mask = img >= base_val

# side by side
fig, ax = plt.subplots(1, 2, figsize=(10, 5))
ax[0].imshow(img, cmap="gray")
ax[0].set_title("Image")
ax[1].imshow(mask, cmap="gray")
ax[1].set_title("Mask")
plt.show()

# Create dataset from an array
train_data_config = _create_ng_data_configuration(
    data_type="array",
    axes="YX",
    patch_size=(64, 64),
    batch_size=1,
    augmentations=[]
)

train_dataset = create_array_dataset(
    config=train_data_config,
    mode=Mode.TRAINING,
    inputs=[img],
    targets=None,
)

# plot patches without filtering
plot(train_dataset)

# add filtering and plot patches
train_dataset.patch_filter_patience = 20
train_dataset.patch_filter = None
train_dataset.coord_filter = MaskCoordFilter(
    mask_extractor=create_array_extractor(
        source=[mask],
        axes="YX"
    ),
    coverage=0.1,
)
# train_dataset.patch_filter = MaxPatchFilter(threshold=10)
# train_dataset.patch_filter = MeanStdPatchFilter(20)
# train_dataset.patch_filter = MeanStdPatchFilter(20, 10)
# train_dataset.patch_filter = ShannonPatchFilter(0.9)

plot(train_dataset)
```

This lead to the following data, considering a foreground patch has having at least a mean of 10 (N=20):
|  Filter  | Fg patch % |
| ------------- | ------------- |
| No filter  | 25%  |
| Mask filter  | 100%  |
| MeanSTD (mean only)  | 95%  
| MeanSTD  | 100%  |
| Shannon  | 100%  |

All patch filters require parameters that are data-dependent. To help them make an informed decision, they can run static methods from all filters to obtain a stack that can then be used to plot expected masking with various parameters:

```python
filter_map = MaxPatchFilter.filter_map(
    image=img[70],
    patch_size=[64, 64],
)

fig, ax = plt.subplots(1, 5, figsize=(20, 5))
for i, thresh in enumerate([50 + i*5 for i in range(5)]):
    ax[i].imshow(MaxPatchFilter.apply_filter(filter_map, thresh))
    ax[i].set_title(f"Threshold: {thresh}")
plt.show()
```

All patch filters have the same API. The docs provide examples on how to use them.

### CSBDeep background removal

Some differences with respect to CSBDeep implementation of the max filter: in CSBDeep a global percentile (high) and a weight (e.g. 0.4) are used and computed over the entire image to get a mask. This mask is then used to generate coordinates.

Since we don't expect to have the full image in memory (e.g. zarr), users give directly a threshold and it is applied on the maximum filtered patch (as opposed to the maximum filtered global image in CSBDeep).

### Sampler

My understanding of the sampler is that coordinates need to be known in advance. Since this is not the case with the random patching, I decided to apply directly the filters in the dataset. 

---

- [x] Pre-commit passes


jdeschamps:
I have to fix the tests, they run fine locally though.

src/careamics/config/data/patch_filter/__init__.py
veegalinova, line 1: I would put all the models into one file since they are all very short

src/careamics/config/data/patch_filter/filter_model.py
veegalinova, line 1: Not inheriting might be preferable, so it's easy to see each model's parameters at a glance. 
But it's not a strong opinion.

veegalinova:
The visualizations of different filtering maps: 
2d_comparison
3d_comparison


An example of how to find a threshold, will need to write it in docs: 
threshold

veegalinova:
@jdeschamps  I have added a few commits, here is an overview of the changes: 

1. added seed propagation to the filters from the main config seed 
2. added train data mask parameter to the lightning data module 
3. refactored the filtering logic in the dataset slightly - moved it to a separate method

I've tested the training using care U2OS dataset. The iteration time increases pretty significantly when enabling the filtering, but it should be manageable with an appropriate number of workers. I've done two changes after testing, though:

4. reduced default patch filtering patience to 5 to speed up the iterations a bit 
5. changed the behaviour of the max filter. Now patch will be filtered if more than 25% of the area is lower than the threshold (the percentage can be configured) 

Lmk what you think and if I'm missing something

jdeschamps:
I will look into it! 

Could you have a go at solving the conflicts? I think they arise from your PR I just merged. Thanks a lot!!!

jdeschamps:
I just pushed a pre-commit fix, your changes are good. On my side, it's ready to merge.

veegalinova (approved):

PR #542, opened 2025-09-15
Chore: Remove py310 and add newer versions

jdeschamps:
## Description

Zarr v3 is unfortunately not compatible with python 3.10. Since Zarr is so central to the future developments of CAREamics, we can hardly ignore that and unfortunately I think we should already drop 3.10, one year in advance.

This PR simply changes the metadata of the package, CIs and pre-commits to enforce >py310.


jdeschamps:
Superseded by https://github.com/CAREamics/careamics/pull/545

PR #545, opened 2025-09-15
Chore: Upgrade python, Zarr and numpy versions

jdeschamps:
## Description

OME-Zarr is now only supporting `zarr` v3, which is only compatible with `py>=311`. Because of the interplay between the two, and with `numpy` 2.0.0, the only easy solution for upgrade is a joint PR.

This PR:
- Update `pyproject.toml` for compatibility with python>=3.11 and zarr>=3.0.0
- Update CIs to run on py311, 312 and 313
- Update pre-commit to run checks with py311
- Update zarr-related classes with Zarr v3 API
    - Allow both path and url to create Zarr, either a local store or a file system store
    - Add `validators` dependency to validate url
- Lift numpy version


src/careamics/file_io/read/zarr.py
```diff
@@ -2,12 +2,11 @@
...
 
-from zarr import Group, core, hierarchy, storage
+from zarr import Array, Group, storage
 
 
-def read_zarr(
-    zarr_source: Group, axes: str
-) -> Union[core.Array, storage.DirectoryStore, hierarchy.Group]:
+# TODO where is this used?
+def read_zarr(zarr_source: Group, axes: str) -> Union[Array, storage.LocalStore, Group]:
```
melisande-c: Yeah I guess this function can be removed as it is old, I guess it was to give to the `CAREamist` `read_func` argument, but it would load all the zarrs into memory?

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
```diff
@@ -17,7 +17,7 @@ class ZarrImageStack:
 
     # TODO: keeping store type narrow so that it has the path attribute
     #   base zarr store is zarr.storage.Store, includes MemoryStore
-    def __init__(self, store: zarr.storage.FSStore, data_path: str, axes: str):
+    def __init__(self, store: LocalStore | FsspecStore, data_path: str, axes: str):
```
melisande-c, line 21: Don't worry about this now but I guess at some point it would be good to make store typed as `StoreLike` and come up with a better value for the `source` property.

pyproject.toml
```diff
@@ -52,9 +53,10 @@ dependencies = [
     'pyyaml<=6.0.2,!=6.0.0',
     'typer>=0.12.3,<=0.17.4',
     'scikit-image<=0.25.2',
-    'zarr<3.0.0',
+    'zarr>=3.0.0a0,<4.0.0', # zarr v3 requires Python >3.10
```
melisande-c: I can't find a version 3.0.0a0 is it a typo? or am I not looking in the right place

melisande-c (approved):
Looks good!

if you didn't notice there are couple pre-commit errors more than the one we have decided to ignore

melisande-c:
> if you didn't notice there are couple pre-commit errors more than the one we have decided to ignore

I think the mypy ones in the `XYRandomRotate` are coming from new numpy typing

pyproject.toml
jdeschamps: They are pre-releases: https://pypi.org/project/zarr/#history, `3.0.0` and others have been yanked. This was a fix for `3.10` actually (since otherwise the CI would crash), when I was doing parallel PRs for python, Zarr and numpy.

I might as well put `3.0.8` as the lower one!

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
jdeschamps: Agreed!

Will update the TODO.

src/careamics/file_io/read/zarr.py
jdeschamps: I will have a second look, but we should yank it.

jdeschamps:
> > if you didn't notice there are couple pre-commit errors more than the one we have decided to ignore
> 
> I think the mypy ones in the `XYRandomRotate` are coming from new numpy typing

Yeah a lot of the pre-commit errors are from main, because some things were merged before they were fixed.... A PR should come to fix that...

But that indeed prevented me from seeing the `xy_random_rotate90` ones which are not in main. So good catch!

pyproject.toml
jdeschamps: Solved here https://github.com/CAREamics/careamics/pull/545/commits/4267d297e5de55eec084f8c5a49f7c9abd019da7

src/careamics/file_io/read/zarr.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/545/commits/8b77aff646cefd9562def846e3ac83c0e469e02e

PR #546, opened 2025-09-16
Feat: remove empty patches with sampler proposal

veegalinova:
This is a proposal for PR "Feat: Exclude background patches during training" #532. 

### Overview - what changed?
Here I propose a different approach to implementing the patch filtering. 
Instead of doing it on the dataset level, move it to the Lightning datamodule level using PyTorch sampler. The main difference is that the sampler is tasked with creating "valid" indices of the training dataset. This simplifies the dataset and separates the concerns. 

Another thing (which I did by mistake at first, but I think is actually logical) - the sampler pre-filters the dataset based on the coordinate filter first, so the masked patches are never loaded. This modifies the behavior from the previous implementation, where they were returned after the patience was exhausted. I think this is more logical, that if the user does not want to use a specific portion of the data, it's never used at all.


src/careamics/dataset_ng/samplers/filtering_sampler.py
```diff
@@ -0,0 +1,76 @@
...
+        if self.coord_filter is None:
+            return list(range(len(self.dataset)))
+
+        valid_indices = []
+        for idx in range(len(self.dataset)):
+            patch_spec = self.dataset.patching_strategy.get_patch_spec(idx)
+            if not self.coord_filter.filter_out(patch_spec):
+                valid_indices.append(idx)
+
+        return valid_indices
```
jdeschamps, line 45: This won't work if `seed` is not set, but `seed` is by default `None` and calling the same index twice will result in different coordinates. That's why I ended up not using the sampler approach.

But come to think about it, if `seed` is not set, then why not letting the NG Dataset choose a random seed? That way the sampling will work and we can log a `seed` that allows reproducing the exact indices.

src/careamics/dataset_ng/samplers/filtering_sampler.py
```diff
@@ -0,0 +1,76 @@
...
+        for idx in indices:
+            patience = self.patience
+            while patience > 0:
+                patch_spec = self.dataset.patching_strategy.get_patch_spec(idx)
+                patch = self.dataset.input_extractor.extract_patch(
+                    data_idx=patch_spec["data_idx"],
+                    sample_idx=patch_spec["sample_idx"],
+                    coords=patch_spec["coords"],
+                    patch_size=patch_spec["patch_size"],
+                )
```
jdeschamps, line 61: This should work:
```suggestion
                patch = self.dataset.input_extractor.extract_patch(
                    **patch_spec,
                )
```

jdeschamps (changes requested):
That looks great!

My only issue, unless I missed something, is that the dataset uses random patching and therefore will return different coordinates when the same index is called twice (once when filtering based on mask, once when pulling the patch). That will cause the filtering on the indices to do nothing in effect. That was the reason I did not use the sampler in the first place.

As I mentioned in the comment, maybe a way around that makes sense is to randomly draw a seed to force reproducibility of that particular instance of the NG Dataset:

```python 
class CareamicsDataset(Dataset, Generic[GenericImageStack]):
    def __init__(
        ...

        self.seed = self.config.seed
        if self.seed is None:
            rng = np.random.default_rng()
            self.seed = rng.integers(0, 1_000_000)
```

[EDIT]: Actually the `RandomPatchingStrategy` does not return the same coordinate for the same index when seeded. Only `FixedRandompatchingStrategy` does, so we would need to use that class and there would be no need to "invent" a seed.

src/careamics/dataset_ng/patching_strategies/random_patching.py
jdeschamps, line 1: Would it make sense to add deprecated mention to the RandomPatching? I guess it would not be used anywhere anymore.

src/careamics/lightning/callbacks/dataset_reshuffle_callback.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        if trainer.datamodule is None or not hasattr(
+            trainer.datamodule, "train_dataset"
+        ):
+            return
+
+        training_dataset = trainer.datamodule.train_dataset
+
+        # Handle CareamicsDataset with FixedRandomPatchingStrategy
+        if isinstance(training_dataset, CareamicsDataset):
```
jdeschamps, line 41: Should we add a TODO in case we never allow a different strategy? (which is quite likely)

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
```diff
@@ -54,6 +55,21 @@ def __init__(
         # TODO: how to support metric evaluation better
         self.metrics = MetricCollection(PeakSignalNoiseRatio())
 
+    def configure_callbacks(self) -> list[L.Callback]:
```
jdeschamps, line 58: Does the list of callbacks used during training get extended with those provided in the trainer or do they override them?

This discussion concerns me a little: https://github.com/Lightning-AI/pytorch-lightning/issues/18784#issue-1938443161.

jdeschamps (changes requested):
In principle, it looks all good to me. I just have one concern regarding the callback!

jdeschamps:
A not so random thought coming from playing with a large dataset: let's say we are dealing with a very large image with mostly background. The fixed random patching strategy would generate `N` patches, of which the vast majority will be background. To guide sampling, let's say we have a mask.

Currently, the `FilteringSampler` does the following:
```python
        valid_indices = []
        for idx in range(len(self.dataset)):
            patch_spec = self.dataset.patching_strategy.get_patch_spec(idx)
            if not self.coord_filter.filter_out(patch_spec):
                valid_indices.append(idx)

        return valid_indices
```

That means that we may end up with an extremely small number of valid indices. Each epoch would then be just a few iterations.

I feel that in such cases it may be better to have the `patching_strategy` loop over `N` to extract valid patches... Ofc in case where there is really not much valid areas and `N` is large, this would mean repeating the same areas many times over the course of an epoch.

In addition, in such cases, it may be still worthy to build a list of valid coordinates and use this as a mask.

[EDIT]: after discussion and playing around, the probability of selecting an signal still scales directly with the percentage of surface occupied by the signal, so it is fair.

src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py
veegalinova: Added a test that checks the callbacks, all good

src/careamics/config/data/ng_data_model.py
```diff
@@ -325,6 +337,62 @@ def validate_dimensions(self: Self) -> Self:
...
+        unless they already have a seed explicitly set.
+
+        Returns
+        -------
+        Self
+            Data model with propagated seeds.
+        """
+        if self.seed is not None:
+            updated_transforms = []
+            for transform in self.transforms:
```
jdeschamps, line 355: Why not changing the seed _in place_ in the transforms?

tests/lightning/test_filtering_sampler_integration.py
```diff
@@ -0,0 +1,164 @@
...
+from careamics.config.data import NGDataConfig
+from careamics.lightning.callbacks import (
+    DatasetReshuffleCallback,
+    HyperParametersCallback,
+)
+from careamics.lightning.dataset_ng.data_module import CareamicsDataModule
+from careamics.lightning.dataset_ng.lightning_modules import N2VModule
+
+
+def create_default_configuration():
```
jdeschamps, line 19: We should probably use fixtures defined in `conftest`. For now a todo to this function will do in order to not have adversarial effects when changing fixtures.

jdeschamps (approved):

veegalinova:
Scrapping this PR due to issues with fixed random patching - with this implementation, for datasets with a low signal ratio number of training patches will be either very low or empty patches will be chosen instead. I did not find any easy solutions to this.

PR #564, opened 2025-10-02
Fix pre-commit

jdeschamps:
## Description

This PR fixes current pre-commit errors that were introduced in a recent merge. It also adds typing ignore to the `czi` errors that we previously had in `main`. These have now a todo to kick them out the day we have the CZI official support.

Note: there were many line too long errors from code, but I expect these to be left over from LVAE madness, so I just broke the code line. It should be clear enough should we ever need to uncomment that.


veegalinova (approved):
All good

PR #572, opened 2025-10-09
Convert CI to uv

jdeschamps:
Convert the CI to `uv` in the hope that it simplifies maintenance and make CI run faster.

What happens:
- CIs are reimplemented following [uv docs](https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi)
- smoke test added to have a simple `CAREamist` import during the CI, after packaging but before upload to PyPI (following `uv` docs)
- The `--pre` release version installation and automated issue on failure mechanism has been removed, as it was broken anyway
- `ruff` python version reference is moved to `pyproject.toml` with `fix = true`
-  `dev` dependency group is now a `dependency-groups` rather than an `optional-dependency`, which makes a difference for packaging
- Add `uv` lock
- Modify `get_careamics_version` as the CI creates `v0.1.dev-<hash>.<hash>` type of versions (instead of `v0.1.dev-<hash>`).

Notes:
- Deployment has not been tested, there might be an issue with Github trusted publishers
- I am not sure yet how to deal with the `uv` lock in the future, it will be modified locally whenever we run `uv` and may be often modified when working on PRs


melisande-c (approved):
I think it looks all good 😄

has codcov job also been edited to run on every run again, including PRs?

jdeschamps:
> has codcov job also been edited to run on every run again, including PRs?

It should run only new version pushes. What I removed was just unnecessary conditions!

https://github.com/CAREamics/careamics/blob/1c9fafe8d8dc973466938249f5e2ced8c175fc78/.github/workflows/codecov.yml#L3C1-L7C1

PR #587, opened 2025-10-17
Feature(Next-Gen Dataset): Pad patches queried outside of data range

### LINKED ISSUE #496: [Feature] Next-Gen Dataset: patches larger than data
## Problem

CZI is compatible with querying patches larger than data (padding with zeroes). This is handy when using many images and allow no sacrificing patch size.

Can we make this a feature compatible with other formats?

melisande-c:
## Description

> [!NOTE]  
> **tldr**: An `ImageStack` might be given coords and a patch size that go over the bounds of the image. Previously in the `InMemoryImageStack` and the `ZarrImageStack` this was handled in the default numpy slicing way, which is that the resulting patch may have smaller dimensions than the specified patch size. This PR modifies the behaviour so that the patch returned will always be the same size and the missing data is filled with zeros.

This is mostly a useful feature during inference. Someone might have a dataset with one image smaller than the rest and choosing a smaller patch size to accommodate this will make the process slower. Or maybe if a dataset is comprised of many small images of different sizes, padding them makes inference quicker as they can be processed in batches.

However, this feature probably shouldn't be used during training because then the model will see only zeros in some areas of patches. ~~I will add a guard against images being smaller than the patch size shortly.~~

EDIT: A guard to prevent data smaller than the patch size during training has been added to the dataset initialisation.

### Overview - what changed?

- `InMemoryImageStack` and `ZarrImageStack` will return zeros from the method `extract_patch` when the patch is outside the range of the data. (`CziImageStack` already did this.)
- `TilingStrategy` has been updated to handle the case for when an image is smaller than the tile size.
- The range the random strategies sample from has also been updated to clip to [0, image_size].
- Test cases have been added to existing tests for image sizes being smaller than the patch size.

Padding is done by inserting the extracted patch data into a zero array of the patch size.

### New features or files

- `pad_patch` function added to new `careamics/dataset_ng/patch_extractor/image_stack/utils.py`

- `InMemoryImageStack`
- `ZarrImageStack`
- `TilingStrategy`
- `_generate_random_coords` in `src/careamics/dataset_ng/patching_strategies/random_patching.py`
- Relevant tests

## How has this been tested?

Existing tests have had parametrisations added for the case that the image is smaller than the patch size.

## Related Issues

- Resolves #496 

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
@@ -13,16 +13,43 @@ def pad_patch(
     data_shape: Sequence[int],
     patch_data: NDArray[T],
 ) -> NDArray[T]:
+    """
+    Pad patch data with zeros where it is outside the bounds of it's source image.
+
+    This ensures the patch data is contained in an array with the expected patch size.
+
+    If `coords` are negative, the start of the patch will be padded with zeros up until
+    where the start of the image would be, and this is where the patch data starts.
+
+    If the `coords + patch_size` are greater than the bounds of the image then the
+    end of the patch will be filled with zeros.
+
+    Parameters
+    ----------
+    coords : Sequence[int]
+        The coordinates that describe where the patch starts in the spatial dimension of
+        the image
+    patch_size : Sequence[int]
+        The size of the patch in the spatial dimensions.
+    data_shape : Sequence[int]
+        The shape of the image the patch originates from, must be in the format SC(Z)YX.
+    patch_data : NDArray[T]
+        The patch data to be padded.
+
+    Returns
+    -------
+    NDArray[T]
+        The resulting padded patch.
+    """
+    coords_ = np.array(coords)
     patch = np.zeros((data_shape[1], *patch_size), dtype=patch_data.dtype)
-    patch_start = np.clip(np.array(coords), 0, None) - np.array(coords)
-    patch_end = np.array(coords) + np.array(patch_size)
-    patch_end = np.clip(patch_end, None, np.array(data_shape[2:])) - np.clip(
-        np.array(coords), None, np.array(data_shape[2:])
-    )
+    # data start will be zero unless coords are negative
+    data_start = np.clip(coords_, 0, None) - coords_
+    data_end = data_start + np.array(patch_data.shape[1:])
     patch[
         (
-            slice(None, None, None),
-            *tuple(slice(s, t) for s, t in zip(patch_start, patch_end, strict=False)),
+            slice(None, None, None),  # channel slice
+            *tuple(slice(s, t) for s, t in zip(data_start, data_end, strict=False)),
         )
     ] = patch_data
     return patch
--- a/src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
@@ -1,0 +1,28 @@
+from collections.abc import Sequence
+from typing import TypeVar
+
+import numpy as np
+from numpy.typing import NDArray
+
+T = TypeVar("T", bound=np.generic)
+
+
+def pad_patch(
+    coords: Sequence[int],
+    patch_size: Sequence[int],
+    data_shape: Sequence[int],
+    patch_data: NDArray[T],
+) -> NDArray[T]:
+    patch = np.zeros((data_shape[1], *patch_size), dtype=patch_data.dtype)
+    patch_start = np.clip(np.array(coords), 0, None) - np.array(coords)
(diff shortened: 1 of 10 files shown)


src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    patch_size: Sequence[int],
+    data_shape: Sequence[int],
+    patch_data: NDArray[T],
+) -> NDArray[T]:
+    patch = np.zeros((data_shape[1], *patch_size), dtype=patch_data.dtype)
+    patch_start = np.clip(np.array(coords), 0, None) - np.array(coords)
+    patch_end = np.array(coords) + np.array(patch_size)
+    patch_end = np.clip(patch_end, None, np.array(data_shape[2:])) - np.clip(
+        np.array(coords), None, np.array(data_shape[2:])
+    )
```
jdeschamps: I don't think I understand what is happening here without running the code. Could you add some comments and function docstring?

src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
jdeschamps, line 1: Can you add tests for the padding function? I think with the inputs it will help understand it better.

src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
jdeschamps: Am I correct in assuming that the padding will always happen on the patch end?

Because looking at the random patching, the coordinates are bound to be positive. So we know that `patch_data` is smaller or equal in shape to `patch_size`, why not simply slice `patch` from 0 to `patch_data.shape`?

src/careamics/dataset_ng/patching_strategies/tiling_strategy.py
```diff
@@ -148,8 +148,11 @@ def _compute_1d_coords(
                 coords.append(i)
                 crop_coords.append(0)
                 stitch_coords.append(0)
-                crop_size.append(tile_size - overlap // 2)
-            elif (i > 0) and (i + tile_size < axis_size):
+                if axis_size <= tile_size:
+                    crop_size.append(axis_size)
+                else:
```
jdeschamps, line 153: In the case of tiling, we don't expect tiles to go over the shape of the data, therefore the only case to tackle here is when the data itself is smaller than the tile. Is that a correct assumption?

jdeschamps (changes requested):
I am probably missing something on the padding function! Once that is resolved, and with padding tests i think this is good to go.

I imagine that the overhead of the padding with numpy is minimum, and we are not slowing down loading patches too much, even if it is applied systematically? would it be worth benchmarking it?

src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
melisande-c: Sorry I forgot the docstring, and wasn't very clear

So explanation:
- This padding function ensures that the returned patch will always have the same patch size, even if the patch is sampled from the edge of an image, pixels that are outside the data range of the image are set to zero.
- In the NG Dataset schema the patch coords refer to the top left of the patch in image space.
- The code works for the cases that:
  1) The coords are negative – the patch is sampled before the start of the image.
  2) The `coords + patch_size > data_size` – the edge of the patch overshoots the edge of the image.
  3) The case that both i & ii are true simultaneously.
- What I want to find and have called (maybe not very clearly) `patch_start` and `patch_end` is the extent of the data in the patch space.
- e.g. if a patch is sampled at the edge of an image and overshoots the edge by 4 pixels, the "start" would be 0 and the "end" would be (`patch_size - 4`), conversely if a patch is sampled 4 pixels before the image start (the coord would be -4) then the "start" would be 4 and the "end" would be `patch_size`.

src/careamics/dataset_ng/patching_strategies/tiling_strategy.py
melisande-c: This will only happen in the case there is 1 tile and it's size is larger than the axis size

melisande-c: And we need to crop it back to the data size afterwards.

src/careamics/dataset_ng/patch_extractor/image_stack/utils.py
melisande-c: And while currently we have no use of negative coordinates elsewhere in the codebase I made the function generalise to anywhere outside of the image range because it was 5 mins extra work, it makes the behaviour more consistent and it's also how the CziImageStack behaves.

jdeschamps (approved):

PR #594, opened 2025-10-22
Feat: Generalizes `ZarrImageStack`

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Zarr files can be arbitrarily organized, this PR makes the `ZarrImageStack` more general and compatible with latest zarr-python.

This PR extends `ZarrImageStack` capabilities to arbitrary organization, and adds a an `image_stack_loader` of sorts that follows a simple heuristics:
- if we supply a path to a `.zarr` file, then either we instantiate all arrays found in the file, or it is an OME-Zarr and we load only one resolution level
- if we supply URIs (`file://path/to/image.zarr/group/other_group/array`), then only the arrays specified by the URIs are instantiated.

The scope of the PR is solely on the `ImageStack` mechanism, and will be followed by other PRs:
- Compatibility with dataset: currently the default mechanism does not allow for passing paths other that those ending with the correct file extension.
- Prediction writer to Zarr

**Important notes and discussion**:
- One important change is that `zarr.open` is not called anymore in `ZarrImageStack`, which contradicts the spririt of the NGDataset. 
    - The reason for this choice is that in order to collect groups and arrays from a Zarr, we need to open it first... We could check how expensive this operation is, and choose to open, list arrays and then pass on the URIs for the `ZarrImageStack` to open again the file.
    - The second reason is that regardless of the type of storage, new Zarr docs basically encourage to use `zarr.open` which returns a `Group` (and not a store)
- A subsequent PR could also add an `in_memory` keyword to the `ZarrImageStack` for loading. But that would probably `ZarrImageStack` to be in charge of opening the file, which may lead to loading of all arrays. Alternatively, it could create a memory store and add the existing array, but that would require copying data. I don't know if there is a way to load in memory an array directly, without "accessing" it.
- I have not found a solution for the moment to be able to explore all arrays in a remote file (in the tests, `group.keys()` and `group.array_keys() hang).
- URI and OME-Zarr validation are using a very minimal set of constraints, which may not hold at all time.
- `create_zarr_image_stacks` could be refactored further, and `zarr_utils` placed somewhere more meaningful

The current `ZarrImageStack` is only compatible with single-array OME-Zarr with `"0"` resolution level. In the future, we want to be flexible since Zarr comes in many flavors, and in particular support having multiple arrays in a single zarr.

### Overview - what changed?

Zarr image stack loader can instantiate multiple `ZarrImageStack` for a single zarr, and allows specifying different arrays in various zarr files as input.

### New features or files

- `zarr_utils` contains all the functions interacting with the Zarr format

- `ZarrImageStack` has been adapted to take a `Group` as input, and now retains `chunk_size` and different `source` (URI-like) for future PRs

### Removed features or files

- all old OME functions.

## How has this been tested?

Pytest

## Breaking changes

All Zarr related code/notebook.

---

- [x] Pre-commit passes


src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    ----------
+    path_to_zarr : str
+        The path to the zarr store.
+
+    Returns
+    -------
+    str
+        The modified path with '_seg' added before the '.zarr' extension.
+    """
+    return path_to_zarr[:-5] + "_seg.zarr"
```
melisande-c, line 136: Btw why did you choose `_seg` for the prediction suffix? Is it like a standard key? Bec obvs N2V results are not segmentations

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
```diff
@@ -69,30 +72,42 @@ def create_tiff_extractor(
...
 
 
-# ZARR case
-def create_ome_zarr_extractor(
-    source: Sequence[Path],
+# # OME-ZARR case
+# def create_ome_zarr_extractor(
+#     source: Sequence[Path],
+#     axes: str,
+# ) -> PatchExtractor[ZarrImageStack]:
```
melisande-c, line 79: delete commented out function

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    ------
+    ValueError
+        If the source string does not start with "file://".
+    ValueError
+        If the source string does not contain a ".zarr" file extension.
+    """
+    key = "file://"
+
+    if source[: len(key)] != key:
+        raise ValueError(f"Remote file not supported: {source}")
```
melisande-c, line 105: Are we not supporting remote just because there's a question on where to save results?

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
```diff
@@ -17,82 +14,33 @@ class ZarrImageStack:
...
-        self._store = store
-        self._array = zarr.open_array(store=self._store, path=data_path, mode="r")
+    def __init__(self, group: zarr.Group, data_path: str, axes: str):
+        if not isinstance(group, zarr.Group):
+            raise TypeError(f"group must be a zarr.Group instance, got {type(group)}.")
+
+        self._group = group
+        self._array = group[data_path]
+        self._store = str(group.store_path)
+        self._source = self._array.store_path
```
melisande-c, line 24: I'm not sure `zarr.Array.store_path` is a guaranteed attribute, I quickly checked when `Group` has as a `MemoryStore` and there is an error, this is why we only accepted 2 types of store before.
```python
store = zarr.storage.MemoryStore()
group = zarr.group(store, path="group")
group["array"] = zarr.array(np.arange(32*32).reshape(32, 32))
print(group["array"])
print(group["array"].store_path)
```
```
<zarr.core.Array '/group/array' (32, 32) int64>
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[13], [line 5](vscode-notebook-cell:?execution_count=13&line=5)
      3 group["array"] = zarr.array(np.arange(32*32).reshape(32, 32))
      4 print(group["array"])
----> [5](vscode-notebook-cell:?execution_count=13&line=5) print(group["array"].store_path)

AttributeError: 'Array' object has no attribute 'store_path'
```

melisande-c (commented):
Looks good! Just a couple of questions/notes

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: Ahah that's silly. I think the only answer is because when I am tired I go back to old habits....

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
melisande-c: maybe we can just save in the `Careamist.work_dir` by default

jdeschamps: I didn't want to complicate too much, but I can't think of any technical reasons why it would not work for remote. This was mostly meant to avoid in memory stores for now (see other answer).

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
jdeschamps: Ah interesting. Earlier I had `self._source = group.store_path + "/" + data_path` which should generalize.

That being said, I wanted to avoid in memory stores for now. The main reason being that, unless I am mistaken, we can't pass a memory store by `str | Path`.

jdeschamps: In which case would we have `MemoryStore` as input data? I am failing to see a good use case for creating a memory store before training (as opposed to numpy arrays).

For the sake of argument:
- does it make sense to load them as array? convert to numpy and pass with `data="array"`?
- accept `MemoryStore` as an input along side `NDArray` and `str | Path`? But then how do we select which array to use? `collect` vs `ome` as done in this PR?
- alternatively, if people want in memory training:
    - have our `InMemoryStack` also load zarr?
    - have a `in_memory: bool` parameter in `ZarrImageStack` and create a `MemoryStore`? Can we create a `MemoryStore` from a `LocalStore` or fsspec store?

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
melisande-c: fair enough, we can think about it later

src/careamics/dataset_ng/patch_extractor/patch_extractor_factory.py
jdeschamps: Deleted.

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: Renamed to "out". Will only be used in the next PR though, it should not have been committed here...

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
melisande-c: Oooh I just realised I was accidentally in environment with zarr 2

melisande-c: Now for the same (slightly updated for v3) code I get
```
<Array memory://4719444096/group/array shape=(32, 32) dtype=int64>
memory://4719444096/group/array
```

Ok I think everything is fine as it is as checking the path is file happens in the constructor functions

jdeschamps: > I'm not sure `zarr.Array.store_path` is a guaranteed attribute, I quickly checked when `Group` has as a `MemoryStore` and there is an error, this is why we only accepted 2 types of store before.
> 
> ```python
> store = zarr.storage.MemoryStore()
> group = zarr.group(store, path="group")
> group["array"] = zarr.array(np.arange(32*32).reshape(32, 32))
> print(group["array"])
> print(group["array"].store_path)
> ```
> 
> ```
> <zarr.core.Array '/group/array' (32, 32) int64>
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> Cell In[13], [line 5](vscode-notebook-cell:?execution_count=13&line=5)
>       3 group["array"] = zarr.array(np.arange(32*32).reshape(32, 32))
>       4 print(group["array"])
> ----> [5](vscode-notebook-cell:?execution_count=13&line=5) print(group["array"].store_path)
> 
> AttributeError: 'Array' object has no attribute 'store_path'
> ```

When running your code I get the following error:

```python
----> 3 group["array"] = zarr.array(np.arange(32*32).reshape(32, 32))
TypeError: arr argument must be numpy or other NDArrayLike array
```

While this runs:
```python
store = zarr.storage.MemoryStore()
group = zarr.group(store, path="group")
arr = group.create_array("array", shape=(10, 10), dtype="uint16")
arr[...] = np.arange(100).reshape(10, 10)

print(group["array"])
print(group["array"].store_path)
```
```python
<Array memory://5652457472/group/array shape=(10, 10) dtype=uint16>
memory://5652457472/group/array
```

I have `zarr==3.1.3`.

melisande-c: ^ yep I was accidentally using zarr 2, I updated it and did something similar to you and then it worked

melisande-c (approved):
Everything looks good! The memory store thing was just bec I was accidentally using zarr v2, since the the constructor functions already handle that it has to be a file I think we can leave it as is.

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
jdeschamps: Ah sorry I didn't see your messages when writing! Thanks!

jdeschamps:
Thanks!

PR #596, opened 2025-10-23
Feature(Next-Gen Dataset): Iterative file loading 

### LINKED ISSUE #406: Next-Gen Dataset: Lazy file loading

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Adding iterative tiff loading to the NG Dataset. New components are `FileImageStack`, `LimitFilesPatchExtractor` and `GroupedIndexSampler` which must be all used in conjunction so that each image file is opened iteratively. These components will be initialised in the `datamodule` if the data type is `TIFF` and `use_in_memory=False`.
 
N.B. Should work in principle but I haven't tested on a large dataset.

This will emulate the previous iterable Careamics dataset while still being the mapped version. If a user has many files that cannot all be loaded into memory this solution might be more convenient than converting all their data to zarr. 

### Overview - what changed?

Components have been added that will allow for iterable file loading. The patching strategies have been modified to have an additional method that will return the indices belonging to a particular dataset. Convenient factory have been added. The dataloaders will be initialised with all the components in the datamodule depending on the config parameters. 

My solution requires 3 classes to work in sync. 
- We have the new `FileImageStack` that will not load the data on initialisation but only when it's `load` method is called. It also has a `close` method that removes the reference to the data so it will get cleared up by the garbage collector.
- Then there is the `LimitFilePatchExtractor` which makes sure that there is never more than one file loaded. 
- Finally there is the `GroupedIndexSampler`, this makes sure that indices from the same file will be sequential (but the order of the files and the order of the indices within a file are shuffled). This is important because otherwise, if the indices were randomly shuffled, the `LimitFilesPatchExtractor` would be opening and closing the same file many times, slowing down training.

### New features or files

- `FileImageStack`
- `LimitFilesPatchExtractor`
- `GroupedIterSampler` 

- Patching strategies (added method `get_patch_indices`)

## How has this been tested?

Added relevant tests for each component, tests for the datamodule that ensure each file is only loaded once during data iteration.

## Related Issues

- Resolves #406 

1. If there are multiple workers, each worker will receive a clone of the entire dataset, therefore the amount of memory taken up will be multiplied by the number of workers. I'm not entirely sure how PyTorch handles multiple workers for the IterableDataset but it is probably similar so this probably would have been the case for the old dataset anyway.
2. I haven't added any automatic detection of memory usage.

---

- [x] Pre-commit passes


tests/dataset_ng/test_grouped_index_sampler.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+
+from careamics.config.configuration_factories import _create_ng_data_configuration
+from careamics.dataset_ng.dataset import Mode
+from careamics.dataset_ng.factory import create_array_dataset
+from careamics.dataset_ng.grouped_index_sampler import GroupedIndexSampler
+
+
+def _assert_indices_grouped(sampler: GroupedIndexSampler):
+    """Assert the indices from the sampler a grouped as expected."""
```
jdeschamps:
```suggestion
    """Assert the indices from the sampler are grouped as expected."""
```

jdeschamps (approved):
Looks great, thanks!

I have no comments... The only thing I could think of is an existing todo with the parameterized number of loaded file, but this can be another PR someday.

Could you open an issue on the documentation to remind us to document the potential memory bloating with multiple workers? Obviously we will need to document the whole iterative tiff loading, but it would be good not to forget that part.

Issue #407, opened 2025-02-17
Next-Gen Dataset: List of MicroSplit requirements


melisande-c:
We can use this issue to list the requirements that the MicroSplit algorithms need for data loading, and discuss how they can be integrated.

#### Requirements
- Lateral context loading
- Training data synthesis, i.e. channel mixing, including mixing one populated channel with one empty channel.

melisande-c:
After discussion with @CatEek, we decided it makes sense to have multiple `PatchExtractor` classes. There will be an `LCPatchExtractor` that will return the concatenated LC inputs for a patch specified with `PatchSpecs`. A question that came up was: how are patches at the edge of an image currently handled in MicroSplit: with mirror padding? zero padding? Or not sampled?

melisande-c:
## MicroSplitDataset Proposal

I believe it makes sense to create a separate dataset class for MicroSplit because of the different requirements during training. The dataset can still use the same components as the CareamicsDataset but with some additional functionality. Once the dataset has been developed, if we see it is not so different to the original dataset after all, we can think about merging it with the CareamicsDataset.

### Handling channel patch mixing during training as a data augmentation.

I think we can handle this by having `N` patching strategy instances for `N` channels during training. So patches located at different spatial locations can be extracted from each channel and then combined in the dataset. We can also take advantage of the background filtering for each channel separately if that is required for mixing empty and non empty patches.

In the case we do not want to mix spatially uncorrelated patches we will only have 1 patching strategy instance.

Additionally if someone has a real target (as opposed to synthetically mixed channels), this can be given to the dataset and there should also only be 1 patching strategy in this case for the input and target.

### Handling lateral context inputs

The lateral context creation should definitely happen at the PatchExtractor level, however, since #596 there are now two patch extractors, the regular `PatchExtractor` and the `LimitFilesPatchExtractor` for iterative file loading. We want MicroSplit to also take advantage of iterative file loading so I propose we inject a patch construction function into the patch extractors. We will have the "basic" patch construction that we currently have and the "lateral_context" patch construction for MicroSplit with LC. It will look something like:

```python
def basic_patch_constr(
    image_stack: ImageStack,
    sample_idx: int,
    coords: Sequence[int],
    patch_size: Sequence[int],
):
    return image_stack.extract_patch(
        sample_idx=sample_idx, coords=coords, patch_size=patch_size
    )


def lateral_context_patch_const(
    image_stack: ImageStack,
    sample_idx: int,
[...]

PR #612, opened 2025-11-06
Feature(Next-Gen Dataset): Lateral context patch construction for MicroSplit

### LINKED ISSUE #607: Next-Gen Dataset(MicroSplit): Lateral Context Patch Construction
MicroSplit requires lateral context patch construction for training and inference. A solution to integrate it into the Next-Gen Dataset framework would be to inject a patch construction function into the patch extractors as outlined in https://github.com/CAREamics/careamics/issues/407#issuecomment-3490510392.

melisande-c:
## Description

> [!NOTE]  
> **tldr**: This PR adds support for lateral context patch construction as the first step in implementing data loading requirements for MicroSplit.

The Next-Gen Dataset framework needs to be used by all algorithms supported by CAREamics, including MicroSplit, which requires lateral context patch construction.

### Overview - what changed?

The `PatchExtractor` classes are now injected with a `PatchConstructor` method; there are two implemented patch construction methods: the "basic" constructor which has the original behaviour, and the "lateral context" constructor which creates patches with a lateral context dimension. The correct patch constructor will be selected at dataset initialisation according to config parameters (future dev), this will require refactoring of all the patch extractor factory functions. 

The reason to use injection here is that, since #596, there are two patch extractors: the regular extractor and now the `LimitFilesPatchExtractor`. We want MicroSplit to be able to take advantage of the iterable file loading mechanism, so this way both extractors can produce lateral context patches, and we don't have a subclass explosion.

The basic constructor is the default, this way the rest of the code doesn't have to change.

Because the lateral context construction is also parametrised — it needs to know how many lateral context levels to create and what padding mode to use — I implemented this as a closure.

The lateral context construction itself simply: 
1. iterates through the lateral context scales, 
2. calculates the lateral context patch size (`patch_size * 2^scale`), 
3. calculates the lateral context patch start and end
4. clips the lateral context to the boundary of the data (we don't use the image stack padding, introduced in #587, because we want to use padding modes),
5. pads the lateral context with the chosen mode using `numpy.pad`
6. resizes the lateral context with `skimage.transform.resize`
7. stacks the LC patches and returns

### New features or files

- `PatchConstructor` protocol
- `basic_patch_constr` function
- `lateral_context_patch_constr` function

- `PatchConstructor`
- `LimitFilesPatchConstructor`

## How has this been tested?

Two new tests added:
- `test_lateral_context_constructor`
- `test_patch_extractor_lc_injection`

## Related Issues

- Resolves #607 
- This is part of the proposal outlined in https://github.com/CAREamics/careamics/issues/407#issuecomment-3490510392

1. The original LC dataset [creates all the LC scaled data at initialisation](https://github.com/CAREamics/careamics/blob/e640e3a74b5895c1ed90e1f7442a530140ee4605/src/careamics/lvae_training/dataset/lc_dataset.py#L63-L82), this new implementation rescales each patch individually. This might mean that the new implementation is slower, however creating each scale at the start won't be compatible with very large data, and it won't be compatible with zarr loading, which is why I went for the on fly scaling.
2. I used `skimage.transform.resize` because that's what the original dataset uses but The [sci-kit image docs](https://scikit-image.org/docs/0.25.x/api/skimage.transform.html#skimage.transform.resize) suggest using `downscale_local_mean` for integer downscaling. And we might want to benchmark other methods in the future such as max pooling?
3. There is this note in the original dataset which I did not implement https://github.com/CAREamics/careamics/blob/e640e3a74b5895c1ed90e1f7442a530140ee4605/src/careamics/lvae_training/dataset/lc_dataset.py#L70

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/dataset_ng/patch_extractor/patch_construction.py
+++ b/src/careamics/dataset_ng/patch_extractor/patch_construction.py
@@ -114,17 +114,17 @@ def lateral_context_patch_constr(
             pad_after = lc_end - end_clipped
             pad_width = np.concat(
                 [
+                    # zeros to not pad the channel axis
                     np.zeros((1, 2), dtype=int),
                     np.stack([pad_before, pad_after], axis=-1),
                 ]
             )
             lc_patch = np.pad(
-                # zeros to not pad the channel axis
                 lc_patch,
                 pad_width,
                 mode=padding_mode,
             )
-            # TODO: test different downscaling, could try max pooling?
+            # TODO: test different downscaling? skimage suggests downscale_local_mean
             lc_patch = resize(lc_patch, (n_channels, *patch_size))
             patch[scale] = lc_patch
         return patch
--- a/src/careamics/dataset_ng/patch_extractor/patch_construction.py
+++ b/src/careamics/dataset_ng/patch_extractor/patch_construction.py
@@ -9,6 +9,7 @@ from .image_stack import ImageStack
 
 
 class PatchConstructor(Protocol):
+    """A patch constructor function creates a patch from a given ImageStack."""
 
     def __call__(
         self,
@@ -16,7 +17,26 @@ class PatchConstructor(Protocol):
         sample_idx: int,
         coords: Sequence[int],
         patch_size: Sequence[int],
-    ) -> NDArray[Any]: ...
+    ) -> NDArray[Any]:
+        """
+        Parameters
+        ----------
+        image_stack: ImageStack
+            The image stack to construct a patch from.
+        sample_idx: int
+            Sample index. The first dimension of the image data will be indexed at this
+            value.
+        coords: Sequence of int
+            The coordinates that define the start of a patch.
+        patch_size: Sequence of int
+            The size of the patch in each spatial dimension.
+
+        Returns
+        -------
+        numpy.ndarray
+            The patch.
+        """
+        ...
 
 
 def basic_patch_constr(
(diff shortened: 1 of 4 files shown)


src/careamics/dataset_ng/patch_extractor/patch_construction.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Returns
+        -------
+        numpy.ndarray
+            The patch.
+        """
+        ...
+
+
+def basic_patch_constr(
```
jdeschamps, line 42: Just a suggestion change `basic` to `default`.

`default_patch_constr`?

src/careamics/dataset_ng/patch_extractor/patch_construction.py
```diff
@@ -0,0 +1,132 @@
...
+        coords: Sequence[int],
+        patch_size: Sequence[int],
+    ) -> NDArray[Any]:
+        shape = image_stack.data_shape
+        spatial_shape = shape[2:]
+        n_channels = shape[1]
+
+        # There will now be an additional lc dimension,
+        # this has to be handled correctly by the dataset
+        # TODO: maybe we want to limit this constructor to only images with 1 channel
```
jdeschamps, line 93: Hmmm technically MicroSplit does not accept channels, so that could be a working assumption here. However, the LVAE module might be shared with HDN, which should accept multi-channel, and LambdaSplit, which definitely has many channels.

MicroSplit will have its own LightningModule, so it might not be such an issue.

src/careamics/dataset_ng/patch_extractor/patch_construction.py
```diff
@@ -0,0 +1,132 @@
...
+
+import numpy as np
+from numpy.typing import NDArray
+from skimage.transform import resize
+
+from .image_stack import ImageStack
+
+
+class PatchConstructor(Protocol):
+    """A patch constructor function creates a patch from a given ImageStack."""
```
jdeschamps, line 12: Can we expand the description to motivate the use of this intermediary class?

jdeschamps (approved):
I did not expect it to be that easy to solve, elegant solution as well! Maybe we should add a bit more documentation details to be able to more easily navigate the various modules, but that is out of scope here.

src/careamics/dataset_ng/patch_extractor/patch_construction.py
melisande-c: Yeah exactly my todo here is: should this function raise an error if the input C > 1 and then we can put the LC and then we can return patches with dimensions `(L, (Z), Y, X)`? But then I was thinking maybe it's less ambiguous to return `(L, C=1, (Z), Y, X)`, which can easily be handled by further processing in the `MicroSplitDataset`. More of a stylistic thing I guess. I would be in favour of leaving it how it is for now and deciding when more of the dataset is fleshed out

jdeschamps: I guess it will come down to: do we need 50 additional lines of code to handle the `L` dimension in the dataset, in which case it may be worth using the `C`, or is it straightforward and we leave it as it is.

PR #616, opened 2025-11-10
Feat: Enable Zarr CAREamics DataModule

### LINKED ISSUE #598: Next-Gen Dataset: do not list files if zarr internal paths
## Problem

For the Zarr internal paths (#594) to work as input in the NGDataset, the `Dataset` needs to detect that these are not file paths (for which is does `.glob("*.zarr")` on the paths) and consume them by calling the zarr image stack loader.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Allow passing paths and URIs of Zarrs to the CAREamics Lightning Datamodule.

Currently, the CAREamics Lightning Datamodule can receive arrays, string, paths or list of the same types. When it comes to zarr, the current design decision is to allow passing paths to groups inside a zarr and collect all arrays that may be stored there. This is not possible with the current path / file solver.

This PR brings the following changes:
- Refactor the array/path/file solving logic, move it to `data_module_utils.py`, and add tests
- Add logic to solve string, path or URIs to Zarr groups and arrays
- Fix some bugs along the way that appeared when testing Zarr with the data module

> [!NOTE]  
> Currently, here are potential pain points or blind angles, which we could address in subsequent PR:
> - Matching names between input and target is based solely on sorting the names of the arrays, not by checking that they agree
> - Online zarr and OME have not been tested from the DataModule level (only at the `ZarrImageStack` level)
> - Passing a mixed list of path to zarr files and URIs will result in only the type of the first element to be retained. I don't like that, but I didn't want to make the function more complex for now. This should be addressed in a subsequent PR (or now if that is the reviewer's opinion).
> -The existence of the arrays or groups corresponding to the URIs is not checked (so as to not open the zarr file)

In order to use URIs (`file://data/folder/my_images.zarr/a_group/another_group`), which allow a high level of granularity and flexibility in defining which arrays should be used for training/validation/prediction, the logic of paths solving in the `CAREamicsDataModule` needed to take into account this new possibility.

More specifically, this PR aims at enabling the following behaviours:

```python
path_to_zarr = ["path/to/zarr/archive.zarr"] # load all arrays in the root
path_to_ome_zarr = ["path/to/zarr/ome_archive.zarr"] # load all arrays in the first level ("0" if not specified)
path_to_folder = ["path/to/folder/"] # load all arrays from the root of all zarr file in the folder
path_to_group = ["file://path/to/zarr/archive.zarr/group/subgroup"] # load all arrays from this group (obtained with g.store_path)
path_to_arrays = [
    "file://path/to/zarr/archive.zarr/group/array1",
    "file://path/to/zarr/archive.zarr/group/array2",
    "file://path/to/other_zarr/other_archive.zarr/other_group/array3",
] # load all specified arrays
```

### Overview - what changed?

- `CAREamicsDataModule` accepts `zarr` data type and offload solving the paths/URIs to a new function in `data_module_utils.py`

As for the arrays, string and Path, the `zarr` datatype triggers a series of checks:
- If `str` or `Path`, check if it corresponds to an existing `Path`, then solve it using the normal path logic
- If `str` but not an existing `Path`, check if it is a URI, then solve it
- If a list, repeat the same as above, but for a lists, based on the first element of the list

### New features or files

- Moved all path solving feature to `data_module_utils.py`, which now also contains Zarr logic, and refactor them a little

- Changed the behaviour of the array collection in Zarr to limit itself to the arrays in the group (rather than recursively go through the sub-groups).

### Removed features or files

- Removed logic from the data module.

## How has this been tested?

- New tests have been implemented.
- BSD68 notebook tested for small iterations

- Resolves #598 

## Future improvements

- Currently one aspect is not great, but I decided not to make it more complex for now: when inspecting a list of inputs, only the first element is checked. So if the first element is a path to a zarr file, it will only consider existing files. If it is a URI, it will only consider URIs. There is no reason (except more complex code) to not just accept all.
- Matching names of arrays / files for input/target solving.
- Checking existence of the arrays in the URI.

---

- [x] Pre-commit passes


src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
```diff
@@ -1,36 +1,45 @@
...
+INPUT = str | Path
 
-# TODO convenience function to return source path?
 
-
-def is_file_uri(source: str) -> bool:
-    """Check if a source string is a file URI.
+def is_zarr_uri(path: str | Path) -> bool:
+    """
+    Check if a path is a Zarr URI.
```
melisande-c, line 15: It looks like this function just returns if this is in the valid uri schemes, not that it is a zarr file, maybe a name something like `is_valid_uri` is better?

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
```diff
@@ -114,26 +110,10 @@ def decipher_zarr_path(source: str) -> tuple[str, str, str]:
...
-    path_to_zarr : str
-        The path to the zarr store.
-
-    Returns
-    -------
-    str
-        The modified path with '_seg' added before the '.zarr' extension.
-    """
-    return path_to_zarr[:-5] + "_seg.zarr"
+    return "/".join(path_to_zarr), "/".join(parent_path), content_path
```
melisande-c, line 116: Will using "/" as a delimiter also be correct for windows? You use `os.path.join` to be safe or get `os.sep`

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
```diff
@@ -250,26 +230,48 @@ def create_zarr_image_stacks(
...
                     ZarrImageStack(group=zarr_group, data_path=array_path, axes=axes)
                 )
 
-        elif is_file_uri(data_str):
+        elif is_zarr_uri(data_str):
             # decipher the uri and open the group
-            store_path, group_path, array_name = decipher_zarr_path(data_str)
+            store_path, parent_path, name = decipher_zarr_path(data_str)
+
+            zarr_group = zarr.open(store_path, mode="r")[parent_path]
```
melisande-c:
```suggestion
            zarr_group = zarr.open_group(store_path, path=parent_path, mode="r")
```
Pylance on strict is giving me type errors otherwise

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
melisande-c:
```suggestion
            zarr_group = zarr.open_group(data_str, mode="r")
```

src/careamics/dataset_ng/dataset.py
melisande-c, line 98: Adding validation here is probably wrong. This error is about not padding patches during training and validation. The validation statistics might be weird if it's including areas padded with zeros.

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -324,12 +324,19 @@ def __init__(
             raise ValueError(
                 "At least one of train_data, val_data or pred_data must be provided."
             )
+        elif train_data is None != val_data is None:
+            raise ValueError(
+                "If one of train_data or val_data is provided, both must be provided."
+            )
```
melisande-c, line 330: What if you don't want to have validation

melisande-c (commented):
Some small things/questions! I didn't fully look through the validation in `data_module_utils` but it makes sense to me to move these functions into a utils module, so we can more clearly see the core functionality of the data module.

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
melisande-c: um not sure why the suggested change is showing original line 212 as empty

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: I like it! Changed it.

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: Interestingly it does not seem to be an issue. Pathlib converts to Windows paths under the hood, and `zarr` seems to handle it correctly. The tests are passing on Windows, so I am not too worried. 

The returned "canonical" path for the store is the following (just a representation though): 
``` python
from pathlib import Path
import zarr
import numpy as np

g = zarr.open(Path(".") / "my_zarr.zarr")
group = g.create_group("data")
array = group.create_array(
        name="array",
        data=np.ones((5, 32, 32)),
        chunks=(1, 16, 16)
)

```
``` python
>>> Store path: file://my_zarr.zarr/data/array
```

Using `zarr.open`:
```
path = "my_zarr.zarr/data/array"
g2 = zarr.open(path)
print(f"Opened: {g2}")
```
``` python
>>> Opened: <Array file://my_zarr.zarr/data/array shape=(5, 32, 32) dtype=float64>
```

Accessing internal paths using `zarr`:
``` python
a2=g["data/array"]
print(f"Got {a2}")
```
``` python
>>> Got <Array file://my_zarr.zarr/data/array shape=(5, 32, 32) dtype=float64>
```

Path using `pathlib`:
``` python
p = Path(path)
print(f"Path exists: {p.exists()} and is {type(p)}")
```
``` python
>>> Path exists: True and is <class 'pathlib._local.WindowsPath'>
```

Now, using `os.path.join` in `zarr` also works:
``` python
ps = ["data", "array"]
in_pth = join(*ps)
print(f"os.path.join: {in_pth}")
a3=g[in_pth]
print(f"Got {a3}")
```
``` python
os.path.join: data\array
Got <Array file://my_zarr.zarr/data/array shape=(5, 32, 32) dtype=float64>
```

melisande-c: Ah I only just noticed in the docs:

https://github.com/CAREamics/careamics/blob/97ca2c4a6ad7da73a8783107f55203f95818472e/src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py#L72-L73

If the input is expected to be in the uri format then it makes sense to also output with the "/" delimiter

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
melisande-c: Also we don't have zarr as a mypy additional dependency in the .pre-commit-config.yaml so we're not checking the correctness of zarr types in the CI at the moment

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: I tried to use `os.path.join` but it seems to have one difference with `"/".join`, is that the leading separator seems to be important for the tests to pass. But only for the `store_path`.

jdeschamps: Added that:

``` python

    path_to_zarr_j = os.path.join(os.sep, *path_to_zarr)
    parent_path_j = (
        os.path.join(*parent_path) 
        if parent_path != [] 
        else ""
    )

    return path_to_zarr_j, parent_path_j, content_path
```

Not pretty but it should do the job.

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: > Also we don't have zarr as a mypy additional dependency in the .pre-commit-config.yaml so we're not checking the correctness of zarr types in the CI at the moment

Ah good point! I added it.

src/careamics/lightning/dataset_ng/data_module.py
jdeschamps: It will fail upon calling `setup` because the hypothesis here is that there is always validation:

https://github.com/CAREamics/careamics/blob/836c9f5410849164cd02b79e550b74665ca087cf/src/careamics/lightning/dataset_ng/data_module.py#L650-L680

melisande-c: ah I see, ok lets's leave that as it is then

src/careamics/dataset_ng/patch_extractor/image_stack/image_utils/zarr_utils.py
jdeschamps: I backtracked from that. I kept having issues in the tests and in the VDI, depending on the OS with or without separator at the beginning of the tests....

So letting `pathlib` and `zarr` figure it out simplifies the code and make the test work, as well as direct tests in the VDI.

melisande-c: makes sense! Sorry for the time wasted

jdeschamps: Not at all, it made sense.

melisande-c (approved):

PR #617, opened 2025-11-11
Feature(Next-Gen Dataset): MicroSplit Input and Target Generation (Proof of Principle)

### LINKED ISSUE #613: Next-Gen Dataset(MicroSplit): Input synthesis
MicroSplit has the option to synthesis the input from the target by summing together input channels. There are a few options in doing this which essentially behave as data augmentation.

#### Option 1: basic
Simply sum together the channels of a patch from the data, with given weights, and a possible option to randomise the weights.

#### Option 2: Mixing channels from different spatial locations
Patches from different spatial locations are selected for each channel and mixed together. This is essentially data augmentation since it gives the model more examples. Note, that one probably wouldn't want to use this if it is not biologically plausible, i.e. spatial feature in each channel are correlated, e.g. nuclei and cell membranes.

#### Option 3: Setting channels to empty patches
This is an extension of option 2, but only one channel has a patch containing structure and the rest are empty. This should give the model an opportunity to know what a single structure looks like.

Note that options 2 & 3 are only applied on a percentage of the data, according to some configurable probability.

---

@CatEek let me know if I missed anything

melisande-c:
## Description

> [!NOTE]  
> **tldr**: This PR implements MicroSplit input and target patch synthesis using the Next-Gen Dataset components. Please note this is a proof of principle and will need refactoring to integrate the functions into a Dataset class. 

The MicroSplit algorithm has the option to synthesis the input by summing the target channels, with some given weights. Optionally — to augment the data, if the channels are spatially uncorrelated — the channels of the target patch can come from different spatial locations. The content of the channels can also be chosen to be empty or have signal by using the patch filtering implemented in #532.

### Overview - what changed?

Added a new module `dataset_ng/microsplit_input_synth.py` containing the MicroSplit patching function. These will need refactoring in future development, possibly being converted to methods on a dataset class, but they demonstrate how the Next-Gen Dataset components can be used for MicroSplit patching.

New method `extract_channel_patch` has been added to the image stacks and patch extractor.

New functions in `microsplit_input_synth.py` are as follows:

- **`extract_microsplit_patch`**: This function makes sure the resulting patch always has the dimensions `LC(Z)YX` adding an additional L dimension where necessary (we have no knowledge whether the patch extractor is using LC patch construction - I find this a bit annoying but I don't have another solution right now). It can also accept one `PatchSpec` for spatially correlated channels or a list with one for each channel.
- **`get_random_channel_patches`**: gets patch specs and patches where each channel is from a random spatial location anywhere in the dataset.
- **`get_empty_channel_patches`** (needs to be renamed): whether a channel should have signal or be empty can be specified by providing a dict of `PatchFilterProtocol`. The function will randomly select patches until it finds a patch with signal / a patch that is empty or until the patience runs out. Only the primary input is checked for signal/emptyness.
- **`create_uncorrelated_input_target`**: this takes the results of `get_random_channel_patches` or `get_empty_channel_patches` and constructs input and target patches with metadata attached in an `UncorrelatedRegionData` named tuple.
- **`create_default_input_target`**: Creates input and target patches with metadata in an `ImageRegionData` named tuple (which is also used by the `CareamicsDataset`).

### New features or files

- new function `extract_microsplit_patch`
- new function `get_random_channel_patches`
- new function `get_empty_channel_patches`
- new function `create_default_input_target`
- new class `UncorrelatedRegionData`

- new method `extract_channel_patch` in `PatchExtractor`
- new method `extract_channel_patch` in `ImageStack`

## How has this been tested?

Tested in notebooks and scripts only so far, see gist in Notes section below.

## Related Issues

- Resolves #613 

## Future work / Potential issues

- If uncorrelated channel data comes from images with different spatial dimension then we need to create a separate patch extractor and patching strategy for each channel, these could be given a list to all the functions.
- `UncorrelatedRegionData` mirrors `ImageRegionData` but has a list instead of a single item for most of the metadata. This might be annoying later and we might have to come up with a better solution.
- Choosing which construction method is used based on probabilities set in the config.

The script in the gist below can be downloaded and run with `uv run demo_microsplit_patches_PoP.py` and it should produce figures similar to those attached below. This demonstrates the possible outputs of `get_empty_channel_patches` on the BioSR data (which does have correlated channels but this is just for demo purposes).

The gist also demonstrates the lateral context patch construction introduced in #612.

The patch filter thresholds where chosen empirically.

Example 3 is how the empty patch replacement worked in the original dataset.

https://gist.github.com/melisande-c/cdacdba731706c8ef89105f386891e0f

MicroSplit_patch_example1
MicroSplit_patch_example2
MicroSplit_patch_example3

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
@@ -20,11 +20,13 @@ class ZarrImageStack:
 
         self._group = group
         self._store = str(group.store_path)
-        self._array = group[data_path]
-        if not isinstance(self._array, zarr.Array):
-            raise TypeError(
+        try:
+            self._array = group[data_path]
+        except KeyError as e:
+            raise ValueError(
                 f"Did not find array at '{data_path}' in store '{self._store}'."
-            )
+            ) from e
+
         self._source = self._array.store_path
 
         # TODO: validate axes
--- a/src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
@@ -49,18 +49,9 @@ class ZarrImageStack:
     def extract_patch(
         self, sample_idx: int, coords: Sequence[int], patch_size: Sequence[int]
     ) -> NDArray:
-        return self._extract_patch(sample_idx, None, coords, patch_size)
+        return self.extract_channel_patch(sample_idx, None, coords, patch_size)
 
     def extract_channel_patch(
-        self,
-        sample_idx: int,
-        channel_idx: int,
-        coords: Sequence[int],
-        patch_size: Sequence[int],
-    ) -> NDArray:
-        return self._extract_patch(sample_idx, channel_idx, coords, patch_size)
-
-    def _extract_patch(
         self,
         sample_idx: int,
         channel_idx: int | None,  # `channel_idx = None` to select all channels,
--- a/src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
+++ b/src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
@@ -19,8 +19,12 @@ class ZarrImageStack:
             raise TypeError(f"group must be a zarr.Group instance, got {type(group)}.")
 
         self._group = group
-        self._array = group[data_path]
         self._store = str(group.store_path)
+        self._array = group[data_path]
+        if not isinstance(self._array, zarr.Array):
+            raise TypeError(
+                f"Did not find array at '{data_path}' in store '{self._store}'."
+            )
         self._source = self._array.store_path
 
         # TODO: validate axes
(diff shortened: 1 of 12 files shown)


src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
```diff
@@ -19,8 +19,12 @@ def __init__(self, group: zarr.Group, data_path: str, axes: str):
             raise TypeError(f"group must be a zarr.Group instance, got {type(group)}.")
 
         self._group = group
-        self._array = group[data_path]
         self._store = str(group.store_path)
+        self._array = group[data_path]
```
jdeschamps: I think this will actually raise an error directly. We should probably catch it here and make a better error message.

jdeschamps (approved):
Regarding tests for the input synthesis, are you waiting for actual integration since API can change?


[EDIT]: I forgot you made a gist to illustrate it.

src/careamics/dataset_ng/patch_extractor/image_stack/zarr_image_stack.py
melisande-c: hmm I don't remember changing this and am not sure why I did 🙈, I will have another look

melisande-c: Ok I updated it and added a test, see [74ab42b](https://github.com/CAREamics/careamics/pull/617/commits/74ab42b89e12de97b3433091680d353c49063ab6)

PR #620, opened 2025-11-17
Feat: Add prediction utilities and callbacks to NG Dataset

### LINKED ISSUE #549: Next-Gen Dataset: Zarr direct tile writing
## Problem

Prediction is currently done over all tiles and then written to file. For Zarr in particular, we need to be able to write tile by tile (or batche by batch) directly into the file.

## Idea for a solution

From @melisande-c 
Use `write_on_batch_end` from [prediction writer callback](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.BasePredictionWriter.html). After each batch is predicted, write tiles to a zarr array.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: implements batch decollate, tile stitching and prediction writer callbacks for the NG Dataset.

We currently lacks several important NG Dataset compatible prediction features:
- Tile stitching from `list[ImageRegionData]`
- Equivalent to `convert_outputs` to decollate and potentially stitch back predictions
- Lightning prediction writers to save prediction to the disk
- Zarr tile prediction writer capable of writing one tile at a time

This PR implements the aforementioned features in various modules.

**Notes**:
- Zarr is written as supported for type writing, but only for tiles.

### Overview - what changed?

- `convert_prediction.py` and `stitch_prediction.py` (in `careamics.lightning.dataset_ng.prediction`) now provide the necessary functions to recombine prediction batches into numpy arrays.
- Previously implemented `PredictionWriterCallback` and `WriteStrategy` have been adapted to the NG Dataset and added to `careamics.lightning.dataset_ng.callbacks.prediction_writer`.
- `WriteTilesZarr` has been implemented to write tiles one by one to Zarr.
- Mocking in the updated tests has been simplified a tad, and is now using `pytest-mock` (a wrapper to the old mock, albeit with some additional benefits, e.g. fixture, spy).

All concerned modules have been updated to use `ImageRegionData` rather than the old `(np.NDArray, TileInformation | None)`. In particular, this meant paying attention to `data_idx` and `sample_idx` when reconstituting arrays from tiles or writing tiles to Zarr.

Here are a few lines on the logic behind the main changes:

- `PredictionWriterCallback`:
    - It retained the flag to switch off writing predictions.
    - Because it must be instantiated and passed to the trainer long before prediction, `write_strategy` can be passed later using `set_write_strategy`. Technically, it is not passed as such, but the parameters to the factory are passed to the method.
    - The callback is responsible for decollating the batches as they come. Writing strategy thus receive a list of decollated `ImageRegionData`.
- `CacheTiles` write strategy:
    - `TileSpecs` have now a new entry `tot_tiles`.
    - The write strategy maintains a `dict[int: list[ImageRegionData]]`.
    - Upon call to `write_batch`, the caches are updated.
    - It then checks if any list is of the expected `tot_tiles` size, and stitch and write those to the disk. 
- `WriteTilesZarr`:
    - The write strategy holds three pointers: zarr store, zarr group and zarr array.
    - Upon receiving a new set of tiles, each tile is passed to `write_tile`.
    - For each tile, the source URI is deciphered into a path to the store, internal parent path and array name.
    - Each of these values is compared to the pointers, and new zarr store, group or array are created if necessary using paths and the metadata contained in the `ImageRegionData` (e.g. shape, chunks)
    - Finally, the tile is cropped and inserted in the correct position.

### New features or files

- `TileSpecs` has been updated to include `last_tile`, used later in `CacheTiles` write strategy.
- `ImageStack` protocol has been updated to include `chunks`, with a default `(1,)` for all image stacks and the Zarr chunks for the `ZarrImageStack`.
- `convert_prediction.py` and `stitch_prediction.py` contain methods to decollate batches, combine `ImageRegionData` that belong to the same `data_idx`, and stitch tiles into whole arrays. A `convert_prediction` method allows to call the correct set of methods to directly convert the output of `trainer.predict` to a list of arrays.
- `PredictionWriterCallback`: `write_strategy` can now be passed upon instantiation or later via `set_write_strategy`, allowing users to change their mind after training.
- `CacheTiles` strategy has been updated with the logic to find correct number of tiles and pop the tiles belonging to the same `data_idx` from the cache.
- `WriteTilesZarr` has been implemented.

## How has this been tested?

Tests have been added or updated.

## Related Issues

- Resolves #549

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -0,0 +1,159 @@
...
+def decollate_image_region_data(
+    batch: ImageRegionData,
+) -> list[ImageRegionData]:
+    """
+    Decollate a batch of `ImageRegionData` into a list of `ImageRegionData`.
+
+    Input batch has the following structure:
+    - data: (B, C, (Z), Y, X) numpy.ndarray
+    - source: sequence of str, length B
+    - data_shape: sequence of tuple of int, each typle being of length B
```
melisande-c: typo
```suggestion
    - data_shape: sequence of tuple of int, each tuple being of length B
```

tests/lightning/dataset_ng/callbacks/prediction_writer_callbacks/test_ng_cache_tiles_write_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+
+    extended_tiles = tiles[: n_tiles + batch_size]
+    assert len(extended_tiles) == len(cache_tiles_strategy.tile_cache[0])
+
+
+# TODO mock as in previous versions of the test?
+@pytest.mark.parametrize("n_data, shape, axes", [(2, (28, 28), "YX")])
+def test_write_batch_with_last_tile(
+    tiles: list[ImageRegionData], cache_tiles_strategy: CacheTiles, tmp_path: Path
+):
```
melisande-c: Is this name still correct, since there is no last tile param anymore?

melisande-c: I think the docstring needs updating as well

src/careamics/dataset_ng/patching_strategies/patching_strategy_protocol.py
```diff
@@ -52,11 +54,14 @@ class TileSpecs(PatchSpecs):
...
         Where the tile will be stitched back into an image, taking into account
         that the tile will be cropped, in coords relative to the image.
+    tot_tiles: bool
+        Number of tiles belonging to the same data.
     """
 
     crop_coords: Sequence[int]
     crop_size: Sequence[int]
     stitch_coords: Sequence[int]
+    tot_tiles: int
```
melisande-c: I would prefer `total_tiles` but that's just me :)

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_strategy.py
```diff
@@ -0,0 +1,258 @@
...
+        Parameters
+        ----------
+        dirpath : Path
+            Path to directory to save predictions to.
+        predictions : list[ImageRegionData]
+            Decollated predictions.
+        """
+
+
+class CacheTiles(WriteStrategy):
```
melisande-c: Maybe `CacheTiles` and `WriteImage` can go in their own files for consistency

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_strategy.py
```diff
@@ -0,0 +1,258 @@
...
+        dirpath : Path
+            Path to directory to save predictions to.
+        """
+        full_images = self._get_full_images()
+        for data_idx in full_images:
+            tiles = self.tile_cache.pop(data_idx)
+            self._stitch_and_write_single(dirpath, tiles)
+
+
+class WriteImage(WriteStrategy):
```
melisande-c: I'm wondering if we change `WholeSamplePatchingStrategy` to output a `TileSpec` object (where the cropping is zero) if it will just work with the `CacheTiles` strategy and then we don't need this strategy 🤔

But we don't have to try that out now

melisande-c (approved):
Looks great! Pretty much no comments 😁

src/careamics/dataset_ng/patching_strategies/patching_strategy_protocol.py
jdeschamps: I think I prefer it as well.

tests/lightning/dataset_ng/callbacks/prediction_writer_callbacks/test_ng_cache_tiles_write_strategy.py
jdeschamps: Updated!

src/careamics/dataset_ng/patching_strategies/patching_strategy_protocol.py
jdeschamps: Updated

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_strategy.py
jdeschamps: Good idea, done!

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_strategy.py
jdeschamps: That's an interesting idea, but it feels a bit hacky in spirit, since both classes (`WholeSamplePatching` and `CachedTiles`) are built with a different intention.

melisande-c: That's true, and I guess it's not hard to select the correct WriteStrategy based on the type of patching strategy in the config.

PR #623, opened 2025-11-20
Chore: rename all configuration Pydantic models and reorganize modules

jdeschamps:
## Description

We made the choice a while back to stick to Pydantic naming (much due my insistence) and have named all configuration modules and classes `*_model.py` or `*Model`. This always created some confusion with the deep-learning understanding of "models". Is `UNetModel` a model or a configuration?

This PR simply changed all the class and modules name to `*_config.py` and `*Config` for clarity, as well as create sub-modules for better organization.


melisande-c (approved):
Looks good!

PR #624, opened 2025-11-20
Refac(Next-Gen Dataset): Factory functions & Directory structure

### LINKED ISSUE #584: Next-Gen Dataset: Code reorganisation & refactor of factory functions
## Description

Currently the organisation of the next-gen dataset module is a bit cumbersome, which is expected during development. Originally I proposed to remove the `ImageStackLoader` in #472  to remove some complexity but actually it is quite a convenient feature. So instead I think it can be kept but the `PatchExtractor` factory functions can be removed instead and I think the dataset factory functions can also be refined.

Additionally some things are buried in the `patch_extractor` package, such as the `ImageStack` definitions, I think these can be moved to the root `dataset_ng` directory.

melisande-c:
## Description

> [!NOTE]  
> **tldr**: This PR reorganises the code, removing all the almost duplicate functions we had between the dataset and patch extractor factory functions. It also moves the `image_stack` directory into the root `dataset_ng` directory, from inside of the `patch_extractor` directory.

Having the `image_stack` package inside the `patch_extractor` package was a bit cumbersome, having it in the root means it is more accessible and we can see all the components of the Next-Gen Dataset in the root.

All the factory functions that call each other was also confusing and added too much unnecessary bloat.

### Overview - what changed?

This PR adds built-in implementations of the `ImageStackLoader` for arrays, tiffs, iterated tiffs, zarrs, czi, and custom files. Now the dataset creation pipeline is much simpler:
1. select the image stack loader (can also be custom)
2. select the patch extractor type (different for the iterable tiffs)
3. initialise the image stacks for the inputs, targets and masks
4. initialise the patch extractors from the image stacks for the inputs, targests and masks
5. initialise the dataset from the patch extractors.

Mostly just moving code around. The image stack loader implementations come from the `create_x_patch_extractor` functions. The zarr loader comes from the implementation `create_zarr_image_stacks` function in `image_stacks.zarr_utils`.

### New features or files

- `select_image_stack_loader` function in `factory.py`
- `select_patch_extractor_type` function in `factory.py`
- `init_patch_extractor` function in `factory.py`

- `create_dataset` function in `factory.py`
- moved `patch_extractor/image_stacks` -> `image_stacks`

### Removed features or files

- all `create_x_dataset` functions
- all `create_x_extractor` functions

## How has this been tested?

Code reorganisation so the original tests should cover it.

## Related Issues

- Resolves #584 

## Breaking changes

Code using the `create_x_extractor` or `create_x_dataset` functions.

My code diff seems to have got a bit messed up with the moved files, not sure what to do about that

## Future Work / Ideas

As I was working on this, I realised there are a couple more things I'd like to change:
1) I'd like to move the `mode` and `in_memory` parameters into the data config. This way the validation of the patching strategies can be done by pydantic (and we can possibly have a convenience function for converting between predict and training configs). And to me it makes sense to have `in_memory` in the config because using the iterable tiff loading changes the order of the patches.
2) I think it would be better to move all the initialisation of the dataset components out of the CAREamics dataset, so we end up with something like the example below. and the initialisation of all these components happens in the `create_dataset` function. (Or maybe it can be a classmethod of the dataset).

    ```python
    class CareamicsDataset(Dataset, Generic[GenericImageStack]):
        def __init__(
            self,
            data_config: NGDataConfig,
            input_extractor: PatchExtractor[GenericImageStack],
            target_extractor: PatchExtractor[GenericImageStack] | None,
            mask_extractor: PatchExtractor[GenericImageStack] | None,
            patching_strategy: PatchingStrategy,
            normalization: Normalization | None,
            augmentations: list[Augmentations],
            patch_filter: PatchFilterProtocol | None,
            coord_filter: CoordFilterProtocol | None,
        ) -> None: ...
    ```

---

- [x] Pre-commit passes


jdeschamps (approved):
Love it! It makes a lot of sense.

Regarding moving parameters to the configuration, I also agree. 

This PR interacts negatively with https://github.com/CAREamics/careamics/pull/624, should we wait to merge it first?

melisande-c:
> This PR interacts negatively with #624, should we wait to merge it first?

Do you mean #620? But yes we can wait to merge I will deal with the merge conflicts

PR #628, opened 2025-11-24
Chore: update tests to prevent writing files outside of tmp folders

### LINKED ISSUE #601: Tests: Running tests creates `metrics.csv` files in repo directory
It seems maybe some of the instances of `Careamist` in the tests do not have the `tmp_path` fixture as the working directory, which means running the tests locally is creating `metrics.csv` files. These are quite annoying to clean up all the time.

jdeschamps:
## Description

Following https://github.com/CAREamics/careamics/issues/601, I identified a few tests that create a `checkpoint` and `csv_logger` folders in the working directory rather than in a `tmp_path` during testing. I updated the tests to prevent that.

Closes https://github.com/CAREamics/careamics/issues/601

### DIFF
--- a/tests/lightning/test_LVAE_lightning_module.py
+++ b/tests/lightning/test_LVAE_lightning_module.py
@@ -554,7 +554,13 @@ def test_training_loop_hdn(
         multiscale_count=1,
         target_ch=1,
     )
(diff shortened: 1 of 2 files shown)


melisande-c (approved):
Thanks for finding and fixing these tests!

PR #631, opened 2025-11-24
Feat(Next-Gen Dataset): return sources from `convert_prediction` in order

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: 
> - Add `return_sources` to `convert_prediction` to get a list of the corresponding sources.
> - Ensure that `stitch_prediction` and `combine_samples` return `data_idx` in order.

Although we expect the sources and tiles to be passed and returned in the same order, tiles may be shifted. We currently cover this possibility, to in addition ensure that users can track which stitched prediction corresponds to which source (in case multiple worker screw up the order of the sources), this PR simply allows returning a list of source from `convert_prediction`.

## Why this would matter?

With the following snippet, except believing that the stitched predictions are returned in the exact order the sources where given to the `inf_data_module`, there is no way to track back what `convert_prediction` will return. 

```python
tiled_predictions = trainer.predict(model, datamodule=inf_data_module)
stitched_predictions = convert_prediction(
     tiled_predictions,
     tiled=True,
)
```

Indeed if the order of `data_idx` is not sorted, then `stitch_prediction` will sort it and return arrays in the sorted order.


src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -120,7 +120,8 @@ def combine_samples(
 def convert_prediction(
     predictions: list[ImageRegionData],
     tiled: bool,
-) -> list[NDArray]:
+    return_sources: bool = False,
+) -> list[NDArray] | tuple[list[NDArray], list[str]]:
```
melisande-c: I am personally not a fan of functions that return different structures depending on the arguments, but at least with overloads we shouldn't have to deal with checks on the union type elsewhere in the code

```suggestion
@overload
def convert_prediction(
    predictions: list[ImageRegionData],
    tiled: bool,
    return_sources: Literal[False],
) -> list[NDArray[Any]]: ...
@overload
def convert_prediction(
    predictions: list[ImageRegionData],
    tiled: bool,
    return_sources: Literal[True],
) -> tuple[list[NDArray[Any]], list[str]]: ...
def convert_prediction(
    predictions: list[ImageRegionData],
    tiled: bool,
    return_sources: bool = False,
) -> list[NDArray[Any]] | tuple[list[NDArray[Any]], list[str]]
```

src/careamics/lightning/dataset_ng/prediction/stitch_prediction.py
```diff
@@ -63,9 +65,12 @@ def stitch_prediction(
 
     # stitch each image separately
     image_predictions: list[NDArray] = []
+    image_sources: list[str] = []
     for data_idx in sorted_tiles.keys():
         image_predictions.append(stitch_single_prediction(sorted_tiles[data_idx]))
```
melisande-c: I've just realised that `group_tiles_by_key` does not guarantee that the data indices are in the correct order (although it is pretty likely). This is because the data indices are added to the dictionary in the order that they appear in the tiles, so if the tiles are shuffled so will the data indices, see:
https://github.com/CAREamics/careamics/blob/1ee121987ae539d360b4692679113c68d1f0b7ef/src/careamics/lightning/dataset_ng/prediction/stitch_prediction.py#L32-L35
(also side note, it still freaks me out that dictionaries are ordered, even though this has been a feature since 3.7) 


So if we want to guarantee that `stitch_predictions` will return in the same order of the input, we should do something like:

```python
data_indices = list(sorted_tiles.keys())
data_indices.sort()
for data_idx in data_indices:
    ...
# rest of code
```

melisande-c (commented):
While I see the motivation for giving arrays a unique source, and it feels kind of pointless to have the image stack source to be `Literal["array"]` for arrays. I'm not convinced having a hash as the source will be particularly useful either.

Users will never see that hash of the original data so if it is returned to them they don't know what it means. And we can make sure that the predictions are returned to them in the correct order based on the data indices.

However, I think returning path sources to users could be useful for them.

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
melisande-c: oops, first overload should also have the default value `return_sources: Literal[False] = False`

jdeschamps: I am liking the overload less and less, because I find them confusing and annoying to maintain. My brain only sees the same pattern and fails to register the difference. (Java requires a function body for overload, and that makes much more sense to me)

Best alternative in my mind: we just always return the sources and users can ignore them! We could return an empty list for arrays.

melisande-c: Yeah I'm fine with always returning the sources. Anything that prevents having to do checks on the return type for mypy later on

src/careamics/lightning/dataset_ng/prediction/stitch_prediction.py
jdeschamps: > I does not guarantee that the data indices are in the correct order (although it is pretty likely). This is because the data indices are added to the dictionary in the order that they appear in the tiles, so if the tiles are shuffled so will the data indices, see:

Yes that was the motivation for that PR.

Side note: Ahah, I tend to forget that dictionaries are ordered... I will also never get used to it. I think it took me a long time before realizing it.

> So if we want to guarantee that stitch_predictions will return in the same order of the input, we should do something like:

Something I failed to realize yesterday evening is that the order of `data_idx` is actually dictated by the input order, and not by the workers. I was picturing the potential of having `data_idx` assigned in a different order, but this happens in the image stack loader, before all the workers are spawned... Therefore your solution to sort the indices works!

This means that we may not need to keep `sources` as a return element. I feel we should keep it as it will simplify life for writing to disk in the `CAREamist`.

melisande-c: > This means that we may not need to keep sources as a return element. I feel we should keep it as it will simplify life for writing to disk in the CAREamist.

Won't this be handled by the prediction writer callback ?

> Something I failed to realize yesterday evening is that the order of data_idx is actually dictated by the input order, and not by the workers.

Yeah the `data_idx` is dictated by the order that the image stacks are passed to the `PatchExtractor`. And in any case, each worker gets an identical copy of the dataset.

I was just seeing returning the sources as potentially convenient for users if they have some additional steps in their pipeline, but they would already have the sources, so it's not strictly necessary

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: Ok, updated!

src/careamics/lightning/dataset_ng/prediction/stitch_prediction.py
jdeschamps: > 
> Won't this be handled by the prediction writer callback ?

Ah true, in `CAREamics` yes. Brain no workin'.  Then for Lightning API folks not using the prediction writer for any reason.

Kept the sources, I find it handy. I added sorting to the stitching function and added tests to ensure order.

melisande-c (approved):

tests/lightning/dataset_ng/prediction/test_ng_stitch_prediction.py
```diff
@@ -164,11 +166,36 @@ def test_stitching_single_prediction(tiles):
...
+
+# parameters are injected automatically in the tiles fixture
+@pytest.mark.parametrize("n_data", [4])
+@pytest.mark.parametrize(
+    "shape, axes",
+    [
+        ((32, 32), "YX"),
+    ],
+)
+def test_stitching_prediction_ordering(tiles):
```
melisande-c, line 189: wait, I didn't know you could do this ??

jdeschamps: Yeah that's my new favorite thing!!!

melisande-c: useful!

PR #635, opened 2025-11-25
Feat(Next-Gen Dataset): enforce axes in CZI image stack loader

jdeschamps:
## Description

Another problem I ran into while testing the internal CZI CI: one can pass set of axes that are not compatible (e.g. missing `S` and `C`) to the `NGDataConfig` because neither `czi` image stack loader (`load_czis`) nor `CziImageStack` check the axes. All other `ImageStack` call `reshape_axes`, which raises errors when there are inconsistencies between shape and axes.

This PR adds a check on the `axes` in `load_czis`, and corresponding tests.


src/careamics/dataset_ng/image_stack_loader/image_stack_loaders.py
```diff
@@ -240,7 +243,19 @@ def load_czis(
...
+    ------
+    ValueError
+        If the provided axes are not valid.
     """
+    if are_axes_valid(axes) is False:
+        raise ValueError(
+            f"Provided axes '{axes}' are not valid. Axes must be in the `SC(Z/T)YX` "
+            f"format, where Z or T are optional, and S and C can be singleton "
+            f"dimensions, but must be provided."
+        )
```
melisande-c: Maybe this can eventually be validated in the NGDataConfig, but it also makes sense to have it here

tests/dataset_ng/image_stack/czi/test_czi_image_stack.py
```diff
@@ -36,141 +39,199 @@ def create_test_czi(file_path: Path, data: NDArray | list[NDArray]):
...
-            coords[1] : coords[1] + patch_size[1],
-        ]
-    else:
-        coords = (2, 11, 4)
-        patch_size = (4, 16, 9)
-
-        extracted_patch = image_stack.extract_patch(
-            sample_idx=sample_idx, coords=coords, patch_size=patch_size
-        )
+class TestCziImageStack:
```
melisande-c, line 42: I think you can mark the whole class with `@pytest.mark.czi`

melisande-c (approved):

src/careamics/dataset_ng/image_stack_loader/image_stack_loaders.py
jdeschamps: Makes a lot of sense.

I have renamed and moved `are_czi_axes_valid` to `image_utils`, and it is now called also in the validation of `axes` in the configuration.

tests/dataset_ng/image_stack/czi/test_czi_image_stack.py
jdeschamps: Oh yes! done!

PR #640, opened 2025-11-26
Feat(Next-Gen Dataset): allow selecting which channels to extract

### LINKED ISSUE #636: Next-Gen Dataset: specify channels
## Problem

Currently, all channels are used for training if they are in the data. We should expose the `channel` parameter of the `default_patch_constr` to allow training on a specific channel.

That would require adding a `channel: Sequence[int] | None` parameter to the configuration. In turn, that would allow validating the input or output channels of the models in the configuration.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Allow specifying channels from which to extract data, rather than always pulling all of them.

This PR introduces the `channels: Sequence[int] | None` parameter in the `NGDataConfig`, which is propagated from the `CAREamicsDataset` all the way to `ImageStack.extract_channel_patch`. Users can specify a single channel, a list of channels, or `None` to select all of them.

Previously, `channel_idx` had been introduced and only allowed a single channel or all, and was not accessible.

By user request, this can be a convenient feature to not have to re-save data in order to avoid training/predicting from a particular channel.

### Overview - what changed?

`channels` parameter can now be found throughout the configuration, `CAREamicsDataset`, `PatchExtractor`, patch extractor constructor, and `ImageStack`.

- new `channels` parameter in `NGDataConfig`
- `CAREamicsDataset` calls `patch_extractor.extract_channel_patch`
- `channel_idx` has been renamed to `channels` and is now `Sequence[int] | None`
- All `ImageStack` now deal with `channels` as a sequence or `None`

### New features or files

- `channels` parameter in `NGDataConfig`

- All `ImageStack` and patch extractors

## How has this been tested?

Implemented tests.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/636

## Breaking changes

The behaviour of `extract_channel_patch` is now different since the type of `channels` (previously `channel_idx`) has changed.

One consequence is that `input` and `target` now have the same number of channels if `channels is not None`.

---

- [x] Pre-commit passes


jdeschamps:
I will also add some validation in the `ImageStack` to display a more informative error message to users if the requested channel does not exist.

src/careamics/dataset_ng/image_stack/czi_image_stack.py
```diff
@@ -228,23 +240,38 @@ def extract_channel_patch(
...
         else:
-            channels = [channel_idx]
-        for channel in channels:
+            channel_iter = range(len(channels))
+            channel_indices = list(channels)
+
+        # for each channel
+        for channel in channel_iter:
+            # requested channel index
+            channel_idx = channel_indices[channel]
```
melisande-c: why not have `channel_iter = channels`, and then `for channel in channel_iter` works in both cases

src/careamics/dataset_ng/image_stack/image_utils/image_stack_utils.py
```diff
@@ -7,8 +8,36 @@
...
+        preserving its dimension.
+    """
+    if channel_idx is None:
+        return ...
+
+    if len(channel_idx) == 0:
+        raise ValueError("Channel index sequence cannot be empty.")
+
+    if len(channel_idx) == 1:
+        return slice(channel_idx[0], channel_idx[0] + 1)  # preserve dimension
```
melisande-c: pretty sure this is unnecessary since e.g. if `array` has shape `(3, 3)` then `array[[0]]` will have the shape `(1, 3)`.

src/careamics/dataset_ng/image_stack/image_utils/image_stack_utils.py
```diff
@@ -7,8 +8,36 @@
 T = TypeVar("T", bound=np.generic)
 
 
+def channel_slice(
+    channel_idx: Sequence[int] | None,
```
melisande-c: maybe `channels` or `channel_indices` is better

src/careamics/dataset_ng/patch_extractor/patch_construction.py
```diff
@@ -92,13 +92,19 @@ def lateral_context_patch_constr(
...
-        channel_idx: int | None,  # `channel_idx = None` to select all channels
+        channels: Sequence[int] | None,  # `channels = None` to select all channels
         coords: Sequence[int],
         patch_size: Sequence[int],
     ) -> NDArray[Any]:
+        if channels is not None and len(channels) > 1:
+            raise NotImplementedError(
+                "Lateral context patch constructor is only implemented for single-"
+                "channel images."
+            )
```
melisande-c, line 104: Actually when constructing the synthetic input for MicroSplit we might be doing the channel summation after the patch is extracted to be compatible with zarr, but this is not finalised. In any case this patch constructor may need to extract multiple channels

src/careamics/dataset_ng/microsplit_input_synth.py
melisande-c, line 367: In fact, this is where the lateral patch constructor is used to extract multiple channels, in regards to my other [comment](https://github.com/CAREamics/careamics/pull/640/files#r2565776772)

melisande-c (commented):
Looks good, some small things

src/careamics/dataset_ng/image_stack/czi_image_stack.py
jdeschamps: For instance if `channels = [0, 2]`, I need both an index on the `range` and the actual channel `index`.

Probably a better solution is actually to do `for i, channel in enumerate(channel_iter)`, where `i` will index the channel in the patch and `channel` the channel in the original data.

src/careamics/dataset_ng/image_stack/image_utils/image_stack_utils.py
jdeschamps: Ah I missed that!

src/careamics/dataset_ng/image_stack/image_utils/image_stack_utils.py
jdeschamps: Oh yes, inconsistent with the rest of the PR as well.

src/careamics/dataset_ng/image_stack/czi_image_stack.py
melisande-c: Oh to insert it back into the patch, I missed that, sorry! read the code too quickly I guess

src/careamics/dataset_ng/patch_extractor/patch_construction.py
jdeschamps: I didn't want to touch that part of the code, and since the function is enforcing `None` or `1` (see `n_channels = shape[1] if channel_idx is None else 1`) I did not want to change the behaviour.

src/careamics/dataset_ng/image_stack/czi_image_stack.py
jdeschamps: No worries, I missed it as well the first time when writing the code!

src/careamics/dataset_ng/patch_extractor/patch_construction.py
melisande-c: Oh right I see, then it's just the error message that is wrong, since `None` selects all the channels. Yeah since that part of the code is not finalised, it makes sense to leave it and I'll deal with it another time

melisande-c (approved):

src/careamics/dataset_ng/image_stack/image_utils/image_stack_utils.py
jdeschamps: Fixed!

src/careamics/dataset_ng/image_stack/image_utils/image_stack_utils.py
jdeschamps: Fixed, function is a lit simpler now!

src/careamics/dataset_ng/patch_extractor/patch_construction.py
jdeschamps: Oh I get it now... Yeah the phrasing was not good.


How about that:
```python
            raise NotImplementedError(
                "Selecting multiple channels is currently not implemented for lateral "
                "context patches. Select a single channel or pass `channels=None` to "
                "select all channels."
            )
```

src/careamics/dataset_ng/image_stack/czi_image_stack.py
```diff
@@ -228,24 +240,33 @@ def extract_channel_patch(
...
+        channel_iter: Iterable
+        if channels is None:
+            channel_iter = range(self.data_shape[1])  # iter over number of requested C
         else:
-            channels = [channel_idx]
-        for channel in channels:
+            channel_iter = list(channels)
+
+        # for each channel
+        for path_channel, data_channel in enumerate(channel_iter):
```
melisande-c: Typo `path` -> `patch` I assume 🙈

jdeschamps: Damn it

jdeschamps: Thanks for catching it, pushing now.

PR #642, opened 2025-11-26
Feat(Next-Gen Dataset): add `in_memory` parameter to configuration

### LINKED ISSUE #638: Next-Gen Dataset: `in_memory` should be a `NGDataConfig` parameter
## Problem

`in_memory` should be a `NGDataConfig` parameter.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Adds `in_memory` parameter to `NGDataConfig`.

Following https://github.com/CAREamics/careamics/issues/638:
- Fix the selection of patch extractor
- Adds `in_memory` to NGDataConfig`:
     - Can be omitted, in which case a default is applied
     - Defaults to `True` for `array`, `tiff`, and `custom`
     - Defaults to `False` for `zarr` and `czi`
     - Must be `True` for `array`
     - Must be `False` for `zarr` and `czi`
- Remove `use_in_memory` and `in_memory` parameters in Lightning DataModule and `create_dataset`, to instead use internally the config parameter.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/638

## Breaking changes

Any code relying on explicit `in_memory` or `use_in_memory` parameters in the various functions and classes.

---

- [x] Pre-commit passes


src/careamics/config/data/ng_data_config.py
```diff
@@ -136,6 +136,12 @@ class NGDataConfig(BaseModel):
     batch_size: int = Field(default=1, ge=1, validate_default=True)
     """Batch size for training."""
 
+    in_memory: bool = Field(default=None, validate_default=True)
```
melisande-c: Could use Pydantic's [default factory option](https://docs.pydantic.dev/latest/concepts/fields/#default-values), I would maybe prefer it to having a "before" validator that modifies the value, because it is less hidden.


```suggestion
    in_memory: bool = Field(
        default_factory=lambda data: (
            False if data["data_type"] in ["zarr", "czi"] else True
        ),
        validate_default=True,
    )
```

src/careamics/config/data/ng_data_config.py
```diff
@@ -230,6 +236,53 @@ def axes_valid(cls, axes: str, info: ValidationInfo) -> str:
...
+            raise ValueError(
+                f"`in_memory` can only be True for 'array', 'tiff' and 'custom' "
+                f"data types, got '{data_type}'. In memory loading of zarr and czi "
+                f"data types is not currently not implemented."
+            )
+
+        if not in_memory and data_type == "array":
+            raise ValueError(
+                "`in_memory` must be True for 'array' data type, got False."
+            )
```
melisande-c, line 299: If setting the default values is moved to the default factory then these can be in an "after" validator

src/careamics/dataset_ng/factory.py
```diff
@@ -112,8 +110,28 @@ def init_patch_extractor(
...
+        The type of data being handled.
+    in_memory : bool
+        Indicates whether data is to be loaded into memory.
+
+    Returns
+    -------
+    type[PatchExtractor]
+        The selected PatchExtractor type.
+    """
+    if in_memory or data_type == SupportedData.ZARR or data_type == SupportedData.CZI:
```
melisande-c: I've realised we are also assuming `data_type == SupportedData.Array` won't happen at the same time as `in_memory == False` here... it is validated in the config though, so not that big a deal

melisande-c (commented):
Small suggestion to use pydantic's default factory option instead of a "before" validator :)

src/careamics/config/data/ng_data_config.py
jdeschamps: Good idea!

I missed you commit suggestion, I already had pushed my commit.

src/careamics/config/data/ng_data_config.py
jdeschamps: Removed the `"before"` validation.

src/careamics/dataset_ng/factory.py
jdeschamps: Fixed it nonetheless!

src/careamics/dataset_ng/factory.py
```diff
@@ -112,8 +110,32 @@ def init_patch_extractor(
...
+        The selected PatchExtractor type.
+    """
+    if in_memory or data_type in (
+        SupportedData.ZARR,
+        SupportedData.CZI,
+        SupportedData.ARRAY,
+    ):
         return PatchExtractor
     else:
         return LimitFilesPatchExtractor
```
melisande-c: mm, sorry if this is annoying but maybe it's better to invert the condition, to me it's clearer to see the cases this way around:
```suggestion
    if not in_memory and data_type in (SupportedData.TIFF, SupportedData.CUSTOM):
        return LimitFilesPatchExtractor
    else:
        return PatchExtractor
```

melisande-c (approved):
Looks good, one last small thing, but feel free to ignore

src/careamics/dataset_ng/factory.py
jdeschamps: No it makes sense, thanks!

PR #645, opened 2025-11-28
Feat(Next-Gen Dataset): add `mode` parameter to `NGDataConfig` and add config conversion

### LINKED ISSUE #639: Next-Gen Dataset: add `mode` parameter to `NGDataConfig`
The new `CAREamicsDataset` works for both training and prediction but it has a `mode` parameter which can be the enum values `Mode.TRAINING`, `Mode.VALIDATING`, `Mode.PREDICTING`. This parameter is only used for validating that the patching strategies are correct, which suggests it should move to the configuration and the validation should be handled with pydantic.

Validation currently happens in this method:
https://github.com/CAREamics/careamics/blob/b029aab272614bd73179088805de30193fac19b4/src/careamics/dataset_ng/dataset.py#L129-L135

**Another thought:** The `NGDataConfig` could have a convenient method to return a copy of the config but in a different mode (so it should switch the patching strategy), with options to update parameters such as the `data_type`. This should not be too hard using the pydantic [`model_copy`](https://docs.pydantic.dev/latest/concepts/models/#model-copy) method.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Move `mode` and patching strategy validation from `CAREamicsDataset` to `NGDataConfig`, and add convenience functions to convert from training to validation or prediction `NGDataConfig`.

Moving `mode` to the config allows moving validation to the Pydantic configuration. This ensures better coherence of the configuration and of the pipeline.

A nice idea raised in https://github.com/CAREamics/careamics/issues/639 is to add a convenience function to convert a configuration from one mode to the other.

Here, I chose to only support `training` to `validating` or `predicting`, as I don't see any other use case. Additionally, `training` requires more parameters and validation, while `validating` and `predicting` have simple parameter sets.

Notes and potential issues:
- One problem with the current design choice is that to ask for all channels (when `channels` was not `None` previously)  when converting the configuration, one cannot specify `channels=None`, as this would just keep the previous `channels`... Solution is probably to pass an empty list (counter-intuitive) or to specify the channels (e.g. `channels = [0, 1, 2]`).
- Probably the warning should simply say that errors will be raised, rather than performance drop, as I believe (untested as of yet) that both UNet and LVAE will throw an error.
- A consequence of the conversion in the `CAREamicsDataModule` is that if the data does not need patching (or different batching), prediction data could also be passed to `CAREamicsDataModule` and predicted on directly after training. This is untested for the moment, and we would need to add some warnings.

Validation of the patching strategies was done in the `CAREamicsDataset`, but it belongs to the Pydantic configuration.

### Overview - what changed?

- `mode` is now a parameter in `NGDatasConfig`
- `NGDatasConfig` has a `self.convert_mode` that allows converting training configurations into validation or prediction, and passing some parameters to be swapped in the configuration.
- A new `FixedRandomPatchingConfig` has been implemented.
- A new factory for creating patching strategies from configuration, `create_patching_strategy`, has also been implemented.
- `TilingStrategy` `tile_size` parameter has been renamed to `patch_size` to be coherent with the configuration and simplify instantiating it in the factory.

Small additional changes:
- `PatchExtractor` parameter `shape` has been renamed `shapes`.

## How has this been tested?

Tests only.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/639

## Breaking changes

Any code directly calling `NGDataset`, `CAREamicsDataset`, and `TilingStrategy`.

Remaining TODOs:
- [ ] Assess warnings "degraded performance", it may just be an error.
- [ ] Test whether a single `CAREamicsDataModule` can be used

---

- [x] Pre-commit passes


src/careamics/dataset_ng/patching_strategies/patching_strategy_factory.py
```diff
@@ -0,0 +1,50 @@
...
+            patch_class = FixedRandomPatchingStrategy
+        case SupportedPatchingStrategy.TILED:
+            patch_class = TilingStrategy
+        case SupportedPatchingStrategy.WHOLE:
+            patch_class = WholeSamplePatchingStrategy
+        case _:
+            raise ValueError(f"Unsupported patching strategy: {patching_config.name}")
+
+    # remove `name` to match the class signatures
+    # tiling requires `tile_size` instead of `patch_size`, hence the aliasing
```
melisande-c, line 47: Is this comment correct it seems the argument of `TilingStrategy` has been updated from `tile_size` to `patch_size`.

https://github.com/CAREamics/careamics/blob/7835cebedcf6062bfa8c5a9b833c819336c92c53/src/careamics/dataset_ng/patching_strategies/tiling_strategy.py#L10-L22

src/careamics/config/data/ng_data_config.py
```diff
@@ -523,6 +624,24 @@ def propagate_seed_to_transforms(self: Self) -> Self:
                     transform.seed = self.seed
         return self
 
+    @model_validator(mode="after")
+    def propagate_seed_to_patching(self: Self) -> Self:
+        """
+        Propagate the main seed to the patching strategy if it supports seeds.
```
melisande-c, line 630: Nice, good idea

src/careamics/config/data/ng_data_config.py
```diff
@@ -661,6 +780,212 @@ def set_means_and_stds(
             target_stds=target_stds,
         )
 
+    # TODO should we expose a patch_size and overlaps parameters rather than
+    #   requiring a new patching strategy object? They could be replaced by the
+    #   training patches if not provided. Tiling could also be requested.
+    # TODO `channels=None` is ambigouous: all channels or same channels as in training?
+    # TODO this method could be private and we could have public `to_validation_config`
+    #   and `to_prediction_config` methods with appropriate parameters
```
melisande-c: Regarding the first and third points, yes for a user facing function having to provide a patching config object is maybe annoying, especially for validation where there is only one option. I guess it makes slightly more sense for prediction because the "whole" strategy doesn't require a patch size or overlap, but maybe if the `patch_size` is `None` we could infer the use of the `WholePatching` strategy?

src/careamics/config/data/ng_data_config.py
```diff
@@ -661,6 +780,212 @@ def set_means_and_stds(
...
+                f"observed if the channel dimension in the data is not a singleton "
+                f"dimension. To select a specific channel, use the `new_channels` "
+                f"parameter (e.g. `new_channels=[1]`).",
+                stacklevel=1,
+            )
+        elif (
+            new_axes is not None
+            and ("C" not in new_axes)
+            and ("C" in self.axes)  # removing C axis
+            and ((self.channels is not None) and (len(self.channels) != 1))
```
melisande-c: `and (len(self.channels) != 1))` is this correct? Even if only 1 channel is chosen how can it be indexed if there is no `"C"` dimension in the data?

Sorry if I'm missing something, I'm finding this if-else block a bit confusing

src/careamics/config/data/ng_data_config.py
```diff
@@ -661,6 +780,212 @@ def set_means_and_stds(
...
+
+        # switching channels
+        if (
+            new_axes is not None
+            and "C" in new_axes
+            and "C" not in self.axes  # adding C axis
+            and (
+                (new_channels is not None) and (len(new_channels) != 1)
+            )  # > 1 channels
+        ):
```
melisande-c: Do you think this is clearer? Ignore if not

```suggestion
        adding_C_axis = (
            new_axes is not None and "C" in new_axes and "C" not in self.axes
        )
        selecting_singleton_C = (new_channels is not None) and (len(new_channels) == 1)
        if adding_C_axis and not selecting_singleton_C:
            ...
```

src/careamics/config/data/ng_data_config.py
```diff
@@ -661,6 +780,212 @@ def set_means_and_stds(
...
+        else:
+            patching_strategy = new_patching_strategy
+
+        # create new config
+        model_dict = self.model_dump()
+        model_dict.update(
+            {
+                "mode": new_mode,
+                "patching": patching_strategy,
+                "batch_size": new_batch_size or self.batch_size,
```
melisande-c, line 974: hmm, I see this potentially leading to some maintenance problems, if we add a new parameter to the NGDataConfig and we forget to add it here then when converting to a different mode this parameter will not retain its value in the new copy but instead it will become the default value, which might be unexpected.

This is the advantage of returning `self.mode_copy(update=updated_dict)`, but `updated_dict` will have to be constructed before, only adding items if `new_x` is not `None`.

src/careamics/config/data/ng_data_config.py
melisande-c: On the second point, yeah I see the problem. I feel that `channels=None` should follow the behaviour of the other arguments - that is, it should mean that the value from the original config is used, otherwise it is quite confusing, especially as `None` is a default value. But to specifically change it so that all channels are used yeah I think it should be either the sequence `[0, 1, ... n]` as you suggest in the PR or maybe a Literal like `"all"`.

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -416,9 +422,19 @@ def setup(self, stage: str) -> None:
...
+                    "CAREamicsDataModule configured for validating cannot be used for "
+                    "prediction. Please create a new CareamicsDataModule with a "
+                    "configuration with mode='predicting'."
+                )
+
             self.predict_dataset = create_dataset(
-                mode=Mode.PREDICTING,
-                config=self.config,
+                config=(
+                    self.config.convert_mode("predicting")
```
melisande-c, line 434: What if the config is already in prediction mode?

melisande-c:
^ Just some discussion points, but looks good, everything's coming together nicely

src/careamics/dataset_ng/dataset.py
```diff
@@ -127,52 +112,10 @@ def __init__(
...
-            else:
-                patching_strategy = WholeSamplePatchingStrategy(
-                    data_shapes=self.input_extractor.shape
-                )
-        else:
-            raise ValueError(f"Unrecognised dataset mode {self.mode}.")
+        patching_strategy = create_patching_strategy(
+            data_shapes=self.input_extractor.shapes,
+            patching_config=self.config.patching,
+        )
```
melisande-c: Actually also, maybe this method can be removed now and in `__init__` we simply have
```python
    self.patching_strategy = create_patching_strategy(
        data_shapes=self.input_extractor.shapes,
        patching_config=self.config.patching,
    )
```

src/careamics/dataset_ng/patching_strategies/patching_strategy_factory.py
jdeschamps: No, I should have removed that!

src/careamics/config/data/ng_data_config.py
jdeschamps: It also break my brain...

The problem is that if there is a singleton channel in the data, users have to pass `C` in axe. As the validation stands, you could also pass `channels=[0]`, and the configuration would be valid.

You could then remove "C" in your prediction data and still predict on it.

Now writing this, I see one issue: you end up with a prediction config that does not have "C" and would anyway not validate.

src/careamics/config/data/ng_data_config.py
jdeschamps: Wouldn't it appear with its value in `self.model_dump()`?

src/careamics/lightning/dataset_ng/data_module.py
jdeschamps: If it is already in "predicting" mode it should then use the one already configured:

https://github.com/CAREamics/careamics/blob/7835cebedcf6062bfa8c5a9b833c819336c92c53/src/careamics/lightning/dataset_ng/data_module.py#L432-L438

src/careamics/config/data/ng_data_config.py
melisande-c: Oh right, sorry I'm a bit tired, you're correct both ways are fine

src/careamics/config/data/ng_data_config.py
melisande-c: right, yeah I think I just was primed from reading the first block and didn't notice that this is now checking the original channels

melisande-c: I feel like this whole thing would be easier to validate if we knew the number of channels

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: Sorry! Keep reading things wrong today 🙃

src/careamics/config/data/ng_data_config.py
jdeschamps: Yes, it's much better! Thanks!!

src/careamics/dataset_ng/patching_strategies/patching_strategy_factory.py
jdeschamps: I updated the comment.

src/careamics/config/data/ng_data_config.py
jdeschamps: I updated the function:

1. and 3.: now `new_patch_size` and `overlap_size` should be passed. If no `new_patch_size` are passed, then it selects the default strategy. 
2. I like the "all" parameter, I added it. It now converts internally into old value vs all channels, and that simplifies the error handling.

src/careamics/config/data/ng_data_config.py
jdeschamps: Yeah the number of channels are in the algorithm configuration... I don't really see how to solve that. But this is really edge case so as long as warnings are clear and we cover all cases, I think we will be ok.

src/careamics/config/data/ng_data_config.py
jdeschamps: From the doc I also have the impression that `model_copy` may not trigger validation:

```
update: Values to change/add in the new model. Note: the data is not validated
                before creating the new model. You should trust this data.
```

```
from pydantic import BaseModel

class MyModel(BaseModel):
    a: list[int]


my_model = MyModel(a=[1, 2, 3])

my_model.model_copy(update={"a": "dsadsa"})
```
Returns:
```
>>> MyModel(a='dsadsa')
```

So let's stick to `model_dump`!

src/careamics/dataset_ng/dataset.py
jdeschamps: I removed the method!

melisande-c (approved):
Nice! I think that if-else channel validation block in the `convert_mode` function is a lot clearer now (or maybe I'm less tired 😅). And I think the `convert_mode` function itself is more user-friendly now that the `PatchingConfig` doesn't have to be created by the user.

And sorry again for all the comments that made no sense, bec I didn't read the code properly aha

jdeschamps:
> And sorry again for all the comments that made no sense, bec I didn't read the code properly aha

I agree, I also had to spend 10 mins on each if statement just to understand what I wrote... That is not good sign! 

No worries can happen, the review helped the PR, that's all that matters! :)

PR #651, opened 2025-12-03
Fix(Next-Gen Dataset): fix issue causing empty channels when predicting on channel subset

### LINKED ISSUE #644: Bug(Next-Gen Dataset): stitching function ignores channel number discrepancy
## Describe the bug

If selecting a number of channels for prediction that is lower than the total amount of channels, missing channels are replaced by empty slices by the stitching function. The reason is that `ImageRegionData` carry `data_shape` with the original number of channels, and that is the basis for the stitched array.

Potential solutions:
- Stitching function can infer the number of channels from the first tile.
- The dataset can modify `data_shape` based on the configuration.
- `ImageRegionData` could carry a `channels` info used by the stitching function.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Fix an issue inserting empty channels in stitching reconstruction.

This PR does the following:
- Update `data_shape`, which is used to reconstruct the predictions, to reflect the change in channels and remove singleton dimensions. This happens in the `CAREamicsDataset` when instantiating `ImageRegionData`.
- Along the way I fixed a wrong parameter in `create_ng_dataset_config`.
- Added docs to `ImageRegionData`.
- Made stitching and whole image writing `squeeze()` images before returning or writing them. Before that, depending on the input, singleton dimensions may or may not be present. Now that are systematically removed.
- Updated tests to also work with channels.

The `channels` parameter allow selecting a subset of channels for training/prediction. However, upon reconstructing the data after prediction, either through stitching or writing Zarr tiles, the original data size is used and is returned with the missing channels being empty arrays. 

### Overview - what changed?

`ImageRegionData` carries all the metadata associated with a patch or tile, including the original `data_shape` used during stitching. This PR modifies the `data_shape` to reflect the changes in channels.

In `CAREamicsDataset`, when instantiating `ImageRegionData`, `data_shape` is adjusted.

- `CAREamicsDataset`: 
    -  adjust `data_shape` for `ImageRegionData` to reflect actual channels.
    - Added docs to `ImageRegionData`.
- `create_ng_dataset_config`: removed `patch_overlap` parameter which is not used since it only creates a training configuration (`random` strategy). 
- `write_image_strategy`, `stitch_prediction` and `convert_prediction`: `squeeze()` images before returning them.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/644

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
melisande-c, line 266: I've just realised the dimension order for shape and chunks can be not in the same order here. `data_shape` in image stacks is the "reshaped" order i.e. SC(Z)YX, but chunks is just `array.chunks` which can be in any order.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -121,6 +121,7 @@ def _create_array(
 
             shape = [i for i in shape if i != 1]
 
+            # TODO currently this prevents predicting from anything else than zarr
             if chunks == (1,):  # guard against the ImageRegionData default
                 raise ValueError("Chunks cannot be (1,).")
```
melisande-c: Yeah I think it would be good to be able to set the chunk size in the configuration, but if not specified then it will take the value from zarr, and if the original data. This way it can be a configuration error if writing to zarr is chosen but the chunksize is not specified when the orginal data is not zarr

src/careamics/dataset_ng/dataset.py
```diff
@@ -172,15 +215,26 @@ def _create_image_region(
...
+        # adjust the number of channels in data_shape if needed
+        data_shape = list(image_stack.data_shape)
+        if self.config.channels is not None:
+            # adjusting channels: image stacks have shape SC(Z)YX, therefore channel is
+            # at index 1.
+            data_shape[1] = len(self.config.channels)
+
+        # remove singleton dimensions from data_shape, as it is used to create
+        # inference arrays (e.g. prediction writer)
+        data_shape_tpl = tuple(dim for dim in data_shape if dim != 1)
```
melisande-c: Why do we need to remove the singleton dimensions here if it happens in stitch prediction also. I would prefer to keep the number of dimensions consistent throughout the pipeline and only squeeze at the very end if we have to

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -108,9 +108,11 @@ def combine_samples(
...
         image_regions.sort(key=lambda x: x.region_spec["sample_idx"])
 
-        # remove singleton dims and stack along S axis
-        combined_data = np.stack([img.data.squeeze() for img in image_regions], axis=0)
-        combined_predictions.append(combined_data)
+        # stack along S axis
+        combined_data = np.stack([img.data for img in image_regions], axis=0)
+
+        # squeeze remaining singleton dimensions
+        combined_predictions.append(combined_data.squeeze())
```
melisande-c, line 148: I don't like the squeeze 🙈

I have run into too many issues in the past that have been hard to debug because of squeeze. 

Also the reduction of dimensions happens in a few different places, which I'm finding confusing why it's necessary, see https://github.com/CAREamics/careamics/pull/651/files#r2588902009

My opinion is, if I was doing analysis using careamics and I wrote a pipeline that after doing the prediction looped through samples, if the sample dimension suddenly disappeared for some images even if I specified it, I think that would be annoying to deal with... 

I would be more on board with converting the predictions back to their original dimension order and shape, but that's a little more work

tests/lightning/dataset_ng/prediction/conftest.py
```diff
@@ -0,0 +1,135 @@
...
+    """A fixture that mimicks the output of CAREamics NG Dataset with tiling."""
+    # create data
+    array, all_tiles = tiles
+
+    if "S" in axes:
+        if "C" in axes:
+            if channels is not None and len(channels) == 1:
+                chunks = (1, 8, 8)
+            else:
+                chunks = (1, 8, 8)
```
melisande-c, line 108: This has made me think, what if the data is not chunked along the channels, if only a subset of channels is selected, this is going to break writing to zarr because the chunksize will be larger than the datasize in the channel dimension.

Maybe an easy fix here is, if a subset of channels is selected, the chunksize in the channel dimension should default to 1.

melisande-c (commented):
Ok, so as I said in the comments, I don't like the squeeze. But if you really want it, I think it needs to be tidied up a bit, I think it would be clearer if the squeeze only happens at the very end when the prediction is returned, i.e. in `convert_predictions`, since `stitch_prediction_single` is more of an internal function. And in the `WriteZarrTiles`, the dimension reduction when creating the array is already handled.

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
melisande-c: Or maybe only removing singleton axes if they are not specified in the `axes` is better?

tests/lightning/dataset_ng/prediction/conftest.py
jdeschamps: Yeah that was my comment in yesterday's meeting. 

I guess we know the axes so we can manipulate the chunks. Shards would also need to be adjusted. I would leave any shards/chunk engineering for a subsequent PR.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Great catch! Zarrs in bioimage analysis are usually following the SCZYX convention, but we can;t count on this (and I believe OME-NGFF is moving away from fixed order).

tests/lightning/dataset_ng/prediction/conftest.py
jdeschamps: Or we merge the shards PR first and fix it in this PR since it is more relevant here.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: I like the idea, but currently we have no place in the configuration for the prediction. For the prediction callback, one needs to call `callback.set_writing_strategy(...)` to specify how we want to write it. This method could take on `*kwargs` or specific `chunks/shards` parameters.

At the highest level, the way `CAREamist` is currently written, these parameters would be passed to the `careamist.predict` method.

We could envision extending the Configuration and convenience functions to have both a `NGDataConfig` for training and one for prediction, the latter one being optional. Alternatively, we double the parameters that make sense. And then exposing parameters to override any relevant prediction parameter.

There is some thoughts we need to put on how we want to handle training vs prediction configuration...

src/careamics/dataset_ng/dataset.py
jdeschamps: The philosophical question here is what is `data_shape` representing. It is only used for prediction / writing, so I think it should represent what the end shape should be (even if reshaped compared to the original data).

Though I can totally live with it having singleton dimensions and representing some internal shape before writing. I honestly can't remember what it broke, so I need to have another look at that.

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: Reshaping to the original shape would make sense and has been a necessary feature since the beginning, we just currently don't carry that information. I even believed that was the original thought behind `data_shape` in the metadata. Except that it is not that and the `ImageRegionData` does not have original axes either. This should then be an ulterior PR, bringing an `original_axes` into the `ImageRegionData` (or replacing `axes` with the original ones?).

Removing singleton axes if they are not specified in the `axes` makes sense.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
melisande-c: Actually yeah it makes sense to me to have the `chunks`/`shards` arguments in `careamist.predict_to_disk`.

Alternatively, in the original config there is the `training_config` that specifies parameters for callbacks used during training, so there could be a `prediction_config` that has the parameters for the `PredictionWriterCallback`.

But, I don't think it's necessary to have a copy of the data configuration for prediction.

jdeschamps:
> Ok, so as I said in the comments, I don't like the squeeze. But if you really want it, I think it needs to be tidied up a bit, I think it would be clearer if the squeeze only happens at the very end when the prediction is returned, i.e. in `convert_predictions`, since `stitch_prediction_single` is more of an internal function. And in the `WriteZarrTiles`, the dimension reduction when creating the array is already handled.

In general, I agree with the squeeze being a problem. I ran into a lot of headache-generating problems in the tests and it is probably due to the anarchic way of testing a lot of random shapes (including singleton dims as input) and the fact that there are multiple ways of dealing with prediction outputs (`convert_prediction` uses two different paths internally, each prediction writer strategy uses something different, including using `stitching`). I don't like that there are so many places where we deal with the prediction output. That's on me, but for instance the zarr tile writer and the stitching are not using the same code, which I know realize is silly.

I think I want to deal with the `sharding` first, and then have a second look at the prediction in general.

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
melisande-c: The `ImageRegionData` does have the original axes since they come directly from the config, there is a comment there though that implies otherwise...

I think the only thing preventing us from reshaping back to the original shape actually is if there is both `T` and `S` dimensions, we lose information when flattening them.

src/careamics/dataset_ng/dataset.py
melisande-c: I just think it would make our life much easier if we can assume throughout the entire pipeline if an array has 4 dimensions they are SCYX and if it has 5 dimensions then it is SCZYX, then there is no ambiguity when indexing

melisande-c: Although the `ImageStack.extract_patch` function returns CYX or CZYX ... but I think that's actually necessary since pytorch handles stacking along the batch dimension

Also we know that S is always 1 there

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: Actually came back here to write that, just realized my error.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: It would be pretty annoying to maintain...

I don't like packing them into the training config, because althought that's what Lightning does, it does not make much sense from a pipeline steps perspective.

jdeschamps: Oh I misread what you wrote, I like the `prediction_config` idea, but it is hard to think that it should not include things that are `NGDataset` relevant. They could internally be used to convert the data training config though, so that could actually be seemless.

melisande-c:
> I don't like that there are so many places where we deal with the prediction output. 

Maybe this can be a follow up PR. This PR can just deal with the number of channels in the prediction bug, but keep the previous behaviour that the prediction is always saved with SC(Z)YX dims. But I didn't notice that the zarr prediction squeezes the data, I guess this can be left how it is and be dealt with later

jdeschamps:
> > I don't like that there are so many places where we deal with the prediction output.
> 
> Maybe this can be a follow up PR. This PR can just deal with the number of channels in the prediction bug, but keep the previous behaviour that the prediction is always saved with SC(Z)YX dims. But I didn't notice that the zarr prediction squeezes the data, I guess this can be left how it is and be dealt with later

The zarr prediction does squeeze it as well, here:

https://github.com/CAREamics/careamics/blob/4184d56344512c6e0cd8b244e000cfa67ceb3bac/src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py#L191C12-L191C83

melisande-c:
> But I didn't notice that the zarr prediction squeezes the data, I guess this can be left how it is and be dealt with later

Sorry this is very unclear and written badly. What I meant was on the PR that introduced the zarr writing I didn't originally notice the squeeze, but now I do :)

src/careamics/dataset_ng/dataset.py
jdeschamps: As noted in https://github.com/CAREamics/careamics/pull/653, there is actually a killer argument against removing the singleton dimensions: if some inputs are SYX and others YX (which is allowed in NG Dataset), then their `data_shape` will not allow collating and an error will be raised.

melisande-c: This is actually not allowed, the axes come from the config and are used to load/reshape every image in the provided source. If we wanted this functionality we would have to allow users to pass an axes value per source, but I think this overcomplicates things and the dataset is already flexible enough.

What the dataset does allow is different sizes in the dimensions, i.e. different S, Y, or X lengths.

jdeschamps:
Closing because it will be superseded by another PR.

Issue #652, opened 2025-12-03
Next-Gen Dataset: Propagate metadata when writing predictions


melisande-c:
## Problem
Some metadata is useful to keep with the prediction such as scales in TIFF. We may also want to add the axis metadata.

The `write_func` that is passed to the `CacheTiles` and `WriteImage` accepts arguments, so it may be possible to specify the metadata there. How to propagate metadata from the source to the prediction is currently unclear.

jdeschamps:
https://github.com/CAREamics/careamics/pull/653 will add `additional_metadata` to `ImageRegionData`. It contains `chunks` and `shards` for Zarr sources, but could also in the future include `scales` for other compatible types. If more additional metadata comes to be needed, we could consider adding `additional_metadata: dict[str, Any]` to `ImageStack`.

Having it in `ImageStack` would allow direct copy in `ImageRegionData`, and therefore propagation all the way to prediction.

We would need whether it be collated easily when batching, or if we need adapting the `decollate` method.

Issue #648, opened 2025-12-02
Next-Gen Dataset: allow predicting from arbitrary source and writing tiles to zarr


jdeschamps:
## Problem

Currently the necessity to know about the source `chunks` size is a pre-requisite for writing to Zarr using the zarr prediction writer. Obviously, this is not possible when the source origin is not Zarr.

jdeschamps:
https://github.com/CAREamics/careamics/pull/653 better highlights the pre-requisites for supporting arbitrary source to Zarr tile:

- `shards` and `chunks` can sometimes be absent, but https://github.com/CAREamics/careamics/pull/653 takes care of automatic chunking and `shards` can be `None`.
- An invalid URI `source` will now raise an `NotImplementedError`, preventing predicting from `array` or `tiff`.

A solution to allow `array` and `Path | str` would be:
- for arrays, create a Zarr `prediction.zarr` and number the arrays inside based on `data_idx`
- for `Path | str`, convert to `Path`, create zarr filename based on `path.stem`

PR #653, opened 2025-12-04
Feat(Next-Gen Dataset): Better Zarr tile writing (prediction with `S`, sharding/chunking, error handling)

### LINKED ISSUE #650: Next-Gen Dataset: sharding compatibility
## Problem

Currently, the ZarrImageStack ignore sharding, but it is an important parameter and should be added to `ImageRegionData`, `decollate` and `TileZarrWriter`.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: This PR fixed some important over sights with the Zarr tile writing, and adds or improves sharding, chunking and error handling.

Sharding is useful to reduce the number of files being written to a zarr folder. Currently, the Zarr tile write ignores (literally) the Zarr source shard size. Furthermore, several problems were identified:

- `chunks` and `shards` dimensions order can (in theory) be different from the `data_shape` received by the prediction writer, whether it is because of dimensions order not being natively `SCZYX`, because of additional axes (e.g. `T`), or missing customary axes (e.g. `C` or `S`). e.g. `STYX` in the original data, `SYX` in tiles received from prediction).
- `chunks` and `shards` may not exist in the source, e.g. `array` or `tiff` source.
- Adding `chunks` as a parameter to all `ImageStack` and `ImageRegionData` is violating the spirit of these classes by forcing unrelated parameters onto some implementations. This can get out of hands and lead to technical debt.
- The tile writing is not compatible with `S` dimension due to a faulty implementation and lack of test coverage.

What this PR does is:

- Make tile writing compatible with `S` dimension.
- Remove random `squeeze()` during prediction writing and explicitly remove only the axes that are not present in `axes` (zarr tile writer).
- Implement in `ZarrImageStack` the reshaping of the `chunks` in order to be compatible with the `data_shape` and prediction data dimension order. This also takes care of the putative presence of both `S` and `T` in chunking.
- Remove `chunks` for `ImageStack` implementations (except Zarr) and `ImageRegionData`
- Add `additional_metadata` in `ImageRegionData` to propagate `shards` and `chunks` for `ZarrImageStack`.
- Add `shards` and `chunks` handling in `ZarrImageStack`, including when these are `None` (see `_auto_chunking`).
- Raise an error when the source is not a URI.

Fixing some problems (`S` dimension prediction, error handling when source is not zarr, mismatch between chunks and data shape, chunks in the presence of T), and adding new features (sharding).

### Overview - what changed?

- `chunks` and `shards` are passed though `additional_metadata` in `ImageRegionData`
- `ZarrImageStack` reshapes `shards` and `chunks` the same way as it does `data_shape`
- `WriteTilesZarr` writes `shards`, and can auto chunk if `chunks` are None (will be useful when implementing prediction from different sources, see Notes), and is now compatible with `S` dims.

## How has this been tested?

New tests to image stack and zarr tile writing.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/650

## Notes

- For https://github.com/CAREamics/careamics/issues/654, the reshaping of `shards` and `chunks` might need to be reverted. However, the argument to implement it here is two-fold: prevent potential problems in the mean time (mismatching dimension orders), takes care of the `T` dimension missing.
- Auto-chunking is not currently used because it is only triggered when `chunks` is `None`, meaning when the source is not Zarr. The full compatibility requires `source: str | Path | Literal["array"]` to be handled properly, which will be a subsequent PR: https://github.com/CAREamics/careamics/issues/648.
- In order to propagate other type of metadata (https://github.com/CAREamics/careamics/issues/652), `additional_metadata` in `ImageRegionData` may be used.
- Following this PR, the case where `C` is in `axes` but only a single channel is extracted will cause `data_shape` when writing to Zarr to have a singleton channel. However, when writing to the array, this may be an issue. Currently, I chose to ignore this problem as it should be tackled in https://github.com/CAREamics/careamics/pull/651.

---

- [x] Pre-commit passes

[UPDATE 09.12]: updated the description to fit the current state of the PR.


src/careamics/dataset_ng/image_stack/in_memory_image_stack.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, source: Union[Path, Literal["array"]], data: NDArray):
         self.data_shape: Sequence[int] = self._data.shape
         self.data_dtype: DTypeLike = self._data.dtype
         self.chunks: Sequence[int] = (1,)  # ImageStack protocol, unused here
+        self.shards: Sequence[int] = (1,)  # same
```
melisande-c: It does feel kinda wrong to have these in the protocol since they are only used by the `ZarrImageStack` and are meaningless for every other implementation. 

Additionally, the more we add to the `ImageStack` protocol the harder it becomes for people to implement their own custom implementations.

Instead, I think only the `ZarrImageStack` should have these attributes and when the `ImageRegionData` is created in `CAREamcisDataset._create_image_region` it should check if the instance is a zarr image stack:

```python
    return ImageRegionData(
            data=patch,
            source=str(image_stack.source),
            dtype=str(image_stack.data_dtype),
            data_shape=image_stack.data_shape,
            chunks=image_stack.chunks if isinstance(image_stack, ZarrImageStack) else None,
            shards=image_stack.shards if isinstance(image_stack, ZarrImageStack) else None,
            # TODO: should it be axes of the original image instead?
            axes=self.config.axes,
            region_spec=patch_spec,
        )
```

What do you think? 

Sorry I hadn't quite formulated this thought yet on the chunks PR

src/careamics/dataset_ng/dataset.py
melisande-c: Shards size is not given the `ImageRegionData`

src/careamics/dataset_ng/dataset.py
```diff
@@ -31,7 +31,9 @@ class ImageRegionData(NamedTuple, Generic[RegionSpecs]):
     axes: str
     region_spec: RegionSpecs  # PatchSpecs or subclasses, e.g. TileSpecs
 
+    # zarr specific
     chunks: Sequence[int] = (1,)  # default value for ImageStack without chunks
+    shards: Sequence[int] = (1,)  # default value for ImageStack without shards
```
melisande-c: Maybe we shouldn't have defaults in this class

src/careamics/dataset_ng/image_stack/in_memory_image_stack.py
jdeschamps: I absolutely agree and I thought about this, unfortunately they can't be `None` because torch `default_collate` cannot handle `None` values. I think i wrote this somewhere...

There is an alternative, which I have tried to avoid because I find these parameters currently easy to deal with and easy to roll back if we find a better solution. That alternative is to implement our own collate function to deal with the `None` values.

It would not be hard to implement, since I believe we simply can call the default_collate on all but the optional values and deal with them separately by aggregating them into `None`. It would be passed to torch `DataLoader` in the `CAREamicsDataModule`.

src/careamics/dataset_ng/dataset.py
jdeschamps: Oups... Too fast...

src/careamics/dataset_ng/dataset.py
jdeschamps: Hmmm yeah they are basically useless here, only defining a default as example.

src/careamics/dataset_ng/image_stack/in_memory_image_stack.py
melisande-c: Ok instead of `None` have 
```python
chunks=image_stack.chunks if isinstance(image_stack, ZarrImageStack) else (1,)
shards=image_stack.shards if isinstance(image_stack, ZarrImageStack) else (1,)
```

My point was more about removing them from the protocol

jdeschamps: How about adding a field `additional_metadata: dict{str: Any}` (or less verbose). Might be compatible with default collate, would be `ImageStack` implementation specific and would avoid the explosion of parameters related to only a single `ImageStack` implementation.

If we are dealing with an instance of `ZarrImageStack` we have a definition of `zarr_additional_metadata` that is typed and that we can interpret from `image_stack.additional_metadata`.

melisande-c: Having `additional_metadata` in the `ImageRegionData` makes sense.

But for now since there is only `chunks` and `shards` in the `ZarrImageStack` I don't feel like it is necessary to add an `additional_metadata` attribute to the `ImageStack` protocol.

When the `ImageRegionData` is created we can do:
```python
if isinstance(image_stack, ZarrImageStack):
    additional_metadata = {
        "chunks": image_stack.chunks,
        "shards": image_stack.shards,
    }
else:
    additional_metadata = {}

return ImageRegionData(
    # all the other data
    # ...
    additional_metadata=additional_metadata
)
```

src/careamics/dataset_ng/dataset.py
jdeschamps: Also forgot it in the model. The lack of test coverage is a problem here!

src/careamics/dataset_ng/image_stack/in_memory_image_stack.py
jdeschamps: We could do it like that, and check whether `shards` is `None` to decide whether to add it to the additional metadata. That would make the type of `shards` similar to the original Zarr type, which is nice.

I do like the `additional_metadata` in `ImageStack` because that is a way we could pass other stuff like `scale` or OME metadata.  But maybe in a different PR.

jdeschamps:
While we are at it, another issue with `chunks` and `shards` is that if their length don't agree across `ImageRegionData`, for instance a patch from a SYX Zarr with chunks (1, 16, 16) and a patch from a YX Zarr with chunks (16, 16), then the default_collate will not work.

It is another argument to not get rid of singletons and impose SC(Z)YX everywhere (see https://github.com/CAREamics/careamics/pull/651).

One solution I can see: force them to be strings.

src/careamics/dataset_ng/image_stack/in_memory_image_stack.py
melisande-c: That's true that could be useful. `scale` is metadata about the content of the data, and `chunks`/`shards` are metadata about the structure of the data but I guess the distinction doesn't matter very much

melisande-c:
> While we are at it, another issue with `chunks` and `shards` is that if their length don't agree across `ImageRegionData`, for instance a patch from a SYX Zarr with chunks (1, 16, 16) and a patch from a YX Zarr with chunks (16, 16), then the default_collate will not work.

I was thinking this when I made this [comment](https://github.com/CAREamics/careamics/pull/651#discussion_r2588842776), but assumed since the tests passed it was probably covered aha.

I think if we end up storing the original data shape then we can probably keep the chunks as the original shape and order, but then maybe that has to be addressed first.

jdeschamps:
> > While we are at it, another issue with `chunks` and `shards` is that if their length don't agree across `ImageRegionData`, for instance a patch from a SYX Zarr with chunks (1, 16, 16) and a patch from a YX Zarr with chunks (16, 16), then the default_collate will not work.
> 
> I was thinking this when I made this [comment](https://github.com/CAREamics/careamics/pull/651#discussion_r2588842776), but assumed since the tests passed it was probably covered aha.
> 
> I think if we end up storing the original data shape then we can probably keep the chunks as the original shape and order, but then maybe that has to be addressed first.

And to pile up on it, data shapes will also not stack (collate) if images have different shapes length, unless we keep the singleton dimensions. So that solves the concern. Pertains to https://github.com/CAREamics/careamics/pull/651 as well.

melisande-c:
> > > While we are at it, another issue with `chunks` and `shards` is that if their length don't agree across `ImageRegionData`, for instance a patch from a SYX Zarr with chunks (1, 16, 16) and a patch from a YX Zarr with chunks (16, 16), then the default_collate will not work.
> > 
> > 
> > I was thinking this when I made this [comment](https://github.com/CAREamics/careamics/pull/651#discussion_r2588842776), but assumed since the tests passed it was probably covered aha.
> > I think if we end up storing the original data shape then we can probably keep the chunks as the original shape and order, but then maybe that has to be addressed first.
> 
> And to pile up on it, data shapes will also not stack (collate) if images have different shapes length, unless we keep the singleton dimensions. So that solves the concern. Pertains to #651 as well.

The dataset currently does not allow different shape lengths, see https://github.com/CAREamics/careamics/pull/651#discussion_r2592110380

jdeschamps:
Hmmm so I think I was too quick on previous PRs, because there are problematic things I am realizing only now.

Here is an update of the latest commits: 
- `additional_metadata` now in `ImageRegionData`
- `shards` and `chunks` are now `str`, that solves the issue of stacking them during collate. Indeed, if some sources are `SYX` and others `YX` tuples will be of different length and `default_collate` will raise errors.
- I ran into a big oversight on my side: the zarr tile writer was not compatible with S dimension (e.g. code was attempting to broadcast `YX` into `SYX`). Fixed by modifying the `stitch_slices` to add `S` dimension if present.
- Since `chunks` can be absent from `additional_metadata` (e.g. predicting from tiff files), I implemented an `auto_chunks` function that sets `SC` dims to `1` and spatial dims to `min(dim, 64)`. It only covers axes present in the original axes.
- Removed `squeeze()` from Zarr tile writer, but in order to maintain the original chunks the axes not present in `axes` are removed.

Now, remaining to be solved or tested:
- what happens when T is in axes?
- currently if the source is not zarr, it will still fail. It should instead create a new zarr with the same name as the source.

melisande-c:
> * `shards` and `chunks` are now `str`, that solves the issue of stacking them during collate. Indeed, if some sources are `SYX` and others `YX` tuples will be of different length and `default_collate` will raise errors.

This is not necessary since currently mixed axes are not possible see, https://github.com/CAREamics/careamics/pull/653#issuecomment-3616164919

jdeschamps:
> > * `shards` and `chunks` are now `str`, that solves the issue of stacking them during collate. Indeed, if some sources are `SYX` and others `YX` tuples will be of different length and `default_collate` will raise errors.
> 
> This is not necessary since currently mixed axes are not possible see, [#653 (comment)](https://github.com/CAREamics/careamics/pull/653#issuecomment-3616164919)

That is true... There is another argument for keeping them as string in there, as long as the `additional_metadata` entries are string, the decollate function does not care what is in there (no need to check whether this or that argument is in there, and what their types are).

But that's me again trying to solve a problem that we don't have... So I'll roll back.

src/careamics/dataset_ng/image_stack/in_memory_image_stack.py
jdeschamps: Added that comment in the relevant issue: https://github.com/CAREamics/careamics/issues/652

src/careamics/dataset_ng/dataset.py
jdeschamps: Outdated, as `chunks` is not a parameter anymore.

src/careamics/dataset_ng/dataset.py
jdeschamps: Outdated as `chunks` and `shards` are not defined explicitly anymore.

jdeschamps:
I updated the description of the PR. In essence, here are the changes / content of the PR now:


- Make tile writing compatible with S dimension.
- Remove random squeeze() during prediction writing and explicitly remove only the axes that are not present in axes (zarr tile writer).
- Implement in ZarrImageStack the reshaping of `shards` and `chunks` in order to be compatible with the `data_shape` and prediction data dimension order. This also takes care of the putative presence of both `S` and `T` in chunking.
- Remove `chunks` for `ImageStack` implementations (except Zarr) and `ImageRegionData`
- Add `additional_metadata` in `ImageRegionData` to propagate `shards` and `chunks` for `ZarrImageStack`.
- Add `shards` and `chunks` handling in `ZarrImageStack`, including when these are `None` (see _auto_chunking`).
- Raise an error when the source is not a URI (e.g. `"array"` or a path).

src/careamics/dataset_ng/image_stack/zarr_image_stack.py
```diff
@@ -41,7 +41,14 @@ def __init__(self, group: zarr.Group, data_path: str, axes: str):
...
         self._data_dtype = self._array.dtype
-        self._chunk_size = self._array.chunks
+        self._chunk_size = reshape_array_shape(
+            axes, self._array.chunks, add_singleton=False
+        )
+        self._shard_size = (
+            reshape_array_shape(axes, self._array.shards, add_singleton=False)
+            if self._array.shards is not None
+            else None
+        )
```
melisande-c, line 51: Why do the chunk size and the shard size not have the singleton axes here when the data shape does? I think it would be better to have `data_shape` consistent with `chunks` and `shards` to avoid confusion, and then the singleton axes can all be removed in the same place - when the data is being written.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -7,13 +7,126 @@
...
+            # relabel T as S
+            axes = axes.replace("T", "S")
+    return axes
+
+
+def _auto_chunks(axes: str, data_shape: Sequence[int]) -> tuple[int, ...]:
+    """Generate automatic chunk sizes based on axes and shape.
+
+    Spatial dimensions will be chunked with a maximum size of 64, other dimensions
+    will have chunk size 1.
```
melisande-c, line 80: This might be a bit small for 2D chunks, [zarr docs](https://zarr.readthedocs.io/en/latest/user-guide/performance/#chunk-optimizations) say that chunks should be at least 1MB for best performance. If we're writing float64, a 64x64 chunk will be ~0.3MB.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -7,13 +7,126 @@
...
+
+    if "C" not in updated_axes:
+        sczyx_offset += 1  # singleton C dim added to data_shape
+
+    # loop through the original axes in order SC(Z)YX
+    #   - original_index is the index of the axis in the original `axes` string
+    #   - idx is the index in SC(Z)YX order of the axes present in `axes`
+    #   - since all non spatial are treated the same, we can recover the spatial dims
+    # index in SC(Z)YX order by using sczyx_offset
+    for idx, original_index in enumerate(indices):
```
melisande-c, line 116: i guess a lot of the complication here is due to having to remove the singleton axes

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -95,18 +208,24 @@ def _create_group(self, group_path: str) -> None:
...
     ) -> None:
         """Create a new array in an existing zarr group.
 
         Parameters
         ----------
         array_name : str
             Name of the array within the zarr group.
-        shape : Sequence[int]
+        axes : str
+            Axes string in SC(Z)YX format with original data order.
```
melisande-c, line 224: so is this always `"SCYX"` or `"SCZYX"`?

EDIT: I just read a bit further and saw S and C don't have to be present

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -162,23 +299,35 @@ def write_tile(self, dirpath: Path, region: ImageRegionData) -> None:
...
-            self._create_array(array_name, shape, chunks)
+            chunks: tuple[int, ...] | None = region.additional_metadata.get(
+                "chunks", None
+            )
+            shards: tuple[int, ...] | None = region.additional_metadata.get(
+                "shards", None
+            )
+            self._create_array(array_name, region.axes, shape, shards, chunks)
 
         tile_spec: TileSpecs = region.region_spec  # type: ignore[assignment]
```
melisande-c: Not necessary since we know that region_spec will be a `TileSpec` in prediction, but if you want to remove the type ignore can use a `TypeGuard`:

```python
from typing import TypeGuard
from careamics.dataset_ng.patching_strategies import TileSpecs, PatchSpecs

def is_tile_specs(specs: PatchSpecs) -> TypeGuard[TileSpecs]:
    return (
        ("crop_coords" in specs)
        and ("crop_size" in specs)
        and ("stitch_coords" in specs)
    )
```

then can do:

```python
if is_tile_spec(region.region_spec):
    tile_spec: TileSpec = region.region_spec
else:
    raise TypeError

```

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -187,7 +336,20 @@ def write_tile(self, dirpath: Path, region: ImageRegionData) -> None:
         )
 
         if self.current_array is not None:
-            self.current_array[stitch_slices] = region.data.squeeze()[crop_slices]
+            # region.data has shape C(Z)YX, broadcast can fail with singleton dims
```
melisande-c, line 341: This is after decollating right? otherwise pytorch should return it with a `B` dimension

src/careamics/dataset_ng/image_stack/zarr_image_stack.py
melisande-c: tbh though you can just ignore this bec I guess it will probably change when we implement converting back to the original shape

melisande-c (approved):
minor stuff, feel free to ignore

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Yes, decollate happens in the prediction data writer `write_batch`!

src/careamics/dataset_ng/image_stack/zarr_image_stack.py
jdeschamps: Yes, it is all meant to go away anyway...

I understand the viewpoint, indeed definition of outputs should be in the writer. Just for completeness, the reason why I went for this is that I find it less intuitive to carry arbitrary additional singleton dimensions (e.g. added `S` and `C`) only to remove them at the end. For `data_shape` it makes sense, because it is the guiding tuple to recompositing the whole array in SC(Z)YX space. For `chunks` and `shards`, I just don't see the point. The advantage of doing it already at the `ImageStack` level is also that it centralizes that operation. If we now implement a Zarr whole image predictor, no need to also reshape `chunks` and `shards`.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Do you think the sentence is not clear enough? Should I write `(S)(C)(Z)YX`?

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Ok, then I'll make it `128`!

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Yes, but I can't imagine returning (1, 1, Y, X) for 2D data, and forcing people's entire post-processing pipeline to deal with our specificity...

The code here is indeed surprisingly non-intuitive for such a few simple lines... I am not sure how to do better.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -7,13 +7,126 @@
...
+    #   - idx is the index in SC(Z)YX order of the axes present in `axes`
+    #   - since all non spatial are treated the same, we can recover the spatial dims
+    # index in SC(Z)YX order by using sczyx_offset
+    for idx, original_index in enumerate(indices):
+        axis = updated_axes[original_index]
+
+        if axis in ("Z", "Y", "X"):
+            dim_size = data_shape[idx + sczyx_offset]
+            chunk_sizes.append(
+                min(64, dim_size)
```
jdeschamps: @melisande-c something that just came up to mind, would it make sense to enforce `z_chunk=1`? Accessing a single plane sounds like a typical operation, but that would require pulling 64 planes worth of chunks...

Now that is going to make the code a little more complex as well...

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
melisande-c: Maybe I was just having a silly moment, these are just the axes from the config right?

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Well, `mypy` was throwing an error and I did not find a solution that would not bloat the code. In normal cases, we would use an `isinstance`, but this does not work on typed dicts.

The `TypeGuard` is interesting, I did not know that. But it is also adding quite a few lines of code... I suppose the `TypeGuard` can be reused and placed in the `patching_strategies` module, and then we can use it in an `assert` to please `mypy` or with you suggestion of a `TypeError`.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
melisande-c: hmm maybe? I don't know how people usually chunk their zarrs, but I guess it would make sense each z plane is really large so not many can be loaded into memory at once. Maybe we can just open an issue to think about the auto-chunking a bit harder, just for the sake of getting this PR through

src/careamics/dataset_ng/image_stack/zarr_image_stack.py
melisande-c: Fair, let's just leave this for now and revisit it when we get to converting the predictions back to the original data shape

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
melisande-c: let's just leave this as it is, I think it's fine for now

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
melisande-c: yeah using assert works too

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Yes, I am wondering whether the sentence make it clear that not all axes need to be present. It should also be "with original axes order".

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: TODO added and issue opened. Let's tackle this later indeed.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
jdeschamps: Let's keep it like this for now.

PR #655, opened 2025-12-09
Feat(Next-Gen Dataset): Configurable normalization

### LINKED ISSUE #400: Next-Gen Dataset: Interface for normalisation strategies
In the next-gen dataset we would like to be able to choose different normalisation strategies, or have no normalisation at all (there are cases when the data will be already normalised). 

Features:
- Normalisation function
- Denormalisation function
- Optional running statistics: (requires a method for internal stats to be updated.)

Note: We need to decide whether we want normalisation to applied in the dataset class or the lightning module.

### TODO:
- [ ] Define interface
- [ ] **Option 1:** Concrete implementations.
  - [ ] Implement `ZeroMeanUnitVariance` normalisation with optional running stats using existing `WelfordStatistics` class.
  - [ ] Implement `Percentile` normalisation with probably not implemented error for running stats.
- [ ] **Option 2:** Concrete implementations.
  - [ ] Implement `ZeroMeanUnitVariance`
  - [ ] Implement `RunningZeroMeanUnitVariance`, (will use `WelfordStatistics` class).
  - [ ] Implement `Percentile` normalisation.
- [ ] Create Pydantic models for each normalisation class.
- [ ] Add normalisation to `DataConfig` as a discriminated union. (After old datasets are deprecated).
- [ ] Implement normalisation application, tbd whether in dataset class or lightning module.

**Note:** #303 has some information about percentile/quantile normalisation.

veegalinova:
## Description

> [!NOTE]  
> **tldr**: This PR introduces configurable normalization options to the NG dataset: `standardize` (mean/std), `minmax`, `quantile`, and `none`. Each option has its own Pydantic config with validations. Statistics can be user-provided or computed automatically by iterating through the dataset. 

We need to support normalization options for different use cases. Users can pre-normalize their data before training, but we also need to support more option in the future. So this PR introduces more options for normalization, as well as an extensible system for future options.

### Overview - what changed?

General idea is to add a configurable normalization options to the NG dataset, similar in spirit with the configuration of patching. 
Key changes:
1. New normalization field in `NGDataConfig` that replaces the old `image_means/image_stds` approach
2. Four Pydantic config models: `StandardizeConfig`, `MinMaxConfig`, `QuantileConfig`, `NoNormConfig` with discriminated union
3. Protocol-based implementation of `NormalizationProtocol` that defines `normalize` and `denormalize` methods, keeping both operations in the same class
4. Normalization is now performed as a separate operation in the dataset and not inside the transforms (personal preference, can discuss!)
5. Automatic statistics computation - if statistics aren't provided, they're computed by iterating through **all patches** of the dataset
6. New `QuantileEstimator` - an algorithm for computing dataset quantiles iteratively on the fly 

Normalization is configured via the normalization parameter in `NGDataConfig` (similar to patching):
```python 
# With user-provided statistics
  config = NGDataConfig(
      data_type="array",
      axes="YX",
      patching={"name": "random", "patch_size": (64, 64)},
      normalization={
          "name": "standardize",
          "input_means": [100.0],
          "input_stds": [25.0],
      },
  )
# Auto-computed statistics
config = NGDataConfig(
    ...
    normalization={"name": "standardize"},
)
```

All normalization strategies implement `NormalizationProtocol`:
```python
class NormalizationProtocol(Protocol):
    def __call__(self, patch: NDArray, target: NDArray | None = None) -> tuple[NDArray, NDArray | None]:
        """Apply normalization to patch (and optionally target). Shape: C(Z)YX."""
        ...

    def denormalize(self, patch: NDArray) -> NDArray:
        """Reverse normalization for model output. Shape: BC(Z)YX."""
        ...
```
This keeps normalization and denormalization logic and parameters together. Factory function `create_normalization()` instantiates the appropriate class based on the config.

When the dataset is created, resolve_normalization_config() checks if statistics are missing and computes them:
- `standardize`: uses existing WelfordStatistics 
- `minmax`: tracks global min/max across all patches
- `quantile`: uses the new `QuantileEstimator`

The `QuantileEstimator` computes quantiles in a iterative fashion using adaptive histogram binning (based on Florian's and Milly's idea). The implementation is a bit complicated, but the main idea is: 
1. Maintain a histogram with X bins per channel (default is 65536 for 16-bit data)
2. Bin edges are computed from current min/max with 10% margin for stability
3. If new data extends beyond current range, the histogram is rebinned to the expanded range
4. Final quantiles are computed via cumulative sum + linear interpolation within bins

This approach has limited precision though. Depending on the value magnitude and bin width, final values can be off by up to like +- 10. I tested another known algorithm (P2) as well, but it performed even worse for extreme quantiles (0.01, 0.99) which we are interested in. I think it is a good enougth approximation anyways, it can definitely be improved and we can test and document the expected precision. 

### New features or files

- `src/careamics/config/data/normalization_config.py` - pydantic models: `StandardizeConfig`, `QuantileConfig`, `MinMaxConfig`, `NoNormConfig`
- `src/careamics/dataset_ng/normalization/` - new module containing:
  - `normalization_protocol.py` - protocol definition
  - `standardization.py` - `Standardize` class (mean/std normalization)
  - `range_normalization.py` - `RangeNormalization` class (used by both minmax and quantile)
  - `no_normalization.py` - `NoNormalization` class
  - `normalization_factory.py` - `create_normalization()` factory function
  `statistics.py` - `resolve_normalization_config()` and statistics computation helpers
  - `running_quantile.py` - `QuantileEstimator` class

- `src/careamics/config/data/ng_data_config.py` - added normalization field, removed `image_means/image_stds/target_means/target_stds`
- `src/careamics/dataset_ng/dataset.py` - integrated normalization into dataset initialization and `__getitem__`
- `src/careamics/lightning/dataset_ng/data_module.py` - removed stats computation from data module (now handled by dataset)
- `src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py` - updated denormalization to use `dataset.normalization.denormalize()`
- `src/careamics/config/configuration_factories.py` - default normalization config in factory

### Removed features or files

- removed `image_means, image_stds, target_means, target_stds` fields from `NGDataConfig`
- removed `set_means_and_stds()` method from `NGDataConfig`
- removed `self.stats` from `CareamicsDataModule` (statistics now live in normalization config)

## How has this been tested?

I added some functional tests that check the normalization with provided stats and with auto calculated stats. Also added some tests for the new Quantile estimator. 
I ran the demo BSD68 noetbook with training and tested the expected metrics 
| Metric | Before PR | Standardize | MinMax | Quantile |
|--------|----------|-------------|--------|----------|
| PSNR/SI-PSNR | 26.29/26.30 | 26.28/26.30 | 26.13/26.17 | 26.24/26.26 |
## Related Issues

- Resolves #400 

## Breaking changes

NGDataConfig no longer accepts `image_means`, `image_stds`, `target_means`, `target_stds` as top-level fields.

---

- [x] Pre-commit passes


src/careamics/dataset_ng/normalization/standardization.py
```diff
@@ -0,0 +1,223 @@
...
+
+    This allows to broadcast the stats (mean or std) to the image dimensions, and
+    thus directly perform a vectorial calculation.
+
+    Parameters
+    ----------
+    stats : list of float
+        List of stats, mean or standard deviation.
+    ndim : int
+        Number of dimensions of the image, including the C channel.
```
melisande-c: maybe can add "includes also B/S dimension" I got a little bit confused with the `swapaxes` in the denormalization method below

src/careamics/dataset_ng/normalization/running_quantile.py
```diff
@@ -0,0 +1,190 @@
...
+        if self.total_counts[ch] == 0:
+            return 0.0
+
+        cumsum = np.cumsum(self.histograms[ch])
+        target_count = self.total_counts[ch] * quantile
+
+        if target_count <= 0:
+            return float(self.mins[ch])
+        if target_count >= self.total_counts[ch]:
+            return float(self.maxes[ch])
```
melisande-c: I feel like if the quantile is less than 0 or greater than 1, that should be an error because it's not meaningful. But it makes sense to simply return the min or max if the quantile is 0 or 1.

src/careamics/dataset_ng/normalization/running_quantile.py
```diff
@@ -0,0 +1,190 @@
...
+                    )
+                    self.histograms[ch] = self._rebin_histogram(
+                        self.histograms[ch], self.bin_edges[ch], new_edges
+                    )
+                    self.bin_edges[ch] = new_edges
+
+            hist, _ = np.histogram(channel_data, bins=self.bin_edges[ch], density=False)
+            self.histograms[ch] += hist
+
+    def _calculate_quantile(self, ch: int, quantile: float) -> float:
```
melisande-c, line 150: I think this function can be simplified a bit by using numpy's `interp` function:
```python
def _calculate_quantiles(self, ch: int, quantiles: Sequence[float]) -> NDArray[np.floating]:
    cff = np.cumsum(self.histograms[ch]) # cumulative frequency function
    cdf = cff / cff[-1] # cumulative density function

    bins = self.bin_edges[ch]
    bin_midpoints = (bins[1:] + bins[:-1]) / 2
    # use numpy interp swapping x and y axes (since cdf is monotonic)
    values = np.interp(np.array(quantiles), cdf, bin_midpoints)
    return values
```

src/careamics/dataset_ng/normalization/running_quantile.py
```diff
@@ -0,0 +1,190 @@
...
+            ch: np.zeros(self.n_bins, dtype=np.int64) for ch in range(self.n_channels)
+        }
+        self.mins: dict[int, float] = dict.fromkeys(range(self.n_channels), np.inf)
+        self.maxes: dict[int, float] = dict.fromkeys(range(self.n_channels), -np.inf)
+        self.total_counts: dict[int, int] = dict.fromkeys(range(self.n_channels), 0)
+        self.bin_edges: dict[int, NDArray] = {}
+
+    def _rebin_histogram(
+        self, old_hist: NDArray, old_edges: NDArray, new_edges: NDArray
+    ) -> NDArray:
```
melisande-c, line 43: Another approach could be to specify a bin width and to concatenate more bins on the ends of the histogram without rebinning, I don't know if this would provide any benefits though, just another way of looking at the problem

src/careamics/config/data/normalization_config.py
```diff
@@ -0,0 +1,491 @@
...
+        Returns
+        -------
+        bool
+            True if input statistics are missing, False otherwise.
+        """
+        return self.input_means is None or self.input_stds is None
+
+    def is_resolved(self) -> bool:
+        """
+        Check if all required statistics are populated.
```
melisande-c: Maybe we only need one of these functions

melisande-c (commented):
Nice looks very good! I just had a couple of thoughts on the quantile estimator. But in general, this will make adding future features like calculating stats during the first epoch of training hopefully quite straight forward :)

Also nice tests, I think they do a good job of testing the behaviour rather than the implementation, which I always do a bad job of.

I think we're just missing the case of using stats over all the channels instead of per-channel, which is necessary when the channels need to preserve their relative intensity, such as with spectral data. But we can leave this for a future PR.

src/careamics/config/data/normalization_config.py
veegalinova: Good catch, removed

src/careamics/dataset_ng/normalization/standardization.py
veegalinova: Yes, I agree. I cleaned it up a bit and added an axis parameter so both norm and denorm can use it correctly. For norm, we expect C(Z)YX, then for denorm it's BC(Z)YX as it's located after the model forward, and axis controls where the C dimension is inserted to the statistics array. 

I also added casting to `float32` in all normalization methods while I was at it.

src/careamics/config/data/normalization_config.py
```diff
@@ -0,0 +1,446 @@
...
+        ):
+            raise ValueError(
+                "input_lower_quantile_values and input_upper_quantile_values "
+                "must be both provided or both None."
+            )
+        if (
+            self.input_lower_quantile_values is not None
+            and self.input_upper_quantile_values is not None
+        ):
+            if len(self.input_lower_quantile_values) != len(
```
jdeschamps, line 213: Annoyingly verbose, but shouldn't we also validate that lower quantile values are lower than upper quantile valies?

src/careamics/config/data/normalization_config.py
```diff
@@ -0,0 +1,446 @@
...
+                    "must have same length."
+                )
+
+        if (self.target_lower_quantile_values is None) != (
+            self.target_upper_quantile_values is None
+        ):
+            raise ValueError(
+                "target_lower_quantile_values and target_upper_quantile_values "
+                "must be both provided or both None."
+            )
```
jdeschamps, line 239: Any reason to not test target upper and lower length against each other? (and maybe also that lower values are lower than upper?)

src/careamics/config/data/normalization_config.py
```diff
@@ -0,0 +1,446 @@
+"""Pydantic models for normalization strategies."""
+
+from typing import Annotated, Literal, Self, Union
+
+from pydantic import BaseModel, ConfigDict, Discriminator, Field, model_validator
+
+
+class StandardizeConfig(BaseModel):
```
jdeschamps: Is "standardize" a canonical name? I find it non intuitive and I would think users will wonder what that means. I like "zero_mean_unit_variance" but it is super long...

I am fine with keeping it as it, I am just wondering whether it is worth brainstorming the name here.

src/careamics/dataset_ng/normalization/range_normalization.py
```diff
@@ -0,0 +1,142 @@
...
+            target_maxes = _reshape_stats(
+                self.target_maxes, target.ndim, channel_axis=0
+            )
+            norm_target = (target - target_mins) / (
+                target_maxes - target_mins + self.eps
+            )
+
+        return norm_patch, norm_target
+
+    def denormalize(self, patch: NDArray) -> NDArray:
```
jdeschamps: The input is typed as `NDArray`, but isn't denormalization performed on `torch.Tensor`? This also concerns the other normalizations.

It is not a problem here because you basically only use normal math on the tensors, but if ever a normalization calls numpy methods, it may be an issue.

src/careamics/config/data/normalization_config.py
jdeschamps: I guess `mean_std` could also do?

jdeschamps (commented):
Great work, thanks a lot @veegalinova !

I have not had enough time to look in details at the quantile estimation nor the tests yet... But since Milly is on it, I don't think I can bring much. I have a few meaningless comments here and there.

Milly brought up the point again, but we need a way to normalize equally over all channels.

I want to bring up two points that will require a little brainstorming on whether we want this, and if so, what would be the best mechanism:
- We probably want to support some simple segmentation in CAREamics, which then means that `targets` may be present but normalization of targets not be required. We could think of an additional parameter to all normalization config that would prevent application of normalization, and be used in the statistics resolving method to avoid running stats over the targets.
- It may be useful for users that want to experiment to have a way to plug their own normalization module, let's say in the Lightning API. That would require some `custom` normalization, as well as a parameter passed to the `CAREamicsDataModule`.

These last points, if we decide to implement these mechanisms, should anyway be follow-up PRs.

src/careamics/config/data/normalization_config.py
veegalinova: renamed everything to mean_std

src/careamics/dataset_ng/normalization/range_normalization.py
veegalinova: Edited the protocol and implementation to accept `torch.tensor`. Now normalization accepts `numpy`, and denormalization accepts `torch.tensor`.

jdeschamps (approved):

src/careamics/dataset_ng/normalization/running_quantile.py
veegalinova: I think it's a bit easier to read currently, with cdf it will be more complicated to understand

src/careamics/dataset_ng/normalization/running_quantile.py
veegalinova: I'll take a note of this, thanks! If rebinning is taking a lot of time then it might be the solution. I'm only worried that the histogram might grow quite large like this.

melisande-c: If the bin width is set to 1, then for say uint16 data it can never become larger than 65536 bins, which is already the default

src/careamics/dataset_ng/normalization/running_quantile.py
veegalinova: added!

src/careamics/dataset_ng/normalization/running_quantile.py
melisande-c: fair enough, but cumulative density function is just a cumulative histogram but scaled so that the total is 1, (last value of cumulative histogram is the total counts). So the y-axis of the cdf is the quantiles of the distribution

melisande-c (approved):

PR #661, opened 2025-12-18
Fix(Next-Gen Dataset): fix issue causing empty channels when predicting on channel subset

### LINKED ISSUE #644: Bug(Next-Gen Dataset): stitching function ignores channel number discrepancy

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Fix an issue inserting empty channels in stitching reconstruction.

This PR does the following:
- Update `data_shape`, which is used to reconstruct the predictions, to reflect the change in channels and remove singleton dimensions. This happens in the `CAREamicsDataset` when instantiating `ImageRegionData`.
- Along the way, fixed a wrong parameter in `create_ng_dataset_config`.
- Added docs to `ImageRegionData`.
- Updated tests to also work with channels and, for some, work directly with the `CAREamicsDataset` to avoid having to update constantly how the test inputs are generated.

The `channels` parameter allow selecting a subset of channels for training/prediction. However, upon reconstructing the data after prediction, either through stitching or writing Zarr tiles, the original data size is used and is returned with the missing channels being empty arrays. 

### Overview - what changed?

`ImageRegionData` carries all the metadata associated with a patch or tile, including the original `data_shape` used during stitching. This PR modifies the `data_shape` to reflect the changes in channels.

In `CAREamicsDataset`, when instantiating `ImageRegionData`, `data_shape` is adjusted.

- `CAREamicsDataset`: 
    -  adjust `data_shape` for `ImageRegionData` to reflect actual channels.
    - Added docs to `ImageRegionData`.
- `create_ng_dataset_config`: removed `patch_overlap` parameter which is not used since it only creates a training configuration (`random` strategy). 

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/644

## Notes

This PR uncovered https://github.com/CAREamics/careamics/issues, for which (failing) tests are implemented (but commented out).

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -341,7 +341,7 @@ def write_tile(self, dirpath: Path, region: ImageRegionData) -> None:
             # region.data has shape C(Z)YX, broadcast can fail with singleton dims
             crop = region.data[crop_slices]
 
-            if region.data.shape[0] == 1:
+            if region.data.shape[0] == 1 and "C" not in region.axes:
```
jdeschamps, line 344: Just a comment here: the choice of not writing to zarr singleton dimensions that do not exists in `axes` remain, but will be updated by writing to original shape. Now, in case where selecting a single channel via the `channels` parameter, channel dim can be a singleton but `C` be in `axes`. 

I chose to keep the axis here for now.

src/careamics/dataset_ng/dataset.py
```diff
@@ -24,12 +24,53 @@
...
+    of the channels has been requested, in which case the channel dimension may
+    be smaller than that of the original data and only correspond to the requested
+    number of channels.
+
+    ImageRegionData may be collated in batches during training by the DataLoader. In
+    that case:
+    - data: arrays are collated into NDArray of shape (B,C,Z,Y,X)
+    - source: list of str, length B
+    - data_shape: list of tuples of int, each tuple being of length B and representing
+        the shape of the original images in the corresponding dimension
```
melisande-c, line 43: Just to make sure I understand, this means that the collated `data_shape` ends up being something like e.g, for a batch size of 4:
```
[
    (S_1, S_2, S_3, S_4),
    (C_1, C_2, C_3, C_4),
    (Y_1, Y_2, Y_3, Y_4),
    (X_1, X_2, X_3, X_4),
]
```

src/careamics/dataset_ng/dataset.py
```diff
@@ -24,12 +24,53 @@
...
     data: NDArray
+    """Patch, tile or image in C(Z)YX format."""
+
     source: Union[str, Literal["array"]]
+    """Source of the data, e.g. file path, zarr URI, or "array" for in-memory arrays."""
+
     data_shape: Sequence[int]
+    """Shape of the original image in (SCZ)YX format and order. Singleton dimensions are
+    ignored. If channels are subsetted, the channel dimension corresponds to the number
+    of requested channels."""
```
melisande-c: It doesn't seem that it is true that singleton dimensions are ignored at this point, in the `_create_image_region` method below, the `data_shape` comes directly from the image stack, which is always `SC(Z)YX` and only the number of channel dimensions are modified, see snippet:

https://github.com/CAREamics/careamics/blob/3f7feacb9789effc2bb1901102bf06da9c3a96d8/src/careamics/dataset_ng/dataset.py#L248-L252

tests/lightning/dataset_ng/prediction/test_ng_stitch_prediction.py
```diff
@@ -43,44 +103,42 @@ def tiles(n_data, shape, axes) -> tuple[NDArray, list[ImageRegionData]]:
...
+        overlaps=data_config.patching.overlaps,
     )
     n_tiles = tiling_strategy.n_patches
 
     # create patch extractor
-    image_stacks = load_arrays(source=[array[i] for i in range(n_data)], axes=axes)
+    if data_type == "tiff":
+        image_stacks = load_tiffs(source=sources, axes=axes)
+    else:
+        image_stacks = load_arrays(source=sources, axes=axes)
```
melisande-c: I think the axes should be accessed from `data_config.axes` instead of as a argument to the `tiles` fixture, to avoid accidental mismatches

melisande-c: same with `data_type`

melisande-c: or I guess pytest passes them through to the `data_config`? but idk it's a bit unclear

melisande-c (approved):
Looks good 👍

src/careamics/dataset_ng/dataset.py
jdeschamps: Yes!

src/careamics/dataset_ng/dataset.py
jdeschamps: I wanted to update that comment, but it skipped my mind. Thanks!!

jdeschamps: Done.

tests/lightning/dataset_ng/prediction/test_ng_stitch_prediction.py
jdeschamps: I see what you mean, it is passed to the `data_config` fixture as well, so these are the exact same input parameters.

jdeschamps: Updated the tests to use the config rather, for clarity.

PR #662, opened 2025-12-19
Feat(Next-Gen CAREamics): draft convenience functions for NG-compatible CAREamics (N2V only)

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Refactor and adapt configuration factories for NG Dataset compatiblity

This PR creates a copy of the convenience function for use with the NG Dataset. The problem we are currently facing is that the `Configuration` is not compatible with the NG Dataset, and there is no replacement for using the NG dataset without creating the configs explicitly (e.g. algorithm config).

This PR introduces a N2V-only set of convenience functions and configuration class to create a `NGConfiguration` that is NG dataset-compatible.

In essence:

- Two new submodules `ng_factories` and `ng_configs`. The former holds the convenience functions, the latter global configurations.
- Makes a `NGConfiguration` that accepts `NGDataConfig`, the only other difference with the old `Configuration` is that there is no `set_3D` method anymore (it was never used) and the axes/model conv dims are not silently changed by the validation in case of mismatch, but explicitly raise an error.
- New `N2VConfiguration`, child of `NGConfiguration`, that performs N2V-specific validation.
- New implementation of `create_n2v_configuration` (shadowing the original name, but in sub module `ng_factories`), which also has recently introduced `in_memory` and `channels` parameters. `channels` has a complex interaction with `n_channels` and `axes` since it is a parameter that allows defining which channels to use.
- All convenience functions have been broken up into different modules for clarity (as opposed to the old and massive `configuration_factories.py`...).

Note: I have not put much thought about the `channels`, `n_channels` and `axes` interaction. I implemented what made sense to me naively, we should have a close look and change it in another PR.

## Breaking changes

Unless code was importing `create_ng_data_config` from `careamics.config.configuration_factory` instead of from `careamics.config`, then all other code base should be unaffected.

---

- [x] Pre-commit passes


src/careamics/config/ng_configs/ng_configuration.py
```diff
@@ -0,0 +1,259 @@
...
+        Self
+            Validated configuration.
+        """
+        if "Z" in self.data_config.axes and not self.algorithm_config.model.is_3D():
+            raise ValueError(
+                "The data axes contain a Z dimension, but the algorithm is set to 2D. "
+                "Either change the data axes to not contain Z, or set the algorithm to "
+                "3D."
+            )
+        elif "Z" not in self.data_config.axes and self.algorithm_config.model.is_3D():
```
melisande-c: don't we also need to allow: `data_config.data_type == "czi" and "T" in data_config.axes` for temporal denoising

src/careamics/config/ng_factories/algorithm_factory.py
```diff
@@ -0,0 +1,162 @@
...
+    adapter: TypeAdapter = TypeAdapter(
+        Annotated[
+            Union[N2VAlgorithm, N2NAlgorithm, CAREAlgorithm],
+            Field(discriminator="algorithm"),
+        ]
+    )
+    return adapter.validate_python(algorithm)
+
+
+def _create_unet_configuration(
```
melisande-c: This function is probably the result of multiple refactors and ngl, it is pretty pointless, in the `create_algorithm_configuration` function below why not simply have:

```python
    if model_params is None:
        model_params = {}
    unet_model = UNetConfig(
        # also it would be good if the architecture field was already set by default
        architecture=SupportedArchitecture.UNET.value,
        conv_dims=3 if "Z" in axes else 2,
        in_channels=n_channels_in,
        num_classes=n_channels_out,
        independent_channels=independent_channels,
        use_n2v2=use_n2v2,
        **model_params,
    )
```

src/careamics/config/ng_factories/algorithm_factory.py
```diff
@@ -0,0 +1,162 @@
...
+    loss : {"n2v", "mae", "mse", "pn2v"}
+        Loss function to use.
+    independent_channels : bool
+        Whether to train all channels independently.
+    n_channels_in : int
+        Number of input channels.
+    n_channels_out : int
+        Number of output channels.
+    use_n2v2 : bool, default=false
+        Whether to use N2V2.
```
melisande-c, line 75: Is validation of `use_n2v2=True` and a non-n2v algorithm already handled in the respective algorithm configs?

src/careamics/config/ng_factories/training_factory.py
melisande-c, line 1: What is the benefit of having these two functions? Why not have it as one function?

tests/config/ng_factories/test_n2v_factory.py
```diff
@@ -0,0 +1,416 @@
...
+        assert config.training_config.lightning_trainer_config == {}
+
+    # TODO arguably this should be tested at the level of the data config only
+    def test_czi_with_T_axes(self):
+        """Test that SCTYX is accepted by N2V configuration for CZI data."""
+        config = create_n2v_configuration(
+            experiment_name="test",
+            data_type="czi",
+            axes="SCTYX",
+            patch_size=[64, 64],
```
melisande-c: There is meant to be a 3D patch here

melisande-c (commented):
Looks good, except I am pretty the configuration is still not allowing 3D training for czi with `"SCTYX"` axes, in the tests you have put a 2D patch but it should be 3D.

\+ some minor stuff

src/careamics/config/ng_factories/algorithm_factory.py
```diff
@@ -0,0 +1,162 @@
...
+    # PN2VAlgorithm,  # TODO not yet compatible with NG Dataset
+)
+from careamics.config.architectures import UNetConfig
+from careamics.config.support.supported_architectures import SupportedArchitecture
+
+
+# TODO rename so that it does not bear the same name as the module?
+def algorithm_factory(
+    algorithm: dict[str, Any],
+) -> Union[N2VAlgorithm, N2NAlgorithm, CAREAlgorithm]:
```
melisande-c, line 20: Also are you intentionally duplicating a fair few functions? This `algorithm_factory` exists (with additional PN2V) also here:

https://github.com/CAREamics/careamics/blob/ed1c67f70d8fa2f6d50f9540799bf89ceac18b60/src/careamics/config/configuration_factories.py#L43-L45

Other duplicated functions include:

* `config.configuration_factories._create_unet_configuration`
* `config.configuration_factories._create_algorithm_configuration`

Is this the result of refactoring for ng dataset but not wanting to change anything about the original implementation?

src/careamics/config/ng_configs/ng_configuration.py
jdeschamps: Yes, I totally misread the issue. My smoothied brain only focused on the axes, but of course the problem was in the interplay between dimension and axes. 

it goes further than that because dimensions in the algorithm/model must also be matched, but the Configuration validation only checks the axes for Z.

- I updated all the NG Configuration with the proper validation.
- I introduced `is_3D` in `NGDataConfig` which is then compared to `model.is_3D`. This limits the use of the `if else` and `czi` to the `NGDataConfig`. 
- algorithm convenience function now takes a `dimensions` parameter instead of the `axes`, and the N2V convenience function ensures that the algorithm is of the same dimension as the data config.

src/careamics/config/ng_factories/algorithm_factory.py
jdeschamps: Yeah, we can get rid of it (done!).

But this way does not work (if users are being unreasonable) because it will break if the same parameter is passed to `model_params` (multiple same parameters passed). So I stuck with the building a dict.

src/careamics/config/ng_factories/algorithm_factory.py
jdeschamps: Yes, CARE and N2N aglorithm configs have a validator as annotation to their model:

```
    model: Annotated[
        UNetConfig,
        AfterValidator(model_without_n2v2),
        AfterValidator(model_without_final_activation),
    ]
```

tests/config/ng_factories/test_n2v_factory.py
jdeschamps: Correct! Fixed.

src/careamics/config/ng_factories/algorithm_factory.py
jdeschamps: I should have made it explicit, being too quick.

Yes, I specifically did not want to change anything about the old configuration for a few reasons:
- I am not sure this module structure will hold in time (but also API, configurations, convenience functions), now it is small enough that we can think about how to best implement it.
- I wanted this PR to be as small as possible, with follow-up PR for refactoring the convenience functions in a manageable code base.
- I expected changes to their API alongside not everything being compatible (typically PN2V and all LVAE) and it's easier to maintain back-compatibility this way. Note that now this point is vindicated sine now the algorithm convenience function has changed API (but it didn't before...).

All is all I thought it was cleaner. Especially because the convenience function module is atrocious and the tests all over the place.

src/careamics/config/ng_factories/training_factory.py
jdeschamps: Here probably none, but see my other comment (https://github.com/CAREamics/careamics/pull/662#discussion_r2635899608)

src/careamics/config/ng_factories/algorithm_factory.py
melisande-c: Makes sense

src/careamics/config/ng_factories/algorithm_factory.py
melisande-c: aah, I see

tbf though there is an argument that `model_params` shouldn't be used by most users because the defaults are good and therefore maybe it shouldn't be in the convenience function? Anyone who wants to experiment with different model params should create the config themselves by initialising the pydantic class. But let's not change this now.

src/careamics/config/data/ng_data_config.py
```diff
@@ -783,6 +783,32 @@ def set_means_and_stds(
...
+        data, "Z" is in the axes or "T" is in the axes and patching `patch_size` has
+        3 dimensions.
+
+        Returns
+        -------
+        bool
+            True if the data is 3D, False otherwise.
+        """
+        if self.data_type == "czi":
+            if hasattr(self.patching, "patch_size"):
```
melisande-c: just for my understanding, is the only patching strategy without a `patch_size` attribute, the "whole patching" strategy?

I have a feeling that whether the data is 3D should be determined from the supplied axes, and then the number of dimensions in the patch size should be validated against it. Feel free to disagree if that doesn't make sense, maybe there's something in the czi axes combinations that I'm not seeing

melisande-c (approved):
Looks good :)

I just have one last thought on how the data is determined to be 3D: https://github.com/CAREamics/careamics/pull/662#discussion_r2661051923

src/careamics/config/data/ng_data_config.py
jdeschamps: Yes, it is "whole patching" is the only strategy with `patch_size` attribute.

You are right, there is a validation function for dimensions (patch size if present vs axes), and there is not point in checking it again. So axes should be the only thing used (if my understanding of czi is correct, because if we want to also accept 2D with T, then patch size is the only thing that can tell us the true dimension of czi format).

jdeschamps: I simplified this function and fixed a validation error in `validate_dimensions` (pushing after checking the tests).

PR #678, opened 2026-01-15
Fix: Struct N2V 3D

### LINKED ISSUE #676: [Question]struct_n2v 3D patch size
## Question

[question here]
Hi, I am trying to use the struct_n2v in careamics for 3d stacks, here is the configuration I use in notebook: 

config = create_n2v_configuration(experiment_name = "root_soil_image_denoise_P6_26h",
                                  data_type = "tiff",
                                  axes = "ZYX",
                                  patch_size = [64,128,128],
                                  batch_size = 2,
                                  trainer_params={"accelerator": "gpu", "devices": 1},
                                  num_epochs = 150,
                                  #use_n2v2 = True,
                                  struct_n2v_axis="vertical",
                                  struct_n2v_span=5
                                  )

This worked when disable the struct_n2v options, but with struc_n2v on, it always stopped the training with error: RuntimeError: a Tensor with 64 elements cannot be converted to Scalar. Could I get any guide here what I should do next?

melisande-c:
## Description

> [!NOTE]  
> **tldr**: This PR fixes struct N2V masking for 3D data.

The case of 3D data was not taken into consideration when calculating the min and max of the patch, in the `_apply_struct_mask_torch` function. The min and max is used to create the uniform distribution for pixel masking.

Additionally the test named `test_apply_struct_mask_3D_torch` was not testing 3D data but instead 2D data.

Fixed the problem by taking the min and max of a view of the patch, with the batch dimension remaining as the first dimension and then all the spatial dimensions are flattened together. This is because in torch you cannot take the min and max over multiple dimensions.

To cover the 3D case in the tests, I removed the old `test_apply_struct_mask_3D_torch` and instead added the 3D cases to `test_apply_struct_mask_torch`. The test had to be modified slightly so that it would work with both 2D and 3D data.

- `_apply_struct_mask_torch` 
- `test_apply_struct_mask_torch`

### Removed features or files

- `test_apply_struct_mask_3D_torch`

## How has this been tested?

Added 3D cases to `test_apply_struct_mask_3D_torch`

## Related Issues

- Resolves #676

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/transforms/pixel_manipulation_torch.py
+++ b/src/careamics/transforms/pixel_manipulation_torch.py
@@ -64,12 +64,13 @@ def _apply_struct_mask_torch(
     )
     mix = mix[valid_indices]
 
-    mins = patch.min(-1)[0].min(-1)[0]
-    maxs = patch.max(-1)[0].max(-1)[0]
+    # flatten in spatial dims to find min and max for each patch in the batch
+    batch_mins = patch.view(patch.shape[0], -1).min(dim=-1).values
+    batch_maxs = patch.view(patch.shape[0], -1).max(dim=-1).values
     for i in range(patch.shape[0]):
         batch_coords = mix[mix[:, 0] == i]
-        min_ = mins[i].item()
-        max_ = maxs[i].item()
+        min_ = batch_mins[i].item()
+        max_ = batch_maxs[i].item()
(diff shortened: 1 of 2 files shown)


jdeschamps (approved):
Awesome, thanks @melisande-c !

PR #686, opened 2026-01-19
Chore: improve error message when patch_size failed validation

jdeschamps:
## Description

As reported by users, when `patch_size` fails validation (e.g. passing an invalid `new_patch_size` to `NGDataConfig.convert_mode`), Pydantic models might throw a key error rather than an explicit validation error. This is due to the key `patch_size` being absent after failed validation.

This PR guards against that error and raises an error with an informative message.


melisande-c (approved):
Do you know why the error from validating the patch size in `_PatchedConfig` is not raised before reaching this point? That would be more desirable, no?

jdeschamps:
> m validating the patch size in `_PatchedConfig` is not raised before reaching this point? That would be more desirable, no?

We briefly discussed it, but just for the public eye: this is how Pydantic works, as it validates all fields in order but does not stop at the first error (`ValueError` or `AssertionError` only) to give users a complete report on parameters failure. 

I could not really think of a useful illustration, but here is a snippet showing that the error report has multiple errors:

```python
from pydantic import BaseModel, field_validator

class TestModel(BaseModel):

    age: int
    weight: float

    @field_validator("age")
    def check_age(cls, value):
        if value < 0:
            raise ValueError("Age must be non-negative")
        return value
    
    @field_validator("weight")
    def check_weight(cls, value):
        if value <= 0:
            raise ValueError("Weight must be positive")
        return value
    

a = TestModel(age=-2, weight=-70.5)
```

``` python
ValidationError: 2 validation errors for TestModel
age
  Value error, Age must be non-negative [type=value_error, input_value=-2, input_type=int]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
weight
  Value error, Weight must be positive [type=value_error, input_value=-70.5, input_type=float]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
```

Any other error than a `ValueError` or `AssertionError` is not caught by Pydantic, and results in the validation errors not being shown (because validation is interrupted the moment this error is raised).

PR #698, opened 2026-01-23
Switch from mypy to basedpyright

veegalinova:
## Description

> [!NOTE]  
> **tldr**: Switch from mypy to basedpyright for better type checking 

When working on a new feature, I got some super annoying mypy false positive errors that were not informative and did not actually protect the code from any real issues. 
Currently, many people have migrated to `pyright` instead. In general, it has more coverage and strictness than `mypy`, should be faster, and (in my opinion) captures more important stuff than mypy does - more on differences [here](https://docs.basedpyright.com/dev/usage/mypy-comparison/).

### Overview - what changed?

I replaced `mypy` with `basedpyright`. Why `basedpyright` and not the regural one? 
`basedpyright` is a community fork of pyright that implements some neat features useful to us. 
The main one is `baseline` that helps with gradual migration. Instead of getting hundreds of new type errors, we fix the current state of code as "acceptable".  Only new code must pass strict type checks. All unresolved stuff is stored in `.basedpyright/baseline.json`, which we can resolve gradually. 
Other improvements in basedpyright vs standard pyright [here](https://docs.basedpyright.com/dev/). 

1. Added `basedpyright` to dev dependencies
2. Created `pyrightconfig.json` with strictness settings
3. Generated baseline file with 432 current errors. 
4. Updated pre-commit hooks to use basedpyright instead of mypy

### New features or files

- `.basedpyright/baseline.json` - snapshot of current unresolved errors 
- `pyrightconfig.json` - type checking configuration
- `.pre-commit-config.yaml` - replaced mypy hook with basedpyright hook

- `pyproject.toml` - added `basedpyright` to dev dependencies

### Removed features or files

- mypy configuration file and `pyproject.toml` section

## Breaking changes

None for users. 

For developers: 
- run `pre-commit install` to update local hooks
- for vscode/cursor highlighting - install the `basedpyright` extension and disable the default pylance/pyright extension
- all new code must pass basedpyright checks (errors not in baseline will fail pre-commit)

---

- [x] Pre-commit passes


melisande-c:
I like it! I have been getting annoyed with mypy recently on some stuff to do with `TypeVar` narrowing that pyright can do correctly but mypy can't. 

It seems though, when run as a pre-commit on github the .env no longer exists, there might be a way to create the env first or we could skip the hook on the CI and instead have a different CI job for `basepyright` as documented here: https://docs.basedpyright.com/latest/benefits-over-pyright/improved-ci-integration/

pyrightconfig.json
```diff
@@ -0,0 +1,30 @@
...
+  ],
+  "exclude": [
+    "src/careamics/lvae_training",
+    "src/careamics/models/lvae",
+    "src/careamics/config/likelihood_model.py",
+    "src/careamics/losses/loss_factory.py",
+    "src/careamics/losses/lvae/losses.py",
+    "**/*.ipynb"
+  ],
+  "pythonVersion": "3.12",
```
melisande-c, line 13: Also we currently support from 3.11 so we should probably use that python version

veegalinova:
Closing this, will re-assess if newer tooling like [ty](https://docs.astral.sh/ty/) or [pyrefly](https://github.com/facebook/pyrefly) is a better fit now

PR #699, opened 2026-01-23
Feat: `torchmetrics`-compatible (scale invariant) PSNR implementations

### LINKED ISSUE #568: [Refactoring] Add torchmetrics compatible scale invariant PSNR metric
## Description

In #565 we are pinning the `torchmetrics` version to avoid a new `data_range` parameter error. 
I think that it is beneficial to keep `torchmetrics` around to have an extensible list of available metrics with a simple API. But, we can not use their `data_range` parameter because we do not know the correct range at the initialization time. 

So I propose to create a metric based on the existing `scale_invariant_psnr` that will be compatible with torchmetric's `MetricCollection`. Glancing at the [tutorial](https://lightning.ai/docs/torchmetrics/stable/pages/implement.html) it seems straightforward enough, a couple of methods will need to be added. 

## Where in the code
`MetricCollection` and `PSNR` metrics are initialized in lightning module `UnetModule`, classes that inherit from it can call it during validation to log metrics.

## References

Reference issue #550 
Current torchmetrics usage: `src/careamics/lightning/dataset_ng/lightning_modules/unet_module.py`

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Add implementations of PSNR and scale-invariant PSNR that are compatible with `torchmetrics`.

NG CAREamics Lightning modules will probably use `torchmetrics.MetricCollection` and we need an implementation of the PSNR that is compatible with 3D and channels. Unfortunately, `torchmetrics.image.PeakSignalNoiseRatio` is not compatible with `Z` and requires a global `data_range` value in the latest `torchmetrics` versions.

### Overview - what changed?

This PR introduces three implementations of the PSNR:
- `RangefreePSNR`: similar to `skimage` and `torchmetrics` PSNR, but with `data_range` equal to `max-min` (local range) and applied to each patch.
- `LocalSIPSNR`: similar to `juglab/ScaleInvPSNR`, where target and prediction are normalized to a similar range and the PSNR is computed with the local range (`max-min` of the patch).
- `GlobalSIPSNR`: similar to `LocalSIPSNR`, albeit with a global range (over all patches). Note that it sums `log10(mse)`, so there is a chance it may explode in value since rescaled `mse` is expected to be small.

I still need to benchmark on examples whether using a global range is useful (e.g. minimizes contribution of background patches).

Note: the PSNRs require the number of channels at initialization. As this is part of the algorithm configuration, this information is available in the lightning modules we implemented.

Fully reimplemented the ScaleInvariantPSNR and PSNR to compute the metrics per channel. 

### New features or files

- `psnr.py` with all metrics.

## How has this been tested?

Tests added, testing against previous implementation, `skimage` and `torchmetrics`.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/568

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+from typing import Any
+
+import torch
+from torch import Tensor, tensor
+from torchmetrics import Metric
+
+# Note: LVAE use a PSNR implementation that is averaging over pixels by excluding NaNs,
+# and counting the number of non-NaN pixels for averaging. Current choice is to ignore
+# that implementation since NaNs should not be present in the model, being a sign of a
+# more fundamental issue with the training.
```
veegalinova: Fully agree!

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # fix range of gt and prediction
+        tar_rescaled, pred_rescaled = normalise_range(target, preds)
+
+        # compute mse
+        mse = torch.mean((tar_rescaled - pred_rescaled) ** 2 + self.eps, dim=dims)
+
+        # update states
+        self.glob_max: torch.tensor = torch.maximum(self.glob_max, batch_max)
+        self.glob_min: torch.tensor = torch.minimum(self.glob_min, batch_min)
```
veegalinova:
```suggestion
        self.glob_max: torch.Tensor = torch.maximum(self.glob_max, batch_max)
        self.glob_min: torch.Tensor = torch.minimum(self.glob_min, batch_min)
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+        self.mse_log += torch.log10(mse.prod(dim=0) + self.eps)
```
veegalinova: Apparently, if the batch is big enough there is an overflow risk for mse.prod(), a better formulation would be through sum:
```suggestion
        self.mse_log += torch.log10(mse).sum(dim=0)
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+        self.total += batch_size
```
veegalinova:
```suggestion
        self.mse_log: torch.Tensor = self.mse_log + torch.log10(mse).sum(dim=0)
        self.total: torch.Tensor = self.total + batch_size
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # normalize range of gt and prediction
+        tar_rescaled, pred_rescaled = normalise_range(target, preds)
+
+        # compute mse
+        mse = torch.mean((tar_rescaled - pred_rescaled) ** 2 + self.eps, dim=dims)
+
+        # update states
+        self.psnr_sum += torch.sum(10 * torch.log10(data_range**2 / mse), dim=0)
+        self.total += batch_size
```
veegalinova:
```suggestion
        self.psnr_sum: torch.Tensor = self.psnr_sum + torch.sum(10 * torch.log10(data_range**2 / mse), dim=0)
        self.total: torch.Tensor = self.total + batch_size
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+        batch_min = torch.amin(target, dim=dims)
+        batch_max = torch.amax(target, dim=dims)
+        data_range = batch_max - batch_min + self.eps
+
+        # compute mse
+        mse = torch.mean((target - preds) ** 2 + self.eps, dim=dims)
+
+        # update states
+        self.psnr_sum += torch.sum(10 * torch.log10(data_range**2 / mse), dim=0)
+        self.total += batch_size
```
veegalinova:
```suggestion
        self.psnr_sum: torch.Tensor = self.psnr_sum + torch.sum(10 * torch.log10(data_range**2 / mse), dim=0)
        self.total: torch.Tensor = self.total + batch_size
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+    Attributes
+    ----------
+    psnr_sum : Tensor
+        Sum of PSNR values for each channel.
+    total : Tensor
+        Total number of samples processed.
+    """
+
+    def __init__(self, n_channels: int, **kwargs: Any):
+        """Initialize the scale invariant PSNR metric.
```
veegalinova:
```suggestion
        """Initialize the PSNR metric.
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+        data_range = batch_max - batch_min + self.eps
+
+        # compute mse
+        mse = torch.mean((target - preds) ** 2 + self.eps, dim=dims)
+
+        # update states
+        self.psnr_sum += torch.sum(10 * torch.log10(data_range**2 / mse), dim=0)
+        self.total += batch_size
+
+    def compute(self) -> Tensor | dict[str, Tensor]:
```
veegalinova:
```suggestion
    def compute(self) -> Tensor:
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+        tar_rescaled, pred_rescaled = normalise_range(target, preds)
+
+        # compute mse
+        mse = torch.mean((tar_rescaled - pred_rescaled) ** 2 + self.eps, dim=dims)
+
+        # update states
+        self.psnr_sum += torch.sum(10 * torch.log10(data_range**2 / mse), dim=0)
+        self.total += batch_size
+
+    def compute(self) -> Tensor | dict[str, Tensor]:
```
veegalinova:
```suggestion
    def compute(self) -> Tensor:
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
```diff
@@ -0,0 +1,336 @@
...
+        # compute mse
+        mse = torch.mean((tar_rescaled - pred_rescaled) ** 2 + self.eps, dim=dims)
+
+        # update states
+        self.glob_max: torch.tensor = torch.maximum(self.glob_max, batch_max)
+        self.glob_min: torch.tensor = torch.minimum(self.glob_min, batch_min)
+        self.mse_log += torch.log10(mse.prod(dim=0) + self.eps)
+        self.total += batch_size
+
+    def compute(self) -> Tensor | dict[str, Tensor]:
```
veegalinova:
```suggestion
    def compute(self) -> Tensor:
```

src/careamics/lightning/dataset_ng/metrics/psnr.py
veegalinova, line 1: I think that metric names could be improved slightly to be closer to "usual" namings, but 
I also struggle to find perfect ones. My best guess - `GlobalSIPSNR`, `SampleSIPSNR`, `SamplePSNR`.

veegalinova (commented):
Added a few suggestions to make mypy happy, and one important comment about potential overflow, otherwise all good!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Makes sense!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: I like the suggestions, let me think further about it!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Done!

src/careamics/lightning/dataset_ng/metrics/psnr.py
jdeschamps: Kept your names, they are good!

jdeschamps:
I implemented the suggestions, and gathered all global and per-sample PSNRs into two implementations with optional parameter `use_range_invariance`. Also added tests for global PSNR (yay a fourth possibility!). Finally, I renamed the global SIPSNR to `SIPSNR`. 

Let's use that one as default for now. But we should investigate further on our benchmarks and examples how they behave.

veegalinova (approved):
All good! I added a commit switching the default metric to SIPSNR in the lightning modules.

jdeschamps:
> All good! I added a commit switching the default metric to SIPSNR in the lightning modules.

Thats good, because it also led to errors in the CI:

```
ValueError: The metric `tensor([10.8681, 10.6651, 11.5457])` does not contain a single element, thus it cannot be converted to a scalar.
```

To be honest, I was kind of expecting issues with the per channel value, and I just completely forgot about it... Obviously, testing it with the metrics collection was not enough. Probably something I missed in the Lightning docs.

We probably need to return a dict?

jdeschamps:
> > All good! I added a commit switching the default metric to SIPSNR in the lightning modules.
> 
> Thats good, because it also led to errors in the CI:
> 
> ```
> ValueError: The metric `tensor([10.8681, 10.6651, 11.5457])` does not contain a single element, thus it cannot be converted to a scalar.
> ```
> 
> To be honest, I was kind of expecting issues with the per channel value, and I just completely forgot about it... Obviously, testing it with the metrics collection was not enough. Probably something I missed in the Lightning docs.
> 
> We probably need to return a dict?

Problem is that `module.log_dict` expects metrics to return a single scalar, which is an issue with the current implementation because it may return a tensor of size N (N being the number of channels).

`torchmetrics.ClasswiseWrapper` is a solution to return a dict for each channel. But that is not enough: https://github.com/Lightning-AI/pytorch-lightning/pull/21507

So the only solution is to either average over channels, or log manually / recursion on the metrics.

jdeschamps:
Actually, there seems to be only two ways in Lightning to deal with that: https://github.com/Lightning-AI/torchmetrics/issues/2683

- Either we make a metrics per channel
- Or we do manual logging, which requires additional methods to be implemented for each Lightning module, and taking care of `update`, `compute` and `reset` ourselves.

jdeschamps:
For clarity, I decided in the end to add the possibility to compute the averaging over all channels, and made it default (`output_channel=-1`).

The possibility to compute on a particular channel is available by setting `output_channel` to that particular value.

Finally, in the lightning modules, I added a SIPSNR metric for each channel (rather than the average).

veegalinova (approved):

PR #701, opened 2026-01-27
Refac(Next-Gen Dataset): Remove inheritance from lightning modules

veegalinova:
## Description

> [!NOTE]  
> **tldr**: 
Refactored Lightning modules to use composition over inheritance by extracting shared utilities and removing the base `UnetModule` class.

The modules were using inheritance from `UnetModule`, which created tight coupling and made the code harder to maintain. Each algorithm module inherited shared functionality like logging and checkpoint loading.
Even though now there is more code duplication, I think it's reasonable for lightning modules, since they have a fixed structure. It is now easier to understand the module as a whole without jumping between multiple files. 

### Overview - what changed?

Removed the UnetModule base class and extracted common functionality into utility functions. Both `CAREModule` and `N2VModule` now inherit directly from `L.LightningModule` and use shared utilities from `module_utils.py`.

### New features or files

- `module_utils.py` with `log_training_stats()`, `log_validation_stats()`, and `load_best_checkpoint()` utilities

- `care_module.py`, `n2v_module.py` - inherit from `L.LightningModule` only

### Removed features or files

- base module `unet_module.py`

## How has this been tested?

Existing tests should pass without changes since the functionality remains the same, just refactored.

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/lightning_modules/care_module.py
```diff
@@ -71,27 +108,94 @@ def training_step(
...
+        dict[str, Any]
+            A dictionary containing the optimizer and learning rate scheduler.
+        """
+        optimizer_func = get_optimizer(self.config.optimizer.name)
+        optimizer = optimizer_func(  # type: ignore[operator]
+            self.model.parameters(), **self.config.optimizer.parameters
+        )
+
+        scheduler_func = get_scheduler(self.config.lr_scheduler.name)
+        scheduler = scheduler_func(optimizer, **self.config.lr_scheduler.parameters)  # type: ignore[operator]
```
jdeschamps: The content of this method looks like it could be refactored into a different function called here.

src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py
```diff
@@ -25,82 +31,159 @@ class N2VModule(UnetModule):
...
-        batch_idx : Any
-            The index of the current batch in the training loop.
+        batch_idx : int
+            The index of the current batch in the validation loop.
         """
         x = batch[0]
-
-        x_masked, x_original, mask = self.n2v_manipulate(x.data)
+        x_data = cast(torch.Tensor, x.data)
+        x_masked, x_original, mask = self.n2v_manipulate(x_data)
```
jdeschamps, line 122: A note to consider for an issue -> I just realized that in `n2v` the masking was pre-computed ([see here](https://github.com/juglab/n2v/blob/c61b3b47c5cacc533f7e0e3db7cefab139d7b62a/n2v/models/n2v_standard.py#L247)), so that:
- it would be faster
- the validation loss is comparable across epochs (since the real train/val data are the masked pixels not the rest of the patch)

jdeschamps (approved):

jdeschamps:
Looks good, thanks!

One minor comment on further refactoring, and one to consider for an issue maybe... At least to keep in mind.

src/careamics/lightning/dataset_ng/lightning_modules/care_module.py
veegalinova: Done

src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py
veegalinova: Opened #703

PR #702, opened 2026-01-27
Fix: prevent file overwriting when predicting with `CachedTiles` and `WriteImage`

### LINKED ISSUE #693: Next-Gen Dataset: `CachedTiles` and `WriteImage` may overwrite files for arrays
## Describe the bug

Since `ImageRegionData.source` is `array` for all arrays, the resulting tiff may be overwritten (or fail) by both `CachedTiles` and `WriteImage` as the file name is created using source.

See:

https://github.com/CAREamics/careamics/blob/13406ec4b98919c683e2af72072921d98c3dc18f/src/careamics/lightning/dataset_ng/callbacks/prediction_writer/cached_tiles_strategy.py#L142

https://github.com/CAREamics/careamics/blob/13406ec4b98919c683e2af72072921d98c3dc18f/src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_image_strategy.py#L83

https://github.com/CAREamics/careamics/blob/13406ec4b98919c683e2af72072921d98c3dc18f/src/careamics/lightning/callbacks/prediction_writer_callback/file_path_utils.py#L31

## Reproduce

Here is how I think we can check whether there is indeed a problem:

``` python
# Training
from pathlib import Path
import numpy as np
from careamics.config.ng_factories import create_n2v_configuration
from careamics.lightning.dataset_ng.data_module import CareamicsDataModule
from careamics.lightning.dataset_ng.lightning_modules import N2VModule
from careamics.lightning.dataset_ng.callbacks.prediction_writer import PredictionWriterCallback

data = rng.normal(size=(10, 64, 64)).astype(np.float32)

cfg = create_n2v_configuration(
    experiment_name="my_experiment",
    data_type="array",
    axes="SYX",
    patch_size=(32, 32),
    batch_size=2,
    num_epochs=1
)

datamodule = CareamicsDataModule(
    data_config=cfg.data_config,
    train_data=[data[:-1]],
    val_data=[data[-1:]],
)

model = N2VModule(cfg.algorithm_config)

pred_callback = PredictionWriterCallback(dirpath=Path("predictions"))

trainer = Trainer(callbacks=[pred_callback], **cfg.training_config.lightning_trainer_config)
trainer.fit(model, datamodule=datamodule)

# prediction
pred_callback.set_writing_strategy("tiff", tiled=True)

pred_cfg = cfg.data_config.convert_mode(
    new_mode="predicting",
    new_patch_size=(32, 32), # necessary for tiling
    overlap_size=(16, 16),
)
datamodule = CareamicsDataModule(
    data_config=pred_cfg,
    pred_data=[
        data[i:i+2] for i in range(data.shape[0] // 2)
    ],
)

# predict and write to the disk
trainer.predict(model, datamodule=datamodule, return_predictions=False)
```

Expected behaviour would be to have 5 tiff files created, but by reading the code it seems it would every time overwrite "array.tiff".

## Solution

- Better handling of `source` when it is equal to `array`
    - Is `write_tiff` throwing an error when the data already exists? if not it should probably raise `FileExistsError`
    - `WriteStrategy` (here `CachedTile` and `WriteImage`) are in charge of creating the destination path
    - `create_write_file_path` could take an optional str argument `` to be added to the filename before the extension (e.g. `create_write_file_path(dir, "array", "tiff", postfix=i)` -> "/dir/array_{i}.tiff")

diyasrivas:
## Description

**tldr**: Fixed filename collision when predicting on multiple numpy arrays by adding unique postfix identifiers to output filenames.

When predicting on multiple numpy arrays, all `ImageRegionData` instances have `source="array"` since there's no file path for in-memory data. Both writing strategies (`CachedTiles` and `WriteImage`) construct output filenames from this source, causing all predictions to write to `array.tiff`, resulting in silent overwrites.

### Overview - what changed?

Added a postfix mechanism to disambiguate array-source filenames:
- `file_path_utils.py`: Added optional `postfix` parameter to `create_write_file_path()`
- `cached_tiles_strategy.py`: Detects array sources and uses `data_idx` from `region_spec` as postfix
- `write_image_strategy.py`: Detects array sources and uses `data_idx` from `region_spec` as postfix 
- `tiff.py`: Added `FileExistsError` check to prevent silent overwrites

`WriteImage.write_batch()` is called multiple times (once per batch), so using the enumerate index `i` fails because it resets each batch. Instead, we use the persistent `data_idx` from `predictions[i].region_spec["data_idx"]`, which remains consistent across all batches.

```python
# CachedTiles (_stitch_and_write_single):
postfix = ""
if source.stem == "array":
    data_idx = tiles[0].region_spec["data_idx"]
    postfix = f"_{data_idx}"

# WriteImage (write_batch):
postfix = ""
if source_path.stem == "array":
    data_idx = predictions[i].region_spec["data_idx"]
    postfix = f"_{data_idx}"

# Both produce: array_0.tiff, array_1.tiff, etc.
```

### Modified files
- `file_path_utils.py`: Added `postfix: str = ""` parameter to `create_write_file_path()`
- `cached_tiles_strategy.py`: Uses `data_idx` from `region_spec` as postfix for array sources
- `write_image_strategy.py`: Uses `data_idx` from `region_spec` as postfix for array sources
- `tiff.py`: Added `FileExistsError` check before writing

## How has this been tested?

Created minimal reproduction tests for both tiled and non-tiled prediction:
```python
# Test setup: 5 arrays from 10×64×64 data
pred_data_list = [data[i:i+2] for i in range(0, 10, 2)]  # 5 arrays

# Tiled prediction
pred_callback.set_writing_strategy("tiff", tiled=True)
trainer.predict(model, datamodule=datamodule)

# Non-tiled prediction  
pred_callback.set_writing_strategy("tiff", tiled=False)
trainer.predict(model, datamodule=datamodule)

# Verification
tiff_files = list(output_dir.glob("*.tiff"))
assert len(tiff_files) == 5
assert {f.name for f in tiff_files} == {"array_0.tiff", ..., "array_4.tiff"}
```

**Results:**
Before: 1 file (`array.tiff` overwritten 5 times)
After: 5 unique files (`array_0.tiff` through `array_4.tiff`)

So now, both tiled and non-tiled predictions now generate unique files per array.

## Related Issues

Resolves #693

## Breaking changes

None. The `postfix` parameter defaults to `""`.

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/callbacks/prediction_writer/file_path_utils.py
```diff
@@ -20,14 +20,15 @@ def create_write_file_path(
         The original file path.
     write_extension : str
         The extension that output files should have.
+    postfix : str, optional
+        Appends to filename before extension, default is empty string.
```
jdeschamps: Just a suggestion to adhere to our docstring conventions (which we don't always follow...)

```suggestion
    postfix : str, default=""
        Appends to filename before extension.
```

jdeschamps (changes requested):
Great thanks! I have one minor comment, but tests are failing and probably need a small adjustment: https://github.com/CAREamics/careamics/actions/runs/21408420567/job/61638605881?pr=702

jdeschamps (approved):

PR #710, opened 2026-01-30
Feature(NG Dataset): Stratified Patching Strategy

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Introducing a stratified patching strategy which has an additional feature that allows regions to be excluded from sampling.

There are two possible applications for this strategy:
1) Validation and train splitting when there is only 1 or few images.
2) A more efficient way to apply a background mask.

It also ensures a more consistent coverage of the data per epoch than the random patching.

### Overview - what changed?

A new `StratifiedPatchingStrategy` class, added without integration into the CAREamics Dataset. Internally this class depends on a `_ImageStratifiedPatching` which in turn depends on a `_SamplingRegion` class.

As mentioned, there is also the feature to exclude patch regions from being sampled, and the number of patches is calculated to be `ceil(prod(shape / patch_size)) - n_excluded_patches`.

The `StratifiedPatchingStrategy` is constructed such that, the mean expected value of a pixel being selected per epoch is approximately equal to 1. The main difference with the random patching strategy, which has a mean expected value slightly greater than 1, is that the number of patches is calculated to be `prod(ceil(shape / patch_size))` where the ceiling is taken before the product.

A demo notebook has also been included.

Apologies it got a bit complicated but I will try my best to explain. I have also tried to include explanatory comments in the code.

#### Structure overview

Sampling regions which have an area of double the patch size are created so that they lie on a grid with a spacing equal to the patch size. This ensures that all possible patch coordinates can be selected. These sampling regions are represented by the `_SamplingRegion` class. The sampling region itself contains subregions, which make it easier to exclude patch regions.

For each sample in each image-stack an `_ImageStratifiedPatching` class is created, this stores `_SamplingRegions` in a `dict[tuple[int, ...], _SamplingRegion]` where the key of the dictionary represents the grid coordinate that the top-left corner of the sampling region lies on. Most of the patching logic happens in this class.

The `StratifiedPatchingStrategy` stores a `list[list[_ImageStratifiedPatching]]` where the elements of the outer list represent each image stack and there is an `_ImageStratifiedPatching` instance for each sample in each image stack.

#### Bin packing sampling regions

Not all the sampling regions will have the same area, unless the image is evenly divisible by the patch size and no patch regions have been excluded. To ensure that the smaller regions have an appropriate probability of being sampled from, they are combined together into bins. The bin packing is done using the best-fit-decreasing algorithm, courtesy of Claude.

Each bin is the same size and each dataset index corresponds to a bin. This means that each possible patch coordinate in each bin has the same probability of being selected.

An implementation detail is that a bin might not be filled up perfectly, and it is also possible to get empty bins, this extra probability is used to select all regions in the sample.

#### Sampling a patch coordinate

When a patch spec is sampled for a particular index:
1) the `StratifiedPatchingStrategy` maps the index to the correct image stack and sample, similarly to how it is done in the random patching strategy. 
2) Then the index is given to the relevant `_ImageStratifiedPatching`; this index maps to a particular bin, which contains a set of `_SamplingRegions`. 
3) A `_SamplingRegion` is chosen from the bin with a probability equal to the ratio of the region's area to the bin size.
4) A patch coordinate is sampled from the valid patch coordinates contained within the sampling region.

#### Excluding patch regions

The `SamplingRegion` contains sub regions, for the 2D case these are 4 regions illustrated by the diagram below

```
 ┌───┬──────────────────────────┐
 │   1                          │
 ├─1─┼─────(patch_size - 1)─────┤
 │   │                          │
 ⋮              ⋮
 └───┴──────────────────────────┘
```
If I want to exclude the bottom-left quadrant I exclude the large bottom left region, if I want to exclude the top-left quadrant I exclude both the bottom-left region and the top-left region.

To exclude a patch region from being sampled I have to exclude the correct quadrant from the 4 sampling regions that cover it.

#### Clipping sampling regions

Sampling regions can also be clipped, this is necessary at the edges of images which do not have an image size which is evenly divisible by the patch size. This is handled by the method `_SamplingRegion.clip`.

### New features or files

- All new classes have been added to `src/careamics/dataset_ng/patching_strategies/stratified_patching.py`
  - `StratifiedPatchingStrategy`
  - `_ImageStratifiedPatching`
  - `_SamplingRegion`

## How has this been tested?

Stratified patching has been added to the `test_all_patching_strategies` file. This tests:
- `test_all_get_patch_spec`
- `test_patches_cover_50percent`
- `test_get_patch_indices`

One new test has been added for the StratifiedPatchingStrategy that ensures that the excluded patches are never sampled.

The included notebook produces the following figures and also demonstrates that mean expected value that a pixel is selected per epoch approximately 1.

59ea9e63-7749-4897-b86d-de30f6e5a0bf
45baf19c-c8db-4d55-8d5c-e2e69b8ea6f2
f9dd9718-e236-406c-bd4e-4427546e7842
41466591-24ef-4394-9c86-a7a0ff57bf52

---

- [x] Pre-commit passes


src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+    Patches will be sampled from sampling regions that are two times the patch size in
+    each dimension. Some sampling regions may be smaller than this because they are on
+    the edge of an image or because a nearby patch has been excluded.
+
+    If the same index is used twice to sample a patch with the method `get_patch_spec`
+    there will be a high probability that it will come from the same sampling region,
+    but not necessarily 100%. Smaller sampling regions may be binned together into a
+    single index. The mean of all the expected values that each pixel will be selected
+    in a patch per epoch is 1.
```
melisande-c, line 43: Depends what you mean, the strategy is seeded once at initialisation, if you sample the same index twice with the same instance you will get different patches. If you make a new instance with the same seed you will get the same patch

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+    """
+    A stratified patching strategy that also allows patches on a grid to be excluded.
+
+    Patches will be sampled from sampling regions that are two times the patch size in
+    each dimension. Some sampling regions may be smaller than this because they are on
+    the edge of an image or because a nearby patch has been excluded.
+
+    If the same index is used twice to sample a patch with the method `get_patch_spec`
+    there will be a high probability that it will come from the same sampling region,
+    but not necessarily 100%. Smaller sampling regions may be binned together into a
```
jdeschamps, line 41: Unless seed is set? (just checking, but we could add a comment)

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+            "data_idx": data_idx.item(),
+            "sample_idx": sample_idx.item(),
+            "coords": tuple(coord.tolist()),
+            "patch_size": self.patch_size,
+        }
+
+    # Note: this is used by the FileIterSampler
+    def get_patch_indices(self, data_idx: int) -> Sequence[int]:
+        """
+        Get the patch indices will return patches for a specific `image_stack`.
```
jdeschamps: Confusing sentence

```suggestion
        Return patches for a specific `image_stack`.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        ----------
+        data_idx : int
+            An index that corresponds to a given `image_stack`.
+
+        Returns
+        -------
+        sequence of int
+            A sequence of patch indices, that when used to index the `CAREamicsDataset
+            will return a patch that comes from the `image_stack` corresponding to the
+            given `data_idx`.
```
```suggestion
            A sequence of patch indices, used to index the `CAREamicsDataset`
            to return a patch that comes from the `image_stack` corresponding to the
            given `data_idx`.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        """
+        patches_per_sample = [
+            [sample.n_patches for sample in image] for image in self.image_patching
+        ]
+        patches_per_image = [sum(samples) for samples in patches_per_sample]
+        cumulative_image_patches = np.cumsum(patches_per_image)
+
+        start = 0 if data_idx == 0 else cumulative_image_patches[data_idx - 1]
```
jdeschamps: Isn't `cumulative_image_patches` already calculated and stored in `self.cumulative_image_patches`?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        )
+
+
+class _ImageStratifiedPatching:
+    """A class used to sample the patch coordinates for a single sample.
+
+    The number of patches is determined from the number of selectable patch coordinates.
+
+    Sampling regions have a size of 2 times the patch size in each dimension, unless
+    the region is near an edge or a nearby patch has been excluded.
```
```suggestion
    the region is near an edge or a nearby patch that has been excluded.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        self.bin_size: int
+        self.bins: list[list[tuple[int, ...]]]
+        self.n_patches: int
+
+        # populate the self.regions and self.areas dictionaries
+        for grid_coord in itertools.product(*[range(s) for s in self.grid_shape]):
+            # find pixel coord
+            coord = np.array(grid_coord) * np.array(patch_size)
+            sampling_region = _SamplingRegion(tuple(coord), self.patch_size, rng)
+            sampling_region.clip(np.zeros(self.ndims, dtype=int), np.array(shape))
```
jdeschamps, line 316: A bit meaningless optimization wise, but `np.array(patch_size)` and `np.array(shape)` could be defined at the beginning of the method and re-used (they are also being called line 262).

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        if index >= self.n_patches:
+            raise ValueError(
+                f"Index {index} out of bounds for image with {self.n_patches} patches."
+            )
+
+        # each index corresponds to a bin*,
+        # regions within the selected bin are sampled proportionally to their area.
+        # if there is remaining space in the bin it is assigned to all the regions
+        # *contradicting the first statement, there may be fewer bins than patches
+        # for empty bins, all regions are sampled with a prob proportional to area
```
jdeschamps: Not super clear to me, maybe let's discuss it!

jdeschamps: More specifically, in which case are there fewer bins than patches? isn't it always the case?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+        # sampling regions will be stored in a dict
+        # the keys correspond to a grid coordinate
+        self.regions: dict[tuple[int, ...], _SamplingRegion] = {}
+        self.areas: dict[tuple[int, ...], int] = {}
+        self.excluded_patches: list[tuple[int, ...]] = []
+        self.probs: dict[tuple[int, ...], float]
+        self.bin_size: int
+        self.bins: list[list[tuple[int, ...]]]
+        self.n_patches: int
```
jdeschamps, line 307: can we add some docstring to all members defined here to state what they hold? Reading the code I guess that `self.bins` will be a list of list of coords that can be used to extract regions from `self.regions`?

Would be useful to better describe these different members to make reading the code easier.

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+    def _sample_region_from_bin(self, bin: list[tuple[int, ...]]) -> "_SamplingRegion":
+        """
+        Sample a region from a given bin. Bins can contain multiple sampling regions.
+        """
+        keys = list(self.regions.keys())
+        probs = np.array([self.probs[key] for key in keys])
+        if len(bin) != 0:
+            indices = np.where(
+                (np.array(keys) == np.array(bin)[:, None]).all(2).any(0)
```
jdeschamps: Can you add a comment to explain `[:, None]).all(2).any(0)`, it is hard to decipher

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+    # Binary search bounds
+    min_size = max(areas.values())
+    max_size = sum(areas.values())
+
+    # Edge case: if we want 1 bin, we need bin size = sum of all values
+    if target_n_bins == 1:
+        return max_size, 1
+
+    # Edge case: if we want as many bins as values, bin size = max value
```
jdeschamps: Naming is bit confusing with `min_size` representing the max value

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
jdeschamps: Am I correct in assuming that each bin is either containing a single full region or one/multiple smaller regions (those close to edge or removed patch)?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        # ├─1─┼─────(patch_size - 1)─────┤
+        # │   │                          │
+        #   ⋮              ⋮
+        # └───┴──────────────────────────┘
+
+        subregion_axis_extent = [
+            ((0, 1), (1, patch_size[i])) for i in range(self.ndims)
+        ]
+
+        # a single subregion is represented by it's extent in each axis
```
```suggestion
        # a single subregion is represented by its extent in each axis
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        return self.rng.integers(start, end) + np.array(self.coord)
+
+    def exclude_orthant(
+        self,
+        orthant: tuple[Literal[0, 1], ...],
+    ) -> None:
+        """
+        Exclude an orthant from the sampling region.
+
+        (An orthant is a quadrant or octant for 2D and 3D respectively).
```
jdeschamps, line 579: ahah thanks!

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+    the edge of an image or because a nearby patch has been excluded.
+
+    If the same index is used twice to sample a patch with the method `get_patch_spec`
+    there will be a high probability that it will come from the same sampling region,
+    but not necessarily 100%. Smaller sampling regions may be binned together into a
+    single index. The mean of all the expected values that each pixel will be selected
+    in a patch per epoch is 1.
+
+    The number of patches is determined from the number of selectable patch coordinates.
+    """
```
jdeschamps, line 46: We should add some of your description from the PR here to provide more explanations.

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
jdeschamps, line 316: Is clipping necessary to cover the cases where the shape is not divisible by the patch size?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        n_patches : int
+            The number of patches.
+        bin_size : int
+            The size of the bins that sampling regions are packed into. Each bin
+            corresponds to a sampling index.
+        bins : list[list[tuple[int, ...]]]
+            Bins containing sampling regions. The regions are packed based on their
+            area.
+        probs : dict[tuple[int, ...], float]
+            The probability that a sampling region will be selected from it's bin.
```
```suggestion
            The probability that a sampling region will be selected from its bin.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+
+class _SamplingRegion:
+    """
+    Represent a small subregion that a patch can be sampled from.
+
+    The region is double the patch size in each dimension. Quadrants or octants, for
+    2D or 3D respectively, can be excluded from the region by calling the
+    `exclude_orthant` method.
+    """
```
jdeschamps, line 510: Can we add a description of the attributes?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
jdeschamps: Also what is "values" in this sentence? number of sampling regions?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+            bins.append([key])
+
+    return bins
+
+
+def _find_bin_size(
+    areas: dict[tuple[int, ...], int], target_n_bins: int
+) -> tuple[int, int]:
+    """Find the minimum bin size that will result in `target_n_bins` or less.
+
```
jdeschamps, line 707: Can you add a description of the parameters?

`areas` are the areas of the sampling regions?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        grid_coords : Sequence[tuple[int, ...]]
+            A sequence of 2D or 3D tuples. Each tuple corresponds to a grid coordinate
+            that will be excluded from sampling. The grid starts at (0, 0) and has a
+            spacing of the given `patch_size`.
+        """
+        self.excluded_patches.extend(grid_coords)
+        for grid_coord in grid_coords:
+            d: tuple[Literal[0, 1], ...] = (0, 1)
+            # exclude the patch from all the sampling regions that cover it
+            # These are the grid coords at: subtract (0, 0), (0, 1), (1, 0) (for 2D)
```
jdeschamps, line 420: Not (1,1)?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+            for d_idx in itertools.product(*[d for _ in range(self.ndims)]):
+                # q is the ID of the orthant to remove
+                q: tuple[Literal[0, 1], ...] = tuple(0 if i == 1 else 1 for i in d_idx)
```
jdeschamps, line 423: Also pretty hard to read, some examples in the comments would help

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+    # TODO: add method to return valid grid coords for removal
+
+    def exclude_patches(
+        self, data_idx: int, sample_idx: int, grid_coords: Sequence[tuple[int, ...]]
+    ):
+        """
+        Exclude patches from being sampled.
+
+        Excluded patches must lie on a grid which starts at (0, 0) and has a spacing of
+        the given `patch_size`.
```
jdeschamps, line 108: Should this raise an error?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
jdeschamps: Resolved

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+    def clip(
+        self,
+        minimum: np.ndarray[tuple[int], np.dtype[np.int_]],
+        maximum: np.ndarray[tuple[int], np.dtype[np.int_]],
+    ):
+        """
+        Clip a sampling region.
+
+        After clipping a selected patch will not overlap the `minimum` or `maximum`.
```
```suggestion
        After clipping a selected patch will not overlap with the `minimum` or `maximum`
        coordinate.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+
+        Parameters
+        ----------
+        minimum : numpy.ndarray[np.int_]
+            An array of length 2 or 3 that represents the minimum coordinate that a
+            patch can be.
```
```suggestion
            An array of length 2 or 3 that represents the minimum coordinate at
            which a patch can be defined.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -0,0 +1,698 @@
...
+        maximum : numpy.ndarray[np.int_]
+            An array of length 2 or 3 that represents the maximum coordinate that a
+            patch can be
```
```suggestion
            An array of length 2 or 3 that represents the maximum coordinate at
            which a patch can be defined.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: Might be a good idea, currently if it is an invalid index it is ignored

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: ah yep I will remove this calculation

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: yes I will write comment

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: that should be the case yes

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: no because final quadrant is only defined by next sampling region, I do however iterate over (1, 1) in the loop below anyway

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: I guess I meant fewer exclusive, rather than fewer than or equal. But yes the number of bins will always be less than or equal to the number of patches

melisande-c: I will try to reword this

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: This is only in the edge case where the requested number of bins is 1. In this case the size of the bin has to be equal to the sum of all the areas of the sampling regions

jdeschamps (approved):

PR #717, opened 2026-02-02
Feature: Next-Gen CAREamist API

### LINKED ISSUE #649: Next-Gen Dataset: base for `CAREamist`-like module as high-level API
## Problem

As the NG Dataset is taking shape, we should start outlining a replacement for the `CAREamist`, which should live in parallel for back compatibility.

This issue is linked with the creation of a more flexible configuration (https://github.com/CAREamics/careamics/issues/634).

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Skeleton of new version of the CAREamist API that uses the Next-Gen Dataset and refactored Lightning Modules. Only initialising from a config/config file or a checkpoint has been implemented.

With everything that has changed with the recent refactoring it is easier to rewrite the CAREamics API rather than try and integrate the changes.

### Overview - what changed?

- If loading from a checkpoint or bmz, the `training_config` and `experiment_name` can be overridden.
  - The plan is to pass the checkpoint path to the trainer which will load the optimiser and LR schedular state dicts, this means however that if someone wants to resume training, they will likely have to increase the max epochs.
  - If finetuning one might want to override the experiment name.
- `HyperParametersCallback` replaced by calls to `save_hyperparameters` in the lightning modules and data_modules plus a new `CareamicsCheckpointInfo` callback.
  - Lighning module save the algorithm config, data module saves the data config, and `CareamicsCheckpointInfo` saves the training_config, the experiment_name and the version.
  - Should be more convenient for Lightning API users (they don't need to remember `HyperParametersCallback`.
  - checkpoints saved without `CareamicsCheckpointInfo` can still be loaded by the CAREamist API if a `training_config` and `experiment_name` are provided at initialisation.

Copied most of the function signatures from original CAREamist API.

Decided to use overloads for different initialisation paths, we could also discuss `@classmethod` constructors.

### New features or files

- CAREamistV2
- `CareamicsCheckpointInfo`
- helper functions for lightning module creation in `src/careamics/lightning/dataset_ng/lightning_modules/get_module.py`

## How has this been tested?

No tests yet

## Related Issues

- Resolves #649 

---

- [ ] Pre-commit passes


src/careamics/careamist_v2.py
```diff
@@ -0,0 +1,447 @@
...
+
+logger = get_logger(__name__)
+
+ExperimentLogger: TypeAlias = TensorBoardLogger | WandbLogger | CSVLogger
+
+# union of configurations
+Configuration = N2VConfiguration
+
+
+class CareanicsInfo(TypedDict):
```
CatEek: typo :)

src/careamics/careamist_v2.py
```diff
@@ -0,0 +1,447 @@
...
+        extension_filter: str = "",
+        # override max_epochs or max_steps in training config
+        # TODO: consider allowing override of all trainer args?
+        max_epochs: int | None,
+        max_steps: int | None,
+    ) -> None: ...
+
+    # TODO: init datamodule
+    # TODO: remember to pass self.checkpoint_path to Trainer.fit
+    # ^ this will load optimizer and lr_schedular state dicts
```
CatEek, line 349: typo :)

src/careamics/careamist_v2.py
```diff
@@ -0,0 +1,447 @@
...
+                experiment_loggers = [csv_logger]
+        return experiment_loggers
+
+    def train(
+        self,
+        *,
+        # data init options
+        train_data: Any | None = None,
+        train_data_target: Any | None = None,
+        train_data_mask: Any | None = None,
```
CatEek: I'd argue that this parameter is quite confusing, unless it's an advanced parameter that will not to shown to most users

src/careamics/careamist_v2.py
```diff
@@ -0,0 +1,447 @@
...
+        val_data: Any | None = None,
+        val_data_target: Any | None = None,
+        val_percentage: float | None = None,
+        val_minimum_split: int = 5,
+        # custom loading options
+        read_source_func: Callable | None = None,
+        read_kwargs: dict[str, Any] | None = None,
+        image_stack_loader: ImageStackLoader | None = None,
```
CatEek: I'd argue that for a high level API this shouldn't be user facing. I'm not 100% sure if ImageStackLoader is wrappable into Lightning Datamodule, so we can discuss this. Motivation is that by exposing something specific to CAREamics will come with additional overhead for users

src/careamics/careamist_v2.py
```diff
@@ -0,0 +1,447 @@
...
+        # instantiate trainer
+        self.trainer = Trainer(
+            callbacks=[self.prediction_writer, *self.callbacks],
+            default_root_dir=self.work_dir,
+            logger=experiment_loggers,
+            **self.config.training_config.lightning_trainer_config or {},
+        )
+
+    @staticmethod
+    def _from_config(
```
CatEek, line 160: We should prob discuss using 3 separate methods, CAREamist.from_config, from_checkpoint etc

jdeschamps:
Just pointing out also that there is an `import` error in the tests!

melisande-c:
> Just pointing out also that there is an `import` error in the tests!

Looks like this might be due to rebase artefact after inheritance of UNetModule was removed from lightning module, I will have a look at that

melisande-c:
There is a new error with the tests, I will have a look tomorrow morning

jdeschamps (approved):

PR #724, opened 2026-02-03
Chore: remove unused `extract_patch` method in `ImageStack` implementations

### LINKED ISSUE #708: Next-Gen Dataset: Remove unused method in `ImageStack` and implementations
## Description

Since https://github.com/CAREamics/careamics/pull/640 `ImageStack.extract_patch` is not used anymore and the dataset instead calls `ImageStack.extract_channel_patch`.

We should potentially kick out `extract_patch` from the protocol and the implementations, and maybe even rename `extract_channel_patch` to `extract_patch`.

https://github.com/CAREamics/careamics/blob/ebba001d8496f4cedece7ce3a8bc60560e206e93/src/careamics/dataset_ng/image_stack/image_stack_protocol.py#L37

jdeschamps:
## Description

Since https://github.com/CAREamics/careamics/pull/640 `ImageStack.extract_patch` is not used anymore and the dataset instead calls `ImageStack.extract_channel_patch`. This PR removes the methods and their calls in tests.

Closes https://github.com/CAREamics/careamics/issues/708


melisande-c (approved):
Should `extract_channel_patch` be renamed to `extract_patch`? There is no need to distinguish a patch from a channel patch and it is a little less verbose

jdeschamps:
> Should `extract_channel_patch` be renamed to `extract_patch`? There is no need to distinguish a patch from a channel patch and it is a little less verbose

That's a good idea!

PR #735, opened 2026-02-04
Updated WriteTilesZarr to support tiff and array as the prediction so…

### LINKED ISSUE #648: Next-Gen Dataset: allow predicting from arbitrary source and writing tiles to zarr

mese79:
## Description

Before this, for prediction and writing the result into a `zarr` file, only a `zarr` input source was allowed.  
This is an update to support `tiff` file and `array` input sources while using `WriteTilesZarr` class to save the results in a `zarr` file.

### Overview - what changed?
I updated the `write_tile` method in the `WriteTilesZarr` class.  

I checked the `region.source` for type of input, and based on that I set the `zarr` file path.

- `write_tiles_zarr_strategy.py`
- `test_write_tile_zarr_strategy.py`

## How has this been tested?
I only used the tests in `test_write_tile_zarr_strategy.py`

## Related Issues
- Resolves #648


src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -279,9 +280,25 @@ def write_tile(self, dirpath: Path, region: ImageRegionData) -> None:
         region : ImageRegionData
             Image region data containing tile information.
         """
-        if is_valid_uri(region.source):
+        if region.source == "array":
+            # data source is an in-memory array:
+            # set a new zarr storage output path
+            parent_path = ""
+            output_store_path = dirpath.joinpath("prediction.zarr")
+            array_name = "prediction"
```
jdeschamps: It may be that there will be multiple arrays, so these need to have different names. For instance you could check their `data_idx` and use this in the array name.

This is a similar issue we have faced with `CachedTiles` and `WriteImage` strategies.

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -279,9 +280,25 @@ def write_tile(self, dirpath: Path, region: ImageRegionData) -> None:
...
+
+        elif Path(region.source).suffix in [".tiff", ".tif"]:
```
jdeschamps: Technically, it could also be any suffix (e.g. `custom` data type), since we are getting the `np.array` here, it does not really matter. You could just take the `source.stem`.

The only thing is that you would need to obviously explicitly test for `.zarr`.

tests/lightning/dataset_ng/callbacks/prediction_writer_callbacks/test_write_tile_zarr_strategy.py
```diff
@@ -339,9 +339,8 @@ def test_write_from_array(tmp_path):
 
     # use writer to write predictionsz
     writer = WriteTilesZarr()
-    with pytest.raises(NotImplementedError):
-        for region in gen_image_regions("YX", patch_extractor, strategy):
-            writer.write_tile(tmp_path, region)
+    for region in gen_image_regions("YX", patch_extractor, strategy):
+        writer.write_tile(tmp_path, region)
```
jdeschamps, line 342: Can you add a file loading step and check whether the arrays are `np.testing.assert_allclose`?

tests/lightning/dataset_ng/callbacks/prediction_writer_callbacks/test_write_tile_zarr_strategy.py
```diff
@@ -367,6 +366,5 @@ def test_write_from_tiff(tmp_path):
 
     # use writer to write predictions
     writer = WriteTilesZarr()
-    with pytest.raises(NotImplementedError):
-        for region in gen_image_regions("YX", patch_extractor, strategy):
-            writer.write_tile(tmp_path, region)
+    for region in gen_image_regions("YX", patch_extractor, strategy):
+        writer.write_tile(tmp_path, region)
```
jdeschamps, line 377: Same here, can you add a file loading step and check whether the arrays are `np.testing.assert_allclose`?

tests/lightning/dataset_ng/callbacks/prediction_writer_callbacks/test_write_tile_zarr_strategy.py
jdeschamps: We should probably also check behaviour for multiple files (multiple `data_idx`).

jdeschamps (changes requested):

src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py
```diff
@@ -279,13 +280,35 @@ def write_tile(self, dirpath: Path, region: ImageRegionData) -> None:
...
+            data_idx = region.region_spec["data_idx"]
+            array_name = f"{data_idx}"
+
+        elif is_valid_uri(region.source):
+            # source is a zarr
             store_path, parent_path, array_name = decipher_zarr_uri(region.source)
             output_store_path = _add_output_key(dirpath, store_path)
+
+        elif ".zarr" not in region.source:
+            # data source is a tiff image:
```
jdeschamps:
```suggestion
            # data source is a tiff or custom format image:
```

jdeschamps (approved):

PR #743, opened 2026-02-04
Feat: split convenience function into standard and advanced.

### LINKED ISSUE #721: Next-Gen CAREamics: standard vs advanced configuration
## Problem

The convenience functions parameters are exploding, maybe we should have two flavours `create_n2v_configuration` and `create_advanced_n2v_config`.

jdeschamps:
## Description

The changes introduced by this PR are the following:

- Reduce the number of parameters in standard convenience function
- Introduce an advanced convenience function with full parameter range
- Add `n_workers` parameter to advanced convenience function
- Changed `augmentations` to `Literal` to not force users to create Pydantic models

Closes https://github.com/CAREamics/careamics/issues/721

### DIFF
--- a/src/careamics/config/ng_factories/n2v_factory.py
+++ b/src/careamics/config/ng_factories/n2v_factory.py
@@ -32,7 +32,6 @@ def create_n2v_config(
     augmentations: Sequence[Literal["x_flip", "y_flip", "rotate_90"]] | None = None,
     use_n2v2: bool = False,
     n_channels: int | None = None,
-    n_workers: int = 0,
 ) -> N2VConfiguration:
     """
     Create a configuration for training N2V.
@@ -61,8 +60,6 @@ def create_n2v_config(
         Whether to use N2V2.
     n_channels : int or None, default=None
         Number of channels (in and out).
-    n_workers : int, default=0
-        Number of workers for data loading.
 
     Returns
     -------
@@ -80,7 +77,6 @@ def create_n2v_config(
         augmentations=augmentations,
         use_n2v2=use_n2v2,
         n_channels=n_channels,
-        n_workers=n_workers,
     )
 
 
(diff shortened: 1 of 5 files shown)


src/careamics/config/ng_factories/n2v_factory.py
```diff
@@ -19,26 +19,198 @@
...
+    Create a configuration for training N2V.
+
+    To activate N2V2, set `use_n2v2` to True.
+
+    The `axes` parameters must reflect the actual axes and axis order from the data,
+    and should be the same throughout all images. The accepted axes are STCZYX. If "C"
+    is in `axes`, then you need to set `n_channels` to the number of channels.
+
+    `patch_size` is only along the spatial dimensions and should be of length 3 if "Z"
+    is present in `axes`, otherwise of length 2.
```
melisande-c: I guess these docs are just copied from the original but this is basically repeating the information in the parameters sections

tests/config/ng_factories/test_n2v_factory.py
```diff
@@ -143,216 +137,10 @@ def test_num_epochs_and_num_steps(self):
...
-        )
-
-        # Explicit parameters should override trainer_params
-        assert (
-            config.training_config.lightning_trainer_config["max_epochs"] == num_epochs
-        )
-        assert (
-            config.training_config.lightning_trainer_config["limit_train_batches"]
-            == num_steps
-        )
```
melisande-c, line 224: maybe it's still worth testing this override behaviour, but equally don't think it's that important

melisande-c (approved):
Looks good, 

made a comment on the docstrings but I guess we can review all the doctrings another time

src/careamics/config/ng_factories/n2v_factory.py
jdeschamps: No I actually rewrote that... ^^ 

I removed it, but we should indeed review the docs.

tests/config/ng_factories/test_n2v_factory.py
jdeschamps: Yes that was a mistake, corrected!

Issue #520, opened 2025-08-12
Lightning API load checkpoint convenience functions


CatEek:
## Problem

In LIghtning API checkpoint doesn;t contain neither hyperparameter nor config, that are necessary to load a checkpoint. That means a user has to recreate a config and load the state dict manually. 

## Potential solution

make it similar to Careamist API or provide a convenience function

melisande-c:
Resolved for Next-Gen CAREamics in #717 

the lightning modules now call `self.save_hyperparameters` which makes it possible to do e.g. `N2VModule.load_from_checkpoint(ckpt_path)` without the `HyperParametersCallback` being added to the trainer.

PR #744, opened 2026-02-04
CareamistV2: initialization cleanup

veegalinova:
## Description

> [!NOTE]  
> **tldr**: clean up `CAREamistV2.__init__` implementation

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist_v2.py
+++ b/src/careamics/careamist_v2.py
@@ -1,15 +1,6 @@
 from collections.abc import Callable
 from pathlib import Path
-from typing import (
-    Any,
-    Literal,
-    Never,
-    NotRequired,
-    TypeAlias,
-    TypedDict,
-    Unpack,
-    overload,
-)
+from typing import Any, Literal, TypedDict, Unpack
 
 import torch
 from numpy.typing import NDArray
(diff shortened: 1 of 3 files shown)


src/careamics/careamist_v2.py
```diff
@@ -90,79 +44,61 @@ def __init__(
...
-        self.config: Configuration
-        self.model: CAREamicsModule
-        self.data_module: CareamicsDataModule
-        self.trainer: Trainer
-        self.callbacks: list[Callback]
-        # checkpoint path is saved to restore optimizer etc. state_dicts during training
-        # only populated if loading from checkpoint.
         self.checkpoint_path = checkpoint_path
-        self.work_dir = self._resolve_work_dir(user_context["work_dir"])
-        # ---
```
melisande-c, line 107: I quite like pre-declaring attributes at the start of `__init__` if the `__init__` is long bec it makes it easier to see what is in the class

src/careamics/careamist_v2.py
```diff
@@ -90,79 +44,61 @@ def __init__(
...
+        self.work_dir = self._resolve_work_dir(user_context.get("work_dir"))
+        self.config, self.model = self._load_model(config, checkpoint_path, bmz_path)
 
-        # override progress bar choice
+        enable_progress_bar = user_context.get("enable_progress_bar", True)
         self.config.training_config.lightning_trainer_config["enable_progress_bar"] = (
-            user_context.get("enable_progress_bar", True)
+            enable_progress_bar
         )
```
melisande-c: if `enable_progress_bar` is not taken from `user_context` anymore, it should be removed from `UserContext`

src/careamics/careamist_v2.py
```diff
@@ -171,47 +107,33 @@ def _from_config(
...
         careamics_info = checkpoint.get("careamics_info", None)
         if careamics_info is None:
             raise ValueError(
                 "Could not find CAREamics related information within the provided "
                 "checkpoint. This means that it was saved without using the "
-                "CAREamics callback `CAREamicsCheckpointInfo`. To load this "
-                "checkpint with the CAREamist API please provide an "
-                "`experiment_name` and `training_config`."
+                "CAREamics callback `CareamicsCheckpointInfo`. "
+                "Please use a checkpoint saved with CAREamics or initialize with a config instead."
```
melisande-c, line 122: Now that Joran merged #725 there is a default `TrainingConfig` so the configuration doesn't need it to be initialised. So now all that is missing is the experiment name.

As we were discussing before, if someone is trying to load a checkpoint for prediction but it doesn't have an experiment name, this is probably fine? Not sure what a good solution is though

Sorry if I didn't write this very clearly

melisande-c (commented):
Looks good, just a couple things.

Main thing is when loading from a checkpoint that was saved via the Lightning API but without `CareamicsCheckpointInfo` callback, so that there will be no experiment name saved, maybe we want to find a way that it can still be loaded by the CAREamist for predicting

src/careamics/careamist_v2.py
melisande-c: but up to you if you think it's bloating the function

veegalinova: I think having clear functions in the init with no extra logic visible is similar in spirit to having an explicit list of attributes. But it is something we can for sure keep striving for 😄

veegalinova:
@melisande-c @jdeschamps So following Milly's comment about checkpoint loading, I tried loading a checkpoint made with the current Careamist and ran into a few things I want to discuss.
The main issue is that the old `HyperParametersCallback` puts everything into `hyper_parameters`, but the new `CareamicsCheckpointInfo` callback writes to a separate `careamics_info` key. Should we just stick with `hyper_parameters` for consistency, or add extra logic to cover this?
And related to this, `data_config` also ends up in different places. The old format has it in `hyper_parameters`, now we also use `save_hyperparameters()` in DataModules that creates `datamodule_hyper_parameters`, which will be empty for old checkpoints. 
And lastly, if the user trains with lightning api following current docs, we do not add the `HyperParameters` callback at all, so all of the keys will be empty. Do we want to cover this also?

melisande-c:
> And related to this, `data_config` also ends up in different places. The old format has it in `hyper_parameters`, now we also use `save_hyperparameters()` in DataModules that creates `datamodule_hyper_parameters`, which will be empty for old checkpoints.

So it is a bit unfortunate that all the config gets split up in the checkpoint this way, but I was trying to follow more closely how lightning expects a checkpoint to be organised. I did consider duplicating the entire config in a different checkpoint key, but ultimately thought it was unnecessary if we know where to extract the different parts of the config.

> And lastly, if the user trains with lightning api following current docs, we do not add the HyperParameters callback at all, so all of the keys will be empty. Do we want to cover this also?

This is partly why I want to go with calling `save_hyperparameters` in the Lightning modules, it means for the lightning API we don't have to tell users to add the `HyperParametersCallback`, without which they would have to completely recreate the config if they want to load the checkpoint, see #520.

It does mean CAREamistV2 is not compatible with old checkpoints... but we have made changes in the past which make checkpoints incompatible, namely changing the model architecture. We should definitely discuss this further though

src/careamics/careamist_v2.py
melisande-c: Ok I was only looking at the diff file so the __init__ still looked long to me, but you're right, it is concise enough to not need the declaration of the attributes

veegalinova:
>  We should definitely discuss this further though

Opened an issue for this - #753, feel free to add thoughts there! 
Otherwise, should this PR be merged?

melisande-c (approved):

melisande-c:
> Otherwise, should this PR be merged? 

Go ahead :)

PR #754, opened 2026-02-06
Feat(Next-Gen CAREamist): prediction methods

### LINKED ISSUE #323: Add prediction writer callback
The idea of creating a lightning [prediction writer callback](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.BasePredictionWriter.html) is to unify writing to TIFF and writing to ZARR.

The strategy for writing a to the different files can be decoupled from the prediction writer itself by creating write strategy interface, with concrete classes that can be an attribute of the prediction writer.

TODO:
- [x] `PredictionWriterCallback`
- [x] write strategies
  - [x] `WriteImage`
  - [x] `WriteTiles`
  - [x] `WriteZarrTiles`
- [ ] ~~File metadata extractors (mostly to extract the number of samples (i.e. tiff pages) per file)~~
- [ ] Reimplement `CAREamist.predict_to_disk` function

### LINKED ISSUE #729: Next-Gen CAREamics: add prediction methods body to `CAREamistv2`
## Problem

The new `CAREamist` has two methods, `predict` and `predict_to_disk`. The first one returns the predictions, while the second one should make use of the `PredictionWriter` callback.

## Creating the data config

Prediction data config is created by calling `self.config.data_config.convert_mode(mode="predicting", ...)`, with new parameters needing to be updated, and needs to be passed the data module.

## Predict to disk

The prediction writer needs to be enabled (see call to disable it in `__init__` method), and the parameters (data_type and tiled) need to be passed `set_writing_strategy` of the callback.

Additional note: tiling is enabled by passing a `tile_size`.

CatEek:
## Description

> [!NOTE]  
> **tldr**: 

Prediction methods for next gen CAREamist

## How has this been tested?

No performance testing yet

- Resolves #

#729 

- [x] Pre-commit passes


src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
+        *,
         batch_size: int = 1,
         tile_size: tuple[int, ...] | None = None,
         tile_overlap: tuple[int, ...] | None = (48, 48),
         axes: str | None = None,
-        data_type: Literal["array", "tiff", "custom"] | None = None,
+        data_type: Literal["array", "tiff", "zarr", "custom"] | None = None,
         # ADVANCED PARAMS
-        # tta_transforms: bool = False, # TODO: hidden till implemented
-        num_workers: int | None = None,
```
jdeschamps, line 296: `n_workers` should probably be added back as it should be a parameter that we can update in `data_config.convert_mode`:

https://github.com/CAREamics/careamics/blob/dc493b30cad182422432a2a23abc69cf6d990508/src/careamics/config/data/ng_data_config.py#L721

src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
         read_source_func: Callable | None = None,
         read_kwargs: dict[str, Any] | None = None,
         extension_filter: str = "",
-    ) -> None:
-        return None
+    ) -> list[NDArray]:
+        """
+        Make predictions on the provided data.
```
jdeschamps:
```suggestion
        Predict on data and returns the prediction.
```

src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
+
+        Input can be a CareamicsDataModule instance, a path to a data file, or a numpy
+        array.
+
+        If `data_type`, `axes` and `tile_size` are not provided, the training
+        configuration parameters will be used, with the `patch_size` instead of
+        `tile_size`.
```
jdeschamps: There is currently a bit of an odd behaviour in `data_config.convert_mode`. If a `new_patch_size` is passed, then it will set the patching to `tiling`.

~~That means that this implementation will always tile, which is not the intended behaviour.~~

src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
+
+        Note that if you are using a UNet model and tiling, the tile size must be
+        divisible in every dimension by 2**d, where d is the depth of the model. This
+        avoids artefacts arising from the broken shift invariance induced by the
+        pooling layers of the UNet. If your image has less dimensions, as it may
+        happen in the Z dimension, consider padding your image.
```
jdeschamps: We need to check, but I am pretty sure we are now automatically padding in `Z` (but zero padding).

src/careamics/careamist_v2.py
jdeschamps: Actually, the behaviour is correct since it is passing `tile_size: tuple[int, ...] | None` to `data_config.convert_mode`. It is the docstring that is wrong about tile size.

src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
+        # If datamodule is provided directly, use it
+        if isinstance(pred_data, CareamicsDataModule):
+            datamodule = pred_data
+        else:
+            # Create prediction data config using convert_mode
+            pred_data_config = self.config.data_config.convert_mode(
+                new_mode="predicting",
+                new_patch_size=tile_size,
+                overlap_size=tile_overlap,
+                new_batch_size=batch_size,
```
jdeschamps: Here it is wrong, as it will always lead to `batch_size=1`. `batch_size` should probably default to `None`.

src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
+        ------
+        ValueError
+            If tile overlap is not specified when tile_size is provided.
+        """
+        # If datamodule is provided directly, use it
+        if isinstance(pred_data, CareamicsDataModule):
+            datamodule = pred_data
+        else:
+            # Create prediction data config using convert_mode
+            pred_data_config = self.config.data_config.convert_mode(
```
jdeschamps: Missing the following parameters:
```
        new_channels: Sequence[int] | Literal["all"] | None = None,
        new_in_memory: bool | None = None,
        new_dataloader_params: dict[str, Any] | None = None,
```

- `channels` and `in_memory` should probably be in the predict function API.
- `new_dataloader_params` should probably include only `num_workers` if that one has been specified

src/careamics/careamist_v2.py
```diff
@@ -392,7 +428,154 @@ def predict_to_disk(
...
         write_func: WriteFunc | None = None,
         write_func_kwargs: dict[str, Any] | None = None,
-    ) -> None: ...
+    ) -> None:
+        """
+        Make predictions on the provided data and save outputs to files.
+
+        The predictions will be saved in a new directory 'predictions' within the set
+        working directory. The directory stucture within the 'predictions' directory
+        will match that of the source directory.
```
jdeschamps: Users should be able to set an independent destination directory that may not be in the working directory (e.g. can be absolute).

src/careamics/careamist_v2.py
```diff
@@ -353,23 +281,131 @@ def train(
...
+        # Predict
+        predictions = self.trainer.predict(model=self.model, datamodule=datamodule)
+
+        # Convert predictions using convert_prediction
+        tiled = tile_size is not None
+        predictions_output, sources = convert_prediction(predictions, tiled=tiled)
+
+        return predictions_output, sources
 
     def predict_to_disk(
```
jdeschamps, line 434: Similarly as the other method, missing some parameters (channels, in memory)

src/careamics/careamist_v2.py
```diff
@@ -392,7 +428,154 @@ def predict_to_disk(
...
+        The predictions will be saved in a new directory 'predictions' within the set
+        working directory. The directory stucture within the 'predictions' directory
+        will match that of the source directory.
+
+        The `pred_data` must be from files and not arrays. The file names of the
+        predictions will match those of the source. If there is more than one sample
+        within a file, the samples will be saved to seperate files. The file names of
+        samples will have the name of the corresponding source file but with the sample
+        index appended. E.g. If the the source file name is 'images.tiff' then the first
+        sample's prediction will be saved with the file name "image_0.tiff".
```
jdeschamps: That's not true, the sample are stacked in the prediction writers, aren't they?

Also, I don't really see why not to prevent predicting and writing from arrays. The prediction writer is compatible with that.

src/careamics/careamist_v2.py
```diff
@@ -392,7 +428,154 @@ def predict_to_disk(
...
+
+        If `data_type`, `axes` and `tile_size` are not provided, the training
+        configuration parameters will be used, with the `patch_size` instead of
+        `tile_size`.
```
jdeschamps: Not `tile_size`

src/careamics/careamist_v2.py
jdeschamps: Reading the code below I think this is indeed the behaviour, so the docstring is wrong.

src/careamics/careamist_v2.py
```diff
@@ -392,7 +428,154 @@ def predict_to_disk(
...
+        self.prediction_writer.set_writing_strategy(
+            write_type=write_type,
+            tiled=tiled,
+            write_func=write_func,
+            write_extension=write_extension,
+            write_func_kwargs=write_func_kwargs,
+        )
+
+        # Enable writing
+        self.prediction_writer.writing_predictions = True
```
jdeschamps: This flag has changed name now.

src/careamics/careamist_v2.py
```diff
@@ -392,7 +428,154 @@ def predict_to_disk(
...
+                    data_config=pred_data_config,
+                    pred_data=pred_data,
+                    pred_data_target=pred_data_target,
+                    read_source_func=read_source_func,
+                    read_kwargs=read_kwargs,
+                    extension_filter=extension_filter,
+                )
+
+            # Predict (writing will be handled by the callback)
+            self.trainer.predict(model=self.model, datamodule=datamodule)
```
jdeschamps: Should you specifically set `return_predictions=False`?

src/careamics/careamist_v2.py
```diff
@@ -392,7 +428,154 @@ def predict_to_disk(
...
+
+        finally:
+            # Disable writing after prediction is complete
+            self.prediction_writer.writing_predictions = False
```
jdeschamps: Also here, flag will have change name

tests/conftest.py
```diff
@@ -354,6 +354,40 @@ def pre_trained(tmp_path, minimum_n2v_configuration):
...
+    from careamics.careamist_v2 import CAREamistV2
+    from careamics.config.ng_factories import create_n2v_configuration
+
+    # training data
+    train_array = np.arange(32 * 32).reshape((32, 32)).astype(np.float32)
+
+    # create configuration
+    patch_size = [8, 8]
+    masked_pixel_percentage = 100 / np.prod(patch_size)
+    config = create_n2v_configuration(
```
jdeschamps:
```suggestion
    config = create_n2v_config(
```

jdeschamps (changes requested):
- There are current merge conflicts with the changes in the CAREamics v2 API (#744)
- The tests fail because `create_n2v_configuration` was renamed to `create_n2v_config` (#743)

src/careamics/careamist_v2.py
CatEek: I dont understand

src/careamics/careamist_v2.py
CatEek: Why 1? It's just the default

jdeschamps: The docs says that training batch size will be used if `batch_size` is not set. But it will use a default that is `1`. 

So there is an inconsistency.

src/careamics/careamist_v2.py
jdeschamps: Missing one line. Docstring says:

"If `data_type`, `axes` and `tile_size` are not provided, the training configuration parameters will be used, with the `patch_size` instead of `tile_size`."

but that's not the case for `tile_size`.

src/careamics/careamist_v2.py
melisande-c: yes there will be automatic zero padding for prediction, in any spatial dimension, if the tile size is greater than the image size

src/careamics/careamist_v2.py
```diff
@@ -264,23 +274,131 @@ def train(
...
+        tile_overlap: tuple[int, ...] | None = (48, 48),
+        axes: str | None = None,
+        data_type: Literal["array"] | None = None,
+    ) -> list[NDArray]: ...
+
     def predict(
         self,
         # BASIC PARAMS
-        pred_data: Any | None = None,
+        pred_data: CareamicsDataModule | Path | str | NDArray,
```
melisande-c: `pred_data` argument doesn't match `source` argument in all the overloads. Also we should accept a list of paths and arrays.

We also had a discussion about whether the CAREamist API should accept pre-initialised data modules for train/predict but I don't remember the conclusion, if there was one, @jdeschamps ?

jdeschamps: It was not entirely clear. I think we ended up on:

0) We agreed it is a largely unlikely user case (needing a specific ImageStack implementation)
1) More people preferred data module over passing `ImageStacks` directly
2) We agreed that it could also be ignored for now and added if requests come in (other users can turn to the Lightning API)

CatEek:
A general comment on docstings are too complicated at some parts. But I think we need to chaeck this in the end once code is ready #757

src/careamics/careamist_v2.py
veegalinova: I ended up removing all overloads and removed datamodule support for now, in line with the train method

src/careamics/config/data/ng_data_config.py
```diff
@@ -707,7 +708,7 @@ def convert_mode(
         new_patch_size: Sequence[int] | None = None,
         overlap_size: Sequence[int] | None = None,
         new_batch_size: int | None = None,
-        new_data_type: Literal["array", "tiff", "zarr", "czi", "custom"] | None = None,
+        new_data_type: SupportedData | None = None,
```
jdeschamps: Using `SupportedData` makes sense, but I'd prefer to keep the types explicit here since this is a method that people will call directly. `SupportedData` was meant to be used internally.

I concede that it brings extra maintenance work...

src/careamics/careamist_v2.py
```diff
@@ -317,40 +323,167 @@ def train(
...
-        pred_data: Any | None = None,
-        batch_size: int = 1,
+        pred_data: InputType,
+        *,
+        batch_size: int | None = None,
         tile_size: tuple[int, ...] | None = None,
         tile_overlap: tuple[int, ...] | None = (48, 48),
         axes: str | None = None,
-        data_type: Literal["array", "tiff", "custom"] | None = None,
+        data_type: SupportedData | None = None,
```
jdeschamps: Same comment here about making the allowed values explicit.

src/careamics/careamist_v2.py
```diff
@@ -359,7 +492,138 @@ def predict_to_disk(
...
+        relative paths are relative to `work_dir`). The directory structure matches
+        the source directory.
+
+        The file names of the predictions will match those of the source. If there is
+        more than one sample within a file, the samples will be stacked along the sample
+        dimension in the output file.
+
+        If `data_type` and `axes` are not provided, the training configuration
+        parameters will be used. If `tile_size` is not provided, the whole image
+        strategy will be used for prediction.
```
jdeschamps:
```suggestion
        parameters will be used. If `tile_size` is not provided, prediction
        will be performed on whole images rather than in a tiled manner.
```

src/careamics/careamist_v2.py
```diff
@@ -317,40 +323,167 @@ def train(
...
+            read_source_func=read_source_func,
+            read_kwargs=read_kwargs,
+            extension_filter=extension_filter,
+        )
+
+        predictions: list[ImageRegionData] = self.trainer.predict(
+            model=self.model, datamodule=datamodule
+        )  # type: ignore[assignment]
+        tiled = tile_size is not None
+        predictions_output, sources = convert_prediction(predictions, tiled=tiled)
```
jdeschamps: `convert_prediction` has now a parameter to restore to the original shape, we should put it to default `True` here.

tests/test_careamist_v2_predict.py
```diff
@@ -0,0 +1,272 @@
...
+    careamist = CAREamistV2(config=config, work_dir=tmp_path)
+    careamist.train(train_data=train_array, val_data=val_array)
+
+    predicted, _ = careamist.predict(
+        train_array, batch_size=batch_size, tile_size=(16, 16), tile_overlap=(4, 4)
+    )
+
+    # TODO revisit after fixing prediction reshaping to orig shape
+    assert predicted[0].shape[0] == samples
+    assert predicted[0].shape[-2:] == (32, 32)
```
jdeschamps: This is now possible since the PR has been merged.

tests/test_careamist_v2_predict.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+
+    for p in predicted:
+        assert p.squeeze().shape == train_array.shape
+
+
+@pytest.mark.mps_gh_fail
+@pytest.mark.parametrize("independent_channels", [False, True])
+@pytest.mark.parametrize("batch_size", [1, 2])
+def test_predict_tiled_channel(
```
jdeschamps, line 121: A more interesting test would be to check end-to-end that we can predict on a subset of channels using `channels`.

jdeschamps (changes requested):
Let's add the `restore_shape=True` in `convert_prediction` to restore the correct shape. (note it will have no effect on prediction writer for Zarr).

tests/test_careamist_v2_predict.py
veegalinova: That was a good idea. I added this test, and an issue surfaced: to predict on a subset of channels, we need to update the stored statistics to include only those channels in `convert_mode`. 
For now, this test is marked with skip. I will create an issue for this, since it's not straightforward to me and is out of scope for this PR.

jdeschamps: If a dataset has `N` channels but we train only on `channels=[0, .., m]` (with `m<N`), are the statistics computed over all `N` channels rather than the `m` channels?

jdeschamps:
Closes https://github.com/CAREamics/careamics/issues/729

tests/test_careamist_v2_predict.py
veegalinova: Yes. So I think we should check that the prediction channels are the subset of the training channels?

jdeschamps: It seems to me that the statistics are computed on the patch extracted with `channels`:

https://github.com/CAREamics/careamics/blob/9730fce6c0bd0efc44c2718194a36d83f3e89561/src/careamics/dataset_ng/normalization/statistics.py#L101-L105

So there should be only `m` channel statistics, where `m` is either `len(channels)` or `shape[1]`.

Therefore I see the following scenarii:
    - if `channels=None`, prediction should raise an error if `data_config.channels` cannot index the data shape (not enough channels given the indices in `data_config.channels`)
    - if `channels="all", prediction should raise an error if any of the predict_data array has a different number of channels from `data_config.channels`
    - if `len(channels)!=len(data_config.channels)`, raise an error

and of course there is the case where `per_channel=False`.

Am I missing something?

jdeschamps (approved):

PR #759, opened 2026-02-07
Feat: Add seed to N2V pixel manipulation

### LINKED ISSUE #739: Next-Gen CAREamics: seed N2V pixel manipulation
## Problem

Currently, it does not seem that the manipulate N2V transform is seeded, which would impede full reproducibility.

CatEek:
## Description

> [!NOTE]  
> **tldr**: Added seed parameter to N2V manipulation transform configuration to enable reproducible pixel masking during training.

The N2V manipulation transform was not seeded, preventing reproducible training runs. Pixel masks generated during training were random and could not be reproduced even with the same configuration, which is important for debugging, comparing experiments, and ensuring reproducibility.

### Overview - what changed?

Added a `seed` field to `N2VManipulateConfig` and propagated it through configuration factories to `N2VManipulateTorch`. This follows the pattern used by other transforms like `XYFlipConfig` and `XYRandomRotate90Config`. When a seed is provided, the transform produces identical masks across runs; when `None`, it maintains random behavior.

1. Added `seed: int | None = None` to `N2VManipulateConfig` following existing transform patterns
2. Updated `create_n2v_configuration()` and `create_pn2v_configuration()` to accept a `seed` parameter and pass it to `N2VManipulateConfig`
3. Modified `N2VManipulateTorch` instantiation in both `FCNModule` and `N2VModule` to pass the seed from config
4. Updated `create_advanced_n2v_config()` in ng_factories to use the seed parameter
5. The seed is used to initialize the PyTorch `Generator` in `N2VManipulateTorch`, ensuring reproducible mask generation

- `N2VManipulateConfig` - Added `seed` field for reproducibility
- `create_n2v_configuration()` - Added `seed` parameter and propagation to config
- `create_pn2v_configuration()` - Added `seed` parameter and propagation to config
- `create_advanced_n2v_config()` - Updated to pass seed to `N2VManipulateConfig`
- `FCNModule` - Updated to pass seed from config to `N2VManipulateTorch`
- `N2VModule` - Updated to pass seed from config to `N2VManipulateTorch`
- `test_manipulate_n2v.py` - Added `test_manipulate_n2v_torch_with_seed()` to verify reproducibility
- `test_careamist.py` - Added `test_train_array_mask_changes_per_epoch()` to verify masks change across epochs when seed is None

## How has this been tested?

- Unit test `test_manipulate_n2v_torch_with_seed()` verifies:
  - Same seed produces identical masks across multiple transform instances
  - No seed (None) produces different masks
- Integration test `test_train_array_mask_changes_per_epoch()` verifies:
  - Masks change across training epochs when seed is None
  - Uses a callback to capture masks from the first batch of each epoch

## Related Issues

- Resolves #739 

## Breaking changes

None. The `seed` parameter defaults to `None`, maintaining backward compatibility. Existing code continues to work without changes, and randomness is preserved when no seed is provided.

The implementation follows the existing pattern established by other transforms in the codebase. When a seed is set, full reproducibility is enabled; when `None`, the transform maintains its original random behavior. This allows users to choose between reproducibility and randomness based on their needs.

Example usage:
```python
    experiment_name="n2v_experiment",
    data_type="array",
    axes="YX",
    patch_size=[64, 64],
    batch_size=32,
    seed=42  # Enable reproducibility
)
```

---

- [x] Pre-commit passes


veegalinova:
I made a few changes to the original PR:
1. Now all code paths should fix a random seed. If a seed is not specified, it is generated randomly and stored in the configuration, so any run can be repeated. 
2. There are a lot of small changes in imports caused by a circular import in current config loading utils.

I do not foresee any issues with n2v masking after this PR, masking class is initialized once in the lightning module and guaranteed to produce different (seeded) values with every call. But I will now test performance just in case, will update this comment.

jdeschamps (commented):
The CI is failing because doctests in `src/careamics/config/configuration.py` are not passing.

veegalinova:
> The CI is failing because doctests in `src/careamics/config/configuration.py` are not passing.

Not only this, also every n-th run the mask consistency between runs with the same seed is not respected, looking into it

veegalinova:
> Not only this, also every n-th run the mask consistency between runs with the same seed is not respected, looking into it

So this is only true for the old dataset implementation, since the seeds are not propagated everywhere. 

I made a couple of runs with the new dataset, loss reproduced, and the final metric is okay 
`Scale invariant PSNR: 26.28 +/- 2.59`
Screenshot 2026-02-23 at 10 25 40

Made a mental note to add a mask reproducibility test to the new careamist's tests once it's ready.

jdeschamps (approved):
One thought that came to mind. The seeds for the filtering are set via Pydantic model validation, while the others are set explicitly in the convenience functions. While the N2V manipulation is defined outside of the NG data configuration, it is a bit inconsistent to have the seed from the data configuration propagate to the filters and not to the augmentations.

PR #761, opened 2026-02-08
Feat(Next-Gen CAREamics): CAREamist v2 training methods

CatEek:
## Description

> [!NOTE]  
> **tldr**: Implemented training methods in `CAREamistV2` using the ng  dataset logic, largely copied from `CAREamist` with updated parameter names and API.

The `CAREamistV2` kinda needs them.

### Overview - what changed?

- Implemented `train()`, `_train_on_array()`, `_train_on_path()`, and `_train_on_datamodule()` methods in `CAREamistV2`
- Updated parameter names to match the new API (`train_data` instead of `train_source`, etc.)

Copied from V1 with minor changes

### New features or files
- `train()`, `_train_on_array()`, `_train_on_path()`, `_train_on_datamodule()` methods in `CAREamistV2`

- `src/careamics/careamist_v2.py`

## How has this been tested?

Added  tests in `tests/test_careamist_v2.py` 
Largely a copy of old tests
All tests use `create_advanced_n2v_config()` to add custom ROI size to avoid 0 masked pixels 

---

- [x] Pre-commit passes
```

### CODE REVIEW

File: src/careamics/careamist_v2.py [hunk 1]
```diff
@@ -258,11 +261,230 @@ def train(
...
+                extension_filter=extension_filter,
+                val_percentage=0.1,
+                val_minimum_split=5,
+            )
+
+            # Train
+            self._train_on_datamodule(datamodule)
+        finally:
+            # Restore original in_memory setting
+            self.config.data_config.in_memory = original_in_memory
```
melisande-c: I don't think `use_in_memory` should be an argument to train anymore now that it is in the config.

Silently changing the value in the config and changing it back is maybe not quite in the spirit of the config being reproducible

src/careamics/careamist_v2.py
melisande-c, line 1: Since we're rewriting the class, we have the opportunity to remove some of the bloat, I don't think we need `_train_on_array`, `_train_on_path`, `_train_on_datamodule`.

`_train_on_array` does nothing and `_train_on_path` only checks that the paths exist but that should arguable happen in this `validate_path_input` function which is called during the initialisation of the `CAREamicsDataModule`.

https://github.com/CAREamics/careamics/blob/dc493b30cad182422432a2a23abc69cf6d990508/src/careamics/lightning/dataset_ng/data_module_utils.py#L184-L189

The train function can simply initialise the datamodule from the arguments and then begin training

src/careamics/careamist_v2.py
```diff
@@ -258,11 +261,230 @@ def train(
         read_source_func: Callable | None = None,
```
melisande-c: Can we update type of `read_source_func` to the `ReadFunc` protocol in the `file_io` module

src/careamics/careamist_v2.py
melisande-c: These were set as type `Any` when we still had the `image_stack_loader` argument but now it should only be arrays, file paths and list of arrays or file paths

src/careamics/careamist_v2.py
```diff
@@ -258,11 +261,230 @@ def train(
...
+                raise ValueError("Training target not compatible with N2V training.")
+
+        # Dispatch the training based on input type
+        if isinstance(train_data, np.ndarray):
+            # mypy checks
+            assert isinstance(val_data, np.ndarray) or val_data is None
+            assert (
+                isinstance(train_data_target, np.ndarray) or train_data_target is None
+            )
+            assert isinstance(val_data_target, np.ndarray) or val_data_target is None
```
melisande-c: These asserts were in the original train function but they won't give a user a good error message if they happen to give different types for `train_data`, `val_data`, `train_data_target` and `val_data_target`.

I think they can be avoided by using a `TypeVar` in the argument type, i.e.:

```python
from typing import TypeVar

InputType = TypeVar(
    "InputType", NDArray[Any], Path, str, Sequence[NDArray[Any]], Sequence[Path | str]
)

class CAREamistV2:
    ...

    def train(
        self,
        *,
        # BASIC PARAMS
        train_data: InputType | None = None,
        train_data_target: InputType | None = None,
        val_data: InputType | None = None,
        val_data_target: InputType | None = None,
        ...
) -> None: ...

```

This ensures that all the arguments have to be the same type or None

tests/test_careamist_v2.py
```diff
@@ -0,0 +1,322 @@
...
+            train_data=tmp_path,
+            train_data_target=tmp_path,
+        )
+
+    # train error with strings
+    with pytest.raises(ValueError):
+        careamics.train(
+            train_data=str(tmp_path),
+            train_data_target=str(tmp_path),
+        )
```
melisande-c: if there is already a test for `careamics.utils.check_path_exists` then I don't think these checks are necessary

src/careamics/careamist_v2.py
CatEek: So we don't allow passing a datamodule ?

melisande-c: it could still accept a datamodule but I just meant that all these extra `_train_on_x` functions seem unnecessary to me and the train func could look something like:

```python
    def train(
        self,
        datamodule: CareamicsDataModule | None = None,
        # BASIC PARAMS
        train_data = None,
        train_data_target = None,
        val_data = None,
        val_data_target = None,
        ...
    ) -> None:
        if datamodule is None:
            datamodule = CareamicsDataModule(
                config=self.config.data_config,
                train_data=train_data,
                train_data_target=train_data_target,
                val_data=val_data,
                ....
            )
       
       # ... rest of train func
```

but we were also discussing whether the CAREamist should accept datamodules in this comment on the predict functions as well https://github.com/CAREamics/careamics/pull/754#discussion_r2774195508

melisande-c (approved):
Looks good :)

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,104 @@ def train(
...
+        val_data_target: InputType | None = None,
         # val_percentage: float | None = None, # TODO: hidden till re-implemented
         # val_minimum_split: int = 5,
         # ADVANCED PARAMS
         filtering_mask: Any | None = None,
-        read_source_func: Callable | None = None,
+        read_source_func: ReadFunc | None = None,
         read_kwargs: dict[str, Any] | None = None,
         extension_filter: str = "",
+        use_in_memory: bool = True,
```
veegalinova:
```suggestion
```
Parameter is not used but still there, otherwise everything looks good!

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,103 @@ def train(
...
+        Parameters
+        ----------
+        train_data : pathlib.Path or str or numpy.ndarray, optional
+            Training data, by default None.
+        train_data_target : pathlib.Path or str or numpy.ndarray, optional
+            Training target data, by default None.
+        val_data : pathlib.Path or str or numpy.ndarray, optional
+            Validation data, by default None.
+        val_data_target : pathlib.Path or str or numpy.ndarray, optional
+            Validation target data, by default None.
```
jdeschamps: The doc does not convey that inputs can also be passed as sequences. What would be the best way to write it? Any idea? 

@CatEek @veegalinova @melisande-c

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,103 @@ def train(
...
     ) -> None:
-        # TODO: init datamodule
-        # TODO: remember to pass self.checkpoint_path to Trainer.fit
-        # ^ this will load optimizer and lr_schedular state dicts
-        raise NotImplementedError("Training is not implemented yet.")
+        """Train the model on the provided data.
+
+        The training data can be provided as arrays or paths. If `use_in_memory`
+        is set to True in the configuration, the source provided as Path or str will
+        be loaded in memory if it fits. Otherwise, training will be performed by loading
```
jdeschamps: `use_in_memory` is gone, so it should be removed from the doc.

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,103 @@ def train(
...
+        filtering_mask : Any, optional
+            Filtering mask for coordinate-based patch filtering, by default None.
+        read_source_func : Callable, optional
+            Function to read the source data, by default None.
+        read_kwargs : dict[str, Any], optional
+            Keyword arguments for the read function, by default None.
+        extension_filter : str, optional
+            Filter for file extensions, by default "".
+        use_in_memory : bool, optional
+            Use in memory dataset if possible, by default True.
```
jdeschamps: `use_in_memory` is gone, so it should be removed from here.

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,103 @@ def train(
...
+            a Path).
+        ValueError
+            If the training target is provided to N2V.
+        ValueError
+            If neither train_data nor a datamodule is provided.
+        """
+        if train_data is None:
+            raise ValueError(
+                "Training data must be provided. Either provide `train_data` or "
+                "use a datamodule (not yet supported)."
```
jdeschamps:
```suggestion
                "Training data must be provided. Provide `train_data`."
```

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,103 @@ def train(
...
+            )
+
+        # Check that inputs are the same type
+        source_types = {
+            type(s)
+            for s in (train_data, val_data, train_data_target, val_data_target)
+            if s is not None
+        }
+        if len(source_types) > 1:
+            raise ValueError("All sources should be of the same type.")
```
jdeschamps:
```suggestion
            raise ValueError(f"All sources should be of the same type. Got {source_types}.")
```

src/careamics/careamist_v2.py
```diff
@@ -247,22 +257,103 @@ def train(
...
+            val_data_target=val_data_target,
+            read_source_func=read_source_func,
+            read_kwargs=read_kwargs,
+            extension_filter=extension_filter,
+            val_percentage=0.1,
+            val_minimum_split=5,
+        )
+
+        # Register datamodule
+        self.train_datamodule = datamodule
```
jdeschamps:
```suggestion
        self.train_datamodule = CareamicsDataModule(
            data_config=self.config.data_config,
            train_data=train_data,
            val_data=val_data,
            train_data_target=train_data_target,
            val_data_target=val_data_target,
            read_source_func=read_source_func,
            read_kwargs=read_kwargs,
            extension_filter=extension_filter,
            val_percentage=0.1,
            val_minimum_split=5,
        )
```

tests/test_careamist_v2.py
```diff
@@ -0,0 +1,307 @@
...
+        experiment_name="test",
+        data_type="tiff",
+        axes="YX",
+        patch_size=(8, 8),
+        batch_size=2,
+        num_epochs=1,
+        roi_size=5,
+        masked_pixel_percentage=5,
+    )
+
```
jdeschamps: duplicated lines

```suggestion
```

tests/test_careamist_v2.py
```diff
@@ -0,0 +1,307 @@
...
+from careamics.config.ng_factories.n2v_factory import create_advanced_n2v_config
+
+
+def random_array(shape: tuple[int, ...], seed: int = 42):
+    """Return a random array with values between 0 and 255."""
+    rng = np.random.default_rng(seed)
+    return (rng.integers(0, 255, shape)).astype(np.float32)
+
+
+def test_v2_train_error_target_unsupervised_algorithm(
```
jdeschamps:
```suggestion
def test_target_unsupported_n2v(
```

tests/test_careamist_v2.py
```diff
@@ -0,0 +1,307 @@
...
+    with pytest.raises(ValueError):
+        careamics.train(
+            train_data=np.ones((32, 32)),
+            train_data_target=np.ones((32, 32)),
+        )
+
+
+@pytest.mark.skip("val splits not implemented")
+@pytest.mark.mps_gh_fail
+def test_v2_train_single_array_no_val(tmp_path: Path, minimum_n2v_configuration: dict):
```
jdeschamps: Since the arrays are simply passed to the Datamodule, I would suggest removing this test.

In addition, I would leave to the val/train split PR the responsibility to add tests.

tests/test_careamist_v2.py
jdeschamps, line 1: An additional test could be to test that an error is raised when no training data is passed.

tests/test_careamist_v2.py
```diff
@@ -0,0 +1,307 @@
...
+    # train CAREamistV2
+    careamist.train(train_data=train_array)
+
+    # check that it trained
+    assert Path(tmp_path / "checkpoints" / "last.ckpt").exists()
+
+
+@pytest.mark.skip("val splits not implemented")
+@pytest.mark.mps_gh_fail
+def test_v2_train_array(tmp_path: Path):
```
jdeschamps, line 96: Since the arrays are simply passed to the Datamodule, I would suggest removing this test.

Also this test is actually doing other things (see checkpoint parameters), these should be in a test with a better name and description, to check that parameters to `Trainer` are properly passed (and it may not require training, making the test faster).

jdeschamps (changes requested):
A few simple comments, the docs and commit suggestions are the requested changes.

The comment about removing tests can be extended to most tests in this PR, but they are part of a larger discussion about testing and organization, so these comments could be ignored.

src/careamics/careamist_v2.py
melisande-c: For the validation splitting integration, I have ended up refactoring the datamodule initialisation a little bit and in the process have changed the input types to this:
```
ArrayInput = NDArray[Any] | Sequence[NDArray[Any]]
PathInput = str | Path | Sequence[str | Path]
InputType = ArrayInput | PathInput
"""Type of input data passed to the dataset."""
```
In the docs it could link to these defined types, maybe that is good enough?

melisande-c: Although there is different behaviour for path vs list of paths, the former being treated as a directory, which we should document for sure

src/careamics/careamist_v2.py
melisande-c: Actually the check above might already happen in the initialisation of the datamodule, so maybe we don't need the duplication

jdeschamps: Yes, we could delegate that to the Datamodule.

veegalinova: Removed the check, datamodule checks that "data pair" of input and target have the same types

tests/test_careamist_v2.py
veegalinova: I think that passing taget should not fail to start the training, so I replaced it with a warning and modified this test

src/careamics/careamist_v2.py
veegalinova: I ended up updating the type annotations to what Milly proposed above, updated docstring to
```train_data : numpy.ndarray, str, pathlib.Path, or sequence of these, optional```

tests/test_careamist_v2.py
jdeschamps: That's a nice idea actually

jdeschamps (approved):
This PR will clash with https://github.com/CAREamics/careamics/pull/789, just flagging that this PR needs to be merged first.

PR #762, opened 2026-02-08
Restore original prediction shape

### LINKED ISSUE #654: Convert predictions back to original shape and dimension order
## Problem

CatEek:
## Description

> [!NOTE]  
> **tldr**: Add option to restore original prediction shape and dimension order. 

**IMPORTANT**: SOME TESTS FAIL BUT WE CAN START DISCUSSING, BECAUSE I"M TIRED AND MAYBE I DID SOME WACKY STUFF 

Because now they might not be in original order and (some) users don't like that 

### Overview - what changed?

Added infrastructure to track and restore original shape/axes:
1. Store original shape and axes at the `ImageStack` level before `reshape_array`
2. Implement `restore_original_shape` to reverse the `reshape_array` transformation
3. Add `restore_shape` parameter to `convert_prediction` to enable shape restoration

- Added `original_data_shape` and `original_axes` properties to the `ImageStack` protocol and implemented them in all concrete classes (`ZarrImageStack`, `InMemoryImageStack`, `FileImageStack`, `CziImageStack`)
- Extended `ImageRegionData` with `original_data_shape` and `original_axes` fields (default empty for backward compatibility)
- Updated `CareamicsDataset._create_image_region` to populate these fields from the `ImageStack`
- Implemented `restore_original_shape`
- Added `restore_shape` parameter to `convert_prediction`

### New features or files
- `original_data_shape` and `original_axes` properties added to `ImageStack` protocol
- `restore_original_shape` function in `convert_prediction.py`
- `restore_shape` parameter in `convert_prediction` function

- All `ImageStack` implementations (`zarr_image_stack.py`, `in_memory_image_stack.py`, `file_image_stack.py`, `czi_image_stack.py`) to store and expose original shape/axes
- `ImageRegionData` NamedTuple to include `original_data_shape` and `original_axes` fields
- `CareamicsDataset._create_image_region` to propagate original shape info
- `decollate_image_region_data` to handle tuple/list types and extract original shape info correctly
- `combine_samples` to preserve C dimension when stacking
- `CAREModule` and `N2VModule` `predict_step` methods to propagate original shape info

## How has this been tested?

- Added `test_restore_original_shape_styx` in `test_ng_convert_prediction.py`

## Related Issues

- Resolves #654 

---

- [] Code builds and passes tests locally, including doctests
- [] Pre-commit passes


src/careamics/dataset_ng/dataset.py
melisande-c: `axes` here is already the original axes so there is no need for the additional `original_axes`, the docstring is incorrect in saying "in format SCZYX", in reality "SCTZYX" are allowed in any order.

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -99,12 +121,97 @@ def decollate_image_region_data(
...
+        return prediction
+
+    # Determine current axes from prediction shape
+    ndim = len(prediction.shape)
+    if ndim == 5:
+        current_axes = "SCZYX"
+    elif ndim == 4:
+        current_axes = "SCYX"
+    elif ndim == 3:
+        current_axes = "SYX"
```
melisande-c: The case of `ndim == 3` will never happen in the pipeline, the data will always be converted to SC(Z)YX

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -99,12 +121,97 @@ def decollate_image_region_data(
...
             additional_metadata=additional_metadata,
+            original_data_shape=original_data_shape,
+            original_axes=original_axes,
         )
         decollated.append(image_region)
 
     return decollated
 
 
+def restore_original_shape(
```
melisande-c: It doesn't look like the case that T is present in the original axes has been taken into account

src/careamics/dataset_ng/dataset.py
```diff
@@ -73,6 +73,12 @@ class ImageRegionData(NamedTuple, Generic[RegionSpecs]):
     """Additional metadata to be stored with the image region. Currently used to store
     chunk and shard information for zarr image stacks."""
 
+    original_data_shape: Sequence[int] = ()
+    """Original shape of the data."""
+
+    original_axes: str = ""
+    """Original axes of the data."""
+
```
melisande-c: Why are there empty defaults, this should always be populated from the image stack in `_create_image_region`

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -91,6 +92,27 @@ def decollate_image_region_data(
...
+            isinstance(batch.original_data_shape, (list, tuple))
+            and batch.original_data_shape
+        ):
+            original_data_shape = tuple(
+                int(dim[i]) for dim in batch.original_data_shape
+            )
+        else:
+            original_data_shape = (
+                tuple(batch.original_data_shape) if batch.original_data_shape else ()
+            )
```
melisande-c: Looking at the code, `original_data_shape` is always added to the `ImageRegionData` from the image stacks so it should never be empty

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
```diff
@@ -99,12 +121,97 @@ def decollate_image_region_data(
...
+    original_data_shape : Sequence[int]
+        Original shape of the data.
+
+    Returns
+    -------
+    numpy.ndarray
+        Prediction reshaped to match original_axes and original_data_shape.
+    """
+    if not original_axes or not original_data_shape:
+        return prediction
```
melisande-c: I don't think we should cover the case that `original_axes` or the `original_data_shape` is empty

melisande-c (commented):
We also need to restore the shape in the `PredictionWriterCallback`

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
melisande-c: I had an idea how to take into account the T dimension, here is what I tried:

```python
def restore_original_shape(
    prediction: NDArray,
    original_axes: str,
    original_data_shape: Sequence[int]
) -> NDArray:
    ndim = len(prediction.shape)
    if ndim == 5:
        # T axis added below
        current_axes = "STCZYX"
    elif ndim == 4:
        current_axes = "STCYX"
    else:
        raise ValueError(
            f"Unexpected prediction shape {prediction.shape}. "
            "Expected 4D (SCYX), or 5D (SCZYX)."
        )

    # Add the T dimension back in
    ST_size = [1, 1] # S and T singleton unless present
    if "T" in original_axes:
        ST_size[1] = original_data_shape[original_axes.index("T")]
    if "S" in original_axes:
        ST_size[0] = original_data_shape[original_axes.index("S")]
    # replace S with (S, T)
    shape_with_T = (*ST_size, *prediction.shape[1:])
    prediction = prediction.reshape(shape_with_T)

    # all the added singletons will be left at the end
    axes_order = [current_axes.index(a) for a in original_axes]
    prediction = np.moveaxis(prediction, axes_order, np.arange(len(axes)))
    # remove singletons that weren't present in original data
    n_singleton_axes = len(current_axes) - len(original_axes)
    prediction = prediction.squeeze(tuple(d for d in range(-n_singleton_axes, 0)))

    return prediction
```

I tested with a few different data shapes and axes, for example:

```python
axes = "SYXTC" # testing weird order
original_shape = (6, 64, 32, 20, 3)

array = np.arange(np.prod(original_shape)).reshape(original_shape)
reshaped = reshape_array(array, axes)
restored = restore_original_shape(reshaped, axes, original_shape)

print(reshaped.shape, restored.shape)
print((array == restored).all())
```
```
(120, 3, 64, 32) (6, 64, 32, 20, 3)
True
```

melisande-c: Oh now I see that the "unflatten S dimension" section of the code was handling the case with T

melisande-c: hmm, I've thought about it more and realised that I didn't consider the case where CZI can flatten Z in S and treat T as Z. I think your "unflatten S" does cover this but it would be good to have a test for `restore_original_shape` that tests a few different possibilities

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: Fixed in https://github.com/CAREamics/careamics/pull/762/changes/193c362ea04e6ad638fe388220bc296c0a40b83a

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: T as Z was not covered, should be covered now with https://github.com/CAREamics/careamics/pull/762/changes/7f81758c273d24ac4c7d9a5bbbb1732e72bd3536

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: I don't see these lines in the code so I assume it has been resolved.

src/careamics/lightning/dataset_ng/prediction/convert_prediction.py
jdeschamps: I don't see these lines in the code so I assume it has been resolved.

jdeschamps:
> We also need to restore the shape in the `PredictionWriterCallback`

https://github.com/CAREamics/careamics/pull/762/changes/a1e073d083f0bb99f802615561518ef227a5cdf7 and https://github.com/CAREamics/careamics/pull/762/changes/76197007491cad72d2a32e76df0a07bbf08a329d moved the call to the reshaping function to the methods that directly handle combining and stitching predictions. As a consequence the tile and image writers are reshaping back to the original shape.

The Zarr tile writer is a bit more hairy. We would need to not reshape chunks and shards, then reshape the tile before inserting, which requires some array slicing engineering... I am not sure it is worth the pain, I will try to think about a potential solution.

Tests need to be fixed since some behaviours have changed due to the reshaping.

tests/lightning/dataset_ng/prediction/test_restore_original_shape.py
```diff
@@ -0,0 +1,88 @@
+import numpy as np
+
+from careamics.lightning.dataset_ng.prediction import restore_original_shape
+
+
+class TestRestoreShape:
```
melisande-c: Can we add a test that is a "round trip" through `reshape_array` and `restore_original_shape` since they should be the inverse of each other, parametrised with all the most common scenarios

melisande-c (approved):

tests/lightning/dataset_ng/prediction/test_restore_original_shape.py
jdeschamps: Great idea, it will not cover CZI cases, but at least the most common use cases.

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/762/commits/3f4f0789b1e44a98821ee23c21bbc59af6e6108e

jdeschamps:
I have decided to keep at that for this PR, two PRs will follow:

- Zarr tile writing to original shape
- Default reshaping to `True`, with all the tests adjusted for new behaviour (which impacts many tests throughout our test suite)

PR #767, opened 2026-02-11
Fix: 3D case for StructN2V + N2V2 with refac of median pixel manipulation

melisande-c:
## Description

> [!NOTE]  
> **tldr**: The function `median_manipulate_torch` made an assumption that patches are always 2D when applying the struct mask to calculate the median of a subpatch. This PR fixes that problem and refactors the `median_manipulate_torch` function so that smaller units of the code can be tested.

N2V2 uses the median pixel in a subpatch to mask the central pixel. When applying StructN2V, the pixels excluded by the struct mask should also not be included in the median calculation. In `median_manipulate_torch` the struct mask is created to exclude the pixels in the median calculation but it was only implemented to consider the 2D case, the line below shows where this assumption happened:

https://github.com/CAREamics/careamics/blob/fbcf24d5fe698033ff7fcbef992a437cf459ebd8/src/careamics/transforms/pixel_manipulation_torch.py#L345-L348

I also took the opportunity to refactor `median_manipulate_torch` into a few smaller functions so that these smaller units of code could be tested more thoroughly.

### Overview - what changed?

Struct mask creation has been moved to the function `_create_struct_mask`, it can handle arbitrary number of dimension.

Central pixel masking has been refactored to mirror struct masking to make the code clearer and the central pixel mask creation happens in `_create_center_pixel_mask`.

Most of the code in `median_manipulate_torch` was to create coordinates to extract the subpatches/rois from the patch in a vectorized way. This code has been refactored and moved to the function `_get_subpatch_coords`.

Subpatch coordinates are calculated in a similar way to the original implementation but uses torch broacasting to add the subpatch center to a meshgrid of coordinates rather than iterating through the dimensions.

### New features or files

- `_create_struct_mask`
- `_create_center_pixel_mask`
- `_get_subpatch_coords`

- `median_manipulate_torch`

## How has this been tested?

New tests for the new functions.

Added an additional parametrisation to `test_median_manipulate_torch`, which is the argument `apply_struct`. This only tests applying horizontal StructN2V, but for both 2D and 2D. 

Sanity checked the output with this code:

```python
import numpy as np
import torch

from careamics.transforms.pixel_manipulation_torch import median_manipulate_torch
from careamics.transforms.struct_mask_parameters import StructMaskParameters

shape = (2, 64, 64, 64) # BZYX
array = torch.arange(np.prod(shape).item(), dtype=torch.float32).reshape(shape)
mask_pixel_percentage = 0.08

z = 16
b = 0
fig, axes = plt.subplots(2, 2, figsize=(8, 8), constrained_layout=True)
fig.suptitle(f"Batch {b} | z-slice {z}")
axes[0, 0].set_title("Mask")
axes[0, 1].set_title("Manipulated Patch")

for struct_axis in [0, 1]:
    manip_median, manip_median_mask = median_manipulate_torch(
        array,
        mask_pixel_percentage,
        struct_params=StructMaskParameters(axis=struct_axis, span=5),
        rng=torch.Generator(),
    )

    axes[struct_axis, 0].imshow(manip_median_mask[b, z])
    axes[struct_axis, 1].imshow(manip_median[b, z])
    axes[struct_axis, 0].set_ylabel(f"Struct Axis {struct_axis}")
```
58ff62aa-3e73-4dcc-bd91-35cedbad49a2

---

- [x] Pre-commit passes


sebi06:
Hi all,

do you already have a rough estimate about when this fixed will be shipped?

src/careamics/transforms/pixel_manipulation_torch.py
```diff
@@ -291,99 +291,189 @@ def median_manipulate_torch(
...
 
     # Update the output tensor with medians
     output_batch = batch.clone()
-    output_batch[tuple(coords_axes)] = medians
-    mask = torch.where(output_batch != batch, 1, 0).to(torch.uint8)
+    output_batch[tuple(subpatch_center_coordinates.T)] = medians
+    mask = (batch != output_batch).to(torch.uint8)
 
     if struct_params is not None:
         output_batch = _apply_struct_mask_torch(
```
veegalinova, line 336: here, rng generator is not passed into `_apply_struct_mask_torch` like in uniform manipulate

veegalinova (commented):
I like the refactoring, reads better than before! 
I did not see any issues (except one that was present before, see comment). But I am not super familiar with the original implementation, so it will certainly benefit from another set of eyes, if there is time.

src/careamics/transforms/pixel_manipulation_torch.py
melisande-c: Oh nice spot, I will fix

jdeschamps:
> Hi all,
> 
> do you already have a rough estimate about when this fixed will be shipped?

Hi @sebi06, we wanted to merge this much earlier but usual reviewers (me included) were coincidentally all unavailable at the same time...

This will probably be merged tomorrow and a new version pushed as soon as the (private GPU-enabled) CI passes!

sebi06:
Hi @jdeschamps 

That would be perfect. 

Sebi

jdeschamps (approved):

PR #777, opened 2026-02-18
Feat: per_channel and global normalization options

### LINKED ISSUE #679: Next-Gen Dataset: Add global channel normalization option
## Problem

Currently, normalization statistics are computed per channel. This works well for most cases, but for spectral data where the relative intensity between channels is meaningful, we need an option to compute statistics globally across all channels instead - as suggested in #655 

## Potential solution

Add a `per_channel` flag to normalization configs that allows computing a single set of statistics across all channels.

veegalinova:
## Description

> [!NOTE]  
> **tldr**: added `per_channel` option to normalization configs, allowing to compute global stats across all channels or separate values for each channel. 

Currently, normalization is applied per-channel and stats are also computed for each channel individually. For some use-cases, it's preferable to have one single value for all of the channels and have a possibility to compute it automatically. 

### Overview - what changed?

Now, statistics for normalization can be either a single float, or a list of floats. The values are validated at runtime against the incoming patch. One value is broadcasted over the channels. 
For automatic computation, user needs to specify `per_channel` field in the configuration (true by default). If false, for calculating the global value, the patch is flattened across all channels. 

- added `per_channel` field to all normalization configurations 
- modified statistics computation to support global value computation
- modified the normalization methods to broadcast a single value across all channels 
- added tests for the new functionality 

### New features or files

- `utils.py` with `broadcast_stats` and `reshape_stats` functions
- `per_channel` parameter in normalization configs

- `normalization_config.py` - added `per_channel` flag and validation
- `statistics.py` - updated computation functions to handle global stats
- `mean_std_normalization.py` and `range_normalization.py`  - use shared utilities, handle broadcasting

## How has this been tested?

Existing tests cover per-channel behavior, added tests for global normalization (for all normalization methods):
- `test_global_auto_computes_stats` - verifies single stat computed across channels
- `test_global_stats_pools_across_channels` - verifies channels are pooled correctly
- `test_global_stats_denormalization_round_trip` - verifies denormalization works

## Related Issues

- Resolves #679 

Example usage:
```python
# Global normalization (single stat across all channels)
config = NGDataConfig(
    ...
    normalization={"name": "mean_std", "per_channel": False}
)

# Per-channel normalization (default)
config = NGDataConfig(
    ...
    normalization={"name": "mean_std", "per_channel": True}
)
```
---

- [x] Pre-commit passes


melisande-c (approved):
Looks good!

Nice tidy-up of the tests and the stats resolving and broadcasting

PR #778, opened 2026-02-20
Feat(Next-Gen): add CARE and N2N configurations and factories

### LINKED ISSUE #732: Next-Gen CAREamics: add CARE and N2N convenience functions
## Problem

Once we have decided on standard vs advanced convenience functions, we should implement the equivalent CARE and N2N methods.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Adding convenience functions to create CARE and N2N configurations.

The convenience functions are the user entry point into configuring CAREamics.

## Changes Made

Added the following convenience functions:
- `create_care_config`
- `create_advanced_care_config`
- `create_n2n_config`
- `create_advanced_n2n_config`

Note that these all share an underlying `_create_advanced_supervised_config` since they essentially require the same parameters.

## Related Issues

- Closes https://github.com/CAREamics/careamics/issues/732

Note that the only reason that N2N and CARE are separated is to allow clarity for users (one require noisy data) and proper citation when exporting to BMZ. We could discuss whether this is something we want to continue maintaining.

---

- [x] Pre-commit passes


src/careamics/config/ng_configs/care_configuration.py
```diff
@@ -0,0 +1,11 @@
...
+
+from careamics.config.algorithms import CAREAlgorithm
+
+from .ng_configuration import NGConfiguration
+
+
+class CAREConfiguration(NGConfiguration):
+    """CARE-specific configuration."""
+
+    algorithm_config: CAREAlgorithm
```
melisande-c: Do the CARE and N2N configurations only narrow the type that the algorithm config has to be? Maybe we don't need these classes if there is no specific validation we need to do between the data and the algorithm like in the N2V config. We could just use the base `NGConfiguration` for care and n2n, what do you think?

melisande-c (approved):

src/careamics/config/ng_configs/care_configuration.py
jdeschamps: Both CARE and N2N have the same validators over the model (no n2v2, and no final activation, maybe int he near future also a validation for residual connection).

Basically they could share everything and just have `name=Literal["care", "n2n"]`, which could be used by BMZ to export the correct citations. That would require extracting the logic for citation and name away from the Pydantic algorithm models. 

That solution, however, does not remove `create_n2n_config` and `create_care_config` which are needed to simply set `name=Literal["care", "n2n"]`.

Their algorithm config could then be the same (`CAREAlgorithm`) and indeed use the general `NGConfiguration` as main config.

melisande-c: Tbh I wasn't suggesting combining the `CAREAlgorithm` and `N2NAlgorithm`, I was only suggesting to remove these new `CAREConfiguration` and `N2NConfiguration` since they do not add anything

Everything else can just stay the same for now

jdeschamps: Yes, I am just pushing the argument one step further to remove the amount of code duplication. It can also be left as is for now and further simplified later (not a fan of leaving to later things though ^^).

jdeschamps: I implemented just what you suggested in https://github.com/CAREamics/careamics/pull/778/changes/5c24c28bc8fb932a898bb783d006e571687f2e45!

jdeschamps:
I removed as suggested the CARE and N2N configs, using directly the NG one. In an effort to reduce the footprint of the new `care_n2n_factory.py` module, I used `locals()` to forward all parameters. I wanted to maintain the function signatures.

src/careamics/config/ng_factories/care_n2n_factory.py
```diff
@@ -0,0 +1,677 @@
...
+        Number of input channels.
+    n_channels_out : int or None, default=None
+        Number of output channels.
+
+    Returns
+    -------
+    NGConfiguration
+        Configuration for training CARE.
+    """
+    return create_advanced_care_config(**locals())
```
melisande-c, line 79: FYI: it looks like `Unpack[TypeDict]` for kwargs should be supported by mkdocstrings:
issue: https://github.com/mkdocstrings/python/issues/207
solving PR: https://github.com/mkdocstrings/griffe/pull/414

But probably numpydocsting will complain in pre-commit about extra arguments in docs

jdeschamps: Do you know if IDE are also covering that case? I am a bit afraid jupyter notebooks will not, and I feel having an explicit signature for `create_care_config` and `create_advanced_config` is the more important feature.

melisande-c: pylance in vscode worked for me last time I checked, this comment in the issue shows what pyright in vim does: https://github.com/mkdocstrings/python/issues/207#issuecomment-3306374223

melisande-c: but we can leave it for now and check all these things thoroughly another time

melisande-c (approved):

PR #789, opened 2026-02-25
Feature(Next-Gen): Integrate validation splitting with refactor of `CareamicsDataModule`

### LINKED ISSUE #416: Next-Gen Dataset: Validation data splitting

melisande-c:
## Description

> [!NOTE]  
> **tldr**: Integrates validation splitting into the data module. This required some refactoring because the patching strategies for the training and validation dataset have to be created first and then passed onto the dataset instances. There is also branching logic for if validation data is provided or a validation splitting percentage is chosen. This also led to some knock on effect refactoring. It is quite a lot of changes so I could try to split some of the refactoring into a first PR.

Allows automatic validation splitting using the previously introduced stratified patching approach.

### Overview - what changed?

- `CareamicsDataset` initialisation now takes a `patching_strategy` as an argument, the `create_dataset` factory function now initialises the patching strategy and passes it onto the dataset.
- parameters to the dataset for custom loading have been grouped together into two data classes `ReadFuncLoading` and `ImageStackLoading`, this makes it clear that the two sets of parameters are mutually exclusive.
- Creation of datasets in the datamodule is now handled by the 3 functions `_create_train_val_datasets`, `_create_val_split_datasets` and `create_pred_dataset`.
  - These functions could probably do with more refining and I will probably have to handle it when implementing the more efficient background removal.
- The 7 overloads for `CareamicsDataModule` initialisation have been replaced with two that I think are actually useful for users. 
  - The first, which will be the default, narrows the input type to only arrays and paths (or sequences of either) for no custom loading or "read function loading".
  - The second widens the input type to `Any` if using custom image stack loading.
  - This means that an ide will show a type error if the inputs are not arrays and paths if an image stack is not provided.
- validations of inputs during the datamodule initialisation has been moved to a `_validate_data` function, this also groups the data into new dataclasses `_TrainVal`, `_TrainValSplit`, and `_Pred`, this makes later branching logic easier to deal with.
  - Currently this means prediction is mutually exclusive with training, but it can easily be changed.
-  typing in `src/careamics/lightning/dataset_ng/data_module_utils.py` has been updated leading to some code that checked types being rendered redundant so it was removed.
   - One consequence of the refactoring in this file meant that lists with arrays and non arrays is now not allowed when they used to be filtered out.

EDIT: Additional changes
- validation splitting is now only parametrised by `n_val_patches` and this is specified in the `NGDataConfig`.
- stratified patching is now the default patches strategy set in `create_ng_data_configuration`.

`_create_val_split_datasets` uses the function `_create_val_split`, previously introduced in #416, to randomly exclude patches from the training patching strategy to be used for validation.

### New features or files

- `ReadFuncLoading` and `ImageStackLoading` dataclasses, for grouping parameters
- `_TrainVal`, `_TrainValSplit`, and `_Pred` dataclasses, for grouping data and simplifying branching logic.
-  `_validate_data`, `_create_train_val_datasets`, `_create_val_split_datasets` and `create_pred_dataset` helper functions.

- `CareamicsDataset.__init__`
- `create_dataset` function
- `CareamicsDataModule` class, mostly things have just been moved to different functions.
- every function in `src/careamics/lightning/dataset_ng/data_module_utils.py`, typing changes and removing redundant code

### Removed features or files

- overloads from `CareamicsDataModule`.

## How has this been tested?

new test for `_validate_data`
new smoke test for validation train splitting that ensures the validation and training patches do not overlap.
Other changes should be covered by existing tests.

## Related Issues

- Resolves #416 

## Breaking changes

Custom loading arguments being combined into one argument in dataset creation.

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -13,27 +14,90 @@
...
+
+
+@dataclass
+class _PredData(Generic[T]):
+    """Data for prediction."""
 
-InputType = Union[ItemType, Sequence[ItemType], None]
-"""Type of input data passed to the dataset."""
+    pred_data: T
+    pred_data_target: T | None = None
```
jdeschamps: Why do we have a `pred_data_target`? Is the idea to allow the possibility to compute metrics?

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -50,30 +114,24 @@ class CareamicsDataModule(L.LightningDataModule):
     val_data_target : Optional[InputType]
         Validation data target, can be a path to a folder,
         a list of paths, or a numpy array.
+    val_percentage : Optional[float]
```
jdeschamps: We started annotating with `float | None` rather than `Optional[float]`. We should stick to one.

I think the reason was rather no thought-through but a consequence of ruff with `py>3.11`. So strong opinions could tip the balance one way or another.

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -122,120 +151,41 @@ class CareamicsDataModule(L.LightningDataModule):
...
+        loading: ReadFuncLoading | None = None,
     ) -> None: ...
 
-    # image stack loader (no mask)
+    # if using ImageStackLoading the input data can be anything.
     @overload
     def __init__(
         self,
-        data_config: NGDataConfig,
+        data_config: NGDataConfig | dict[str, Any],
```
jdeschamps, line 144: What is the reason to allow `dict[str, Any]`? That implies that an unvalidated configuration could find its way through.

jdeschamps: I see that this is covered by an `if` statement later. Is this a potential consequence of loading the config from a checkpoint? How would otherwise such a dict be passed?

src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -536,9 +416,274 @@ def predict_dataloader(self) -> DataLoader:
...
+        config=validation_config,
+        inputs=data.val_data,
+        targets=data.val_data_target,
+        loading=loading,
+    )
+
+    return train_dataset, val_dataset
+
+
+def _create_val_split_datasets(
```
jdeschamps: Most of the logic here could be refactored in a `create_split_dataset` in `factory.py` to be consistent with `_create_train_val_datasets`

src/careamics/lightning/dataset_ng/data_module_utils.py
```diff
@@ -363,33 +427,27 @@ def initialize_data_pair(
...
+                data_type, data[0], data[1], extension_filter=ext
+            )
+        case (SD.CUSTOM, ImageStackLoading()):
+            input_data, target_data = input_data, target_data
+        case _:
             raise ValueError(
-                f"Unsupported input type for {data_type}: {type(input_data)}"
+                f"Invalid combination: data_type={data_type!s}, "
+                f"input type is {type(input_data)}. For custom data a `read_func` or "
+                "and `image_stack_loader` must be provided."
```
jdeschamps: `read_func` and `image_stack_loader` read like pre-refactoring parameters. Can you adapt the error message to the new `loading` parameter?

jdeschamps (approved):
I like that refactoring, it is quite elegant! Only minor comments or questions.

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: yeah I guess that is the idea, I kept it because it was already there

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: Ah I will change it, this is just here because I reordered the arguments so I also reordered the docs but without changing the content

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: yes it is there to possibly load from a checkpoint, it was added in #717 but maybe it got removed again through some erroneous merging

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: I was thinking that tbf, I can move these 3 functions to factory.py?

src/careamics/lightning/dataset_ng/data_module.py
jdeschamps: Or I just don't remember it!

src/careamics/lightning/dataset_ng/data_module.py
jdeschamps: Yes, that would make sense to me!

melisande-c: done in [a985961](https://github.com/CAREamics/careamics/pull/789/commits/a985961b6de06b13a7a0b69331eb2e07db1c1f2b)

src/careamics/lightning/dataset_ng/data_module.py
melisande-c: updated in [e92a5e6](https://github.com/CAREamics/careamics/pull/789/commits/e92a5e6db389f6d30e1e9fb2721fe9ef1daa8e8b), also added more detailed description to some parameters

src/careamics/lightning/dataset_ng/data_module_utils.py
melisande-c: updated in [2d6dc7e](https://github.com/CAREamics/careamics/pull/789/commits/2d6dc7e3b62767b70a256c3dee6b2489dddabd9a)

melisande-c:
A consequence of updating the `_initialise_data_pair` function is that now if a train_target is provided, a validation target also has to be provided so I had to update this test that was introduced in #761 and provide a val target as well.
https://github.com/CAREamics/careamics/blob/2d6dc7e3b62767b70a256c3dee6b2489dddabd9a/tests/test_careamist_v2_train.py#L54-L60
Note that the warning will also be emitted when a target is provided but val splitting is chosen.

jdeschamps:
We discussed it face to face, but for transparency I am adding the content of the discussion here: the original thought was to removed `ImageStackLoader` parameters because we could pass a `DataModule`. Now we don't even accept a `DataModule` as input to train and predict, so this functionality as disappeared altogether. Since you refactored the various parameters into `ImageStackLoader` and `ReadFunc`, it would be easy to add the image stack loaders parameters as input back again.

I think it makes sense, as it is a powerful way to load custom data, and even though people using it might be technical enough to deal with the lightning API, they might not want to. So there is value in having it.

We should document with examples in the docstrings how to use it (later on) and clearly explain it on the website/docs.

jdeschamps (approved):

melisande-c:
Following an in-person discussion with @jdeschamps I have made the following changes:
- validation splitting is now only parametrised by `n_val_patches` and this is specified in the `NGDataConfig`.
- stratified patching is now the default patches strategy set in `create_ng_data_configuration`.

PR #790, opened 2026-02-26
Feat(Next-Gen CAREamist): improved array reshaping and restoring

### LINKED ISSUE #654: Convert predictions back to original shape and dimension order

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: The methods for reshaping array to SCZYX and back (the latter introduced in https://github.com/CAREamics/careamics/pull/762) are difficult to read and maintain. This PR introduces a potential replacement.

`dataset_utils.py:reshape_array` and `restore_original_shape.py:restore_original_shape` are methods used to reshape arrays to and from SCZYX-space. While they are concise, they are difficult to read, and do not allow accessing intermediate states necessary for reshaping tiles (no S dimension) or building the stitching coordinates to write such tile into an array in the original space.

This PR introduces an alternative approach to improve maintainability, readability and application to stitching and tile reshaping. Currently, the new source code is self-contained and not integrated.

A future PR would replace existing methods by those proposed here, ensuring the output is the same.

### Overview - what changed?

New `reshape_arrays.py` module which contains all methods aim at replacing the existing ones, as well as the methods missing to reshape tiles and write them into arrays (Zarr example).

Implemented `AxesTransform`, a dataclass that computes various properties related to the transformation from original space to SCZYX. Using only `original_axes` and `original_shape`, the knowledge of the various steps (merging S and T, adding singleton dims, re-ordering of the axes) and results (resulting shape) can be stored using several parameters in the data class.

This dataclass is used by several utilities to reshape or restore arrays and tiles.

### New features or files

-`AxesTransform`: a dataclass holding all the information necessary to reshape or restore arrays between original and SCZYX space.
- `reshape_array`: reshape array from original shape to SCZYX.
- `restore_array`: restore an array from SCZYX to its original shape.
- `restore_tile`: same as `restore_array`, but with the additional constrain that tiles do not have a sample dimension.
- `get_original_stitch_slices`: return a tuple of indices and slices used to index an array in original space in order to write a tile.

## How has this been tested?

New tests have been implemented.

While I wanted to also tackle the CZI case (where if T is used as depth axis, original axes might be "SCTYX" but S and T are not multiplexed). In other cases, namely SCYX and SCZYX, all axes are already in the right order, no need for reshaping or restoring. Given that CZI code only uses axes to determine the depth axis, there is no need for reshaping or restoring. 

So currently the best course of action in my mind would be to make these functions explicitly incompatible with that case, and assume in the code that restoring array will never be done in the context of CZI.

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
```diff
@@ -0,0 +1,431 @@
...
+        Computed property. Whether C is added as a singleton.
+    has_z : bool
+        Computed property. Whether original data contains a Z axis.
+    t_becomes_s : bool
+        Computed property. Whether T is the sole sample dim, relabeled as S.
+    st_merged : bool
+        Computed property. Whether S and T are both sample dims and merged together.
+    dim_sizes : dict[str, int]
+        Computed property. Map from axis name to original size.
+    dl_axes : str
```
melisande-c: what does DL stand for?

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
```diff
@@ -0,0 +1,431 @@
...
+        sizes = tuple(transform.dim_sizes[d] for d in sample_dims)
+        array = array.reshape(sizes + array.shape[1:])
+        current_axes = list(sample_dims) + current_axes[1:]
+    elif len(sample_dims) == 1 and sample_dims[0] != "S":
+        # relabel sample axis (typically S to T)
+        current_axes[0] = sample_dims[0]
+    elif len(sample_dims) == 0:
+        # remove singleton
+        array = np.squeeze(array, axis=0)
+        current_axes.pop(0)
```
melisande-c: II'm pretty sure your code for the first case actually covers all 3 cases already

```python
# sizes will be empty for no sample dims so will effectively be a squeeze on the second line
sizes = tuple(transform.dim_sizes[d] for d in sample_dims)
array = array.reshape(sizes + array.shape[1:])
# similarly sample_dims will be empty list for no sample dims
current_axes = list(sample_dims) + current_axes[1:]
```

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
```diff
@@ -0,0 +1,431 @@
...
+    @property
+    def st_merged(self) -> bool:
+        """S and T are both sample dimensions and are merged together.
+
+        Returns
+        -------
+        bool
+            True if both S and T are in sample_dims, False otherwise.
+        """
+        return "S" in self.sample_dims and "T" in self.sample_dims
```
melisande-c: I'm not seeing the properties `t_becomes_s` and `st_merged` being used anywhere, do you have plans to use them in the future?

melisande-c (approved):
If this made it easier to write the `restore_tile` and `get_original_stitch_slices` then I'm happy to accept, I think by it's nature these functions will always be a bit clunky, but if it works and we don't have to change it that's fine 😄

Couple of minor comments

jdeschamps:
Closes https://github.com/CAREamics/careamics/issues/654

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
jdeschamps: Deep-learning... I lacked creativity

jdeschamps: Changed to `transformed_*` for clarity!

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
jdeschamps: Oh, good catch!

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
jdeschamps: Ah yeah, I was using this early on and then only for tests. But it makes little sense. Removed them and replaced by equivalent expressions in the tests.

https://github.com/CAREamics/careamics/pull/790/changes/b84d59d259f6db5db548f862ee1aea4b74e4bc22

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
jdeschamps: Fixed in https://github.com/CAREamics/careamics/pull/790/changes/b84d59d259f6db5db548f862ee1aea4b74e4bc22

src/careamics/lightning/dataset_ng/prediction/reshape_arrays.py
jdeschamps: Fixed in https://github.com/CAREamics/careamics/pull/790/changes/b84d59d259f6db5db548f862ee1aea4b74e4bc22

PR #796, opened 2026-03-02
feat: raise ValidationError for batch_size in dataloader params (closes #793)

### LINKED ISSUE #793: `batch_size` present in `NGDataConfig` dataloader params should raise a validation error
## Problem
The `NGDataConfig` contains the explicit `batch_size` attribute along with dictionaries for train, val and predict dataloader params. If `batch_size` is present in any of the dictionaries a `SyntaxError` for duplicate keyword argument will be raised, because this is how the dataloaders are initialised in the data module:

https://github.com/CAREamics/careamics/blob/f41c7d032eb6e21c307f9b5807b1f34346a11caa/src/careamics/lightning/dataset_ng/data_module.py#L493-L506

## Potential solution

We should catch this early in the config and raise a clear validation error.

jinukuntlaakhilakumargoud-web:
Closes #793

### Description
Added a @field_validator in NGDataConfig across train_dataloader_params, val_dataloader_params, and pred_dataloader_params.
This instantly raises a helpful ValueError if a user attempts to manually configure batch_size via the dataloader_params mapping instead of explicitly setting the batch_size attribute of the configuration class, thus avoiding the obscure SyntaxError: keyword argument repeated error.

### Testing
- Added tests in tests/config/data/test_ng_data_config.py to verify that providing {"batch_size": 16} within any of the three dataloader params properly errors out during Pydantic initialization.

### DIFF
--- a/src/careamics/config/data/ng_data_config.py
+++ b/src/careamics/config/data/ng_data_config.py
@@ -577,6 +577,41 @@ class NGDataConfig(BaseModel):
                 self.patching.seed = self.seed
         return self
 
+    @field_validator(
+        "train_dataloader_params",
+        "val_dataloader_params",
+        "pred_dataloader_params",
+        mode="before",
+    )
+    @classmethod
+    def check_batch_size_not_in_dataloader_params(
+        cls, dataloader_params: dict[str, Any]
+    ) -> dict[str, Any]:
+        """
+        Check that 'batch_size' is not passed directly via dataloader_params.
+
+        Parameters
+        ----------
+        dataloader_params : dict of {str: Any}
+            The dataloader parameters.
+
+        Returns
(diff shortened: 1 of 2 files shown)


jdeschamps:
Hi,

Thank you for your interest in CAREamics. We have not yet put out our contributing guidelines, so I apologize for this. Unfortunately, while it was good etiquette to ask for our agreement before working on it, you have not waited for our answer.

We prefer to reserve such issues for genuine users of CAREamics or our junior devs. In addition, I don't really pick up on any signal showing that this is not the product of a fully automated AI agent, which is an additional road block for accepting this PR. Indeed, we will not review work that has not been reviewed by another human first, for reasons that should be obvious.

Finally, there is PR template!

PR #798, opened 2026-03-03
Feat: integrate new reshaping/restoring implementations, zarr tile shape restoration

### LINKED ISSUE #654: Convert predictions back to original shape and dimension order

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Substitute old implementations of reshaping and restoring arrays by the methods and classes from https://github.com/CAREamics/careamics/pull/790/. Implement array shape restoration in zarr tiling prediction writer.

Follow up PR to https://github.com/CAREamics/careamics/pull/790, which introduced a small framework for reshaping and restoring arrays, tiles, shapes, and slices. This PR:

- Tested new implementations against old ones (see intermediate commits, removed in the last commit alongside old implementations)
- Substituted new implementations to the old ones
- Implemented zarr tile restoring and saving in the zarr tile writer

Since https://github.com/CAREamics/careamics/pull/790, new methods have been implemented for reshaping and restoring arrays between original shape and SCZYX space. These methods were not integrated.

In particular, there is no implementation of zarr tile restoration (reshaping of both stitching slices and tiles).

### Overview - what changed?

Old methods (`reshape_arrays`, `restore_original_shape`, `reshape_array_shape`) have been substituted for the implementations from https://github.com/CAREamics/careamics/pull/790. Thus consolidating all array and shape manipulation into a single module.

Regarding substitution:
- In a first step, I added tests comparing old vs new implementations (see [here](https://github.com/CAREamics/careamics/blob/faa53b0c2971424c63d9761372378d23561dcd32/tests/utils/test_reshape_array.py#L408)).
- I then replaced all imports by those of the new implementation.
- Finally, I removed tests comparisons and old implementations.

Regarding the zarr tile writing:
- Removed reshaping from `_chunk_size` and `_shard_size` in `ZarrImageStack`
- Created array using original shape rather than current shape in `WriteTilesZarr`
- Used `restore_tile` on tile and `get_original_stitch_slices` to index the array on the original dimensions  in `WriteTilesZarr`

- Removed `original_shape` parameter from `reshape_array` as it could be inferred from array directly
- Removed `array_shape` parameter from `get_original_stitch_slices` as it could be inferred from the other parameters
- Moved `reshape_array.py` module (renamed to remove the plural) to `careamics.utils`
- Modified `ZarrImageStack` to not reshape `_chunk_size` and `_shard_size`
- Adapted `write_tiles_zarr_strategy.py` to create arrays with original shape, restore tile shape and reorder stitching slices
- All files calling `reshape_arrays` (old implementation), `restore_original_shape`, and `reshape_array_shape` have been updated to use functions from `reshapearray.py` module (new implementation)

### Removed features or files

- Removed `restore_original_shape` module
- Removed `reshape_array_shape` method from ImageStack utils
- Removed `reshape_array` from dataset utils

## How has this been tested?

Tests.

## Related Issues

Actually, this finally 
Resolves https://github.com/CAREamics/careamics/issues/654

## Breaking changes

Any code still relying on the old implementations.

---

- [x] Pre-commit passes


melisande-c (approved):
Looks good!

PR #801, opened 2026-03-03
Feat: add residual connection to UNet

### LINKED ISSUE #785: Residual UNet for CARE and N2V
## Problem

CSBDeep uses a residual UNet (residual connection on by default), while `n2v` switches it on for a few examples only.

We should implement that residual connection and benchmark whether it makes a difference on various examples.

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Add residual connection in UNet for future benchmarking and comparisons with `CSBDeep` and `n2v`.

This PR adds the residual connection to the UNet, since it was used by default for `CARE` (see CSBDeep) and on some examples of Noise2Void (e.g. BSD68). Quick tests have not shown a particular improvement on BSD68 and Noise2Void, but it should be integrated to our benchmarks in the future.

Since existing CSBDeep and some n2v examples have the residual connection on, we should have this possibility in CAREamics for reproducibility.

### Overview - what changed?

Modified the UNet module so that the forward method adds the input to the output of the encoder.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/785

## Breaking changes

No breaking change.

---

- [x] Pre-commit passes


src/careamics/models/unet.py
```diff
@@ -442,8 +448,14 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         torch.Tensor
             Output of the model.
         """
+        input_tensor = x
```
veegalinova: This should be good since I can not see any inplace operations on x. But just to be sure, we might do:
```suggestion
        input_tensor = x.clone()
```

veegalinova (approved):

PR #805, opened 2026-03-06
Fix: fix ng config loading

jdeschamps:
## Description

Fix an error preventing `NGConfiguration` from being read after export. `Mode` was exported as python object to yaml (meaning with a yaml tag). One easy work around is to ensure that `Mode` is exported as a string. Casting to the enum is taken care by Pydantic.

In addition, configurations other than N2V were ignored by the type.

I have not implemented the test, as it will depend on fixtures from the testing PR that was discussed internally.


melisande-c (approved):
Nice

PR #807, opened 2026-03-09
Feat: improve checkpoints name by adding experiment name and val loss

### LINKED ISSUE #792: Include experiment name in last checkpoint and export loss for k-best
## Problem
1. If multiple experiments are run in the same working directory, the last checkpoints will be saved as `"last.ckpt"` `"last-v1.ckpt"` etc. this makes it hard to know which experiment the checkpoint was from. This was encountered as a problem in https://github.com/CAREamics/careamics/issues/779#issuecomment-3942495862.

2. Additionally, with the k-best checkpoints they are saved only with the number of steps and the epoch, but no way to know which had the best validation loss.

## Potential solution
1. We can set the name of the last checkpoint to include the experiment name as documented in the lightning docs https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html:

    it has to be done through setting the attribute
    ```python
    checkpoint_callback.CHECKPOINT_NAME_LAST = f"{experiment_name}-last"
    ```

2. The validation loss of the k-best checkpoints could be included in the file name, or at the end of training we could save the data from `ModelCheckpoint` in a file.

diyasrivas:
## Description

> [!NOTE]  
> **tldr**: 
Added the experiment name in the last checkpoint filename and the validation loss in k-best checkpoint filenames.

When multiple experiments are run from the same working directory, the last checkpoint is always saved as last.ckpt, producing versioned names (last-v1.ckpt, last-v2.ckpt, etc.) with no indication of which experiment they belong to. Additionally, the k-best checkpoints are saved with only epoch and step in the filename, which makes it difficult to identify which checkpoint achieved the best validation loss without loading each file.

### Overview - what changed?

The last checkpoint filename now includes the experiment name ({experiment_name}_last.ckpt). The k-best checkpoint filenames now include the validation loss ({experiment_name}_epoch={epoch}_step={step}_{val_loss:.4f}.ckpt).

## Changes Made

Last checkpoint name: PyTorch Lightning's ModelCheckpoint exposes a CHECKPOINT_NAME_LAST attribute that controls the stem of the last checkpoint. Setting this on the constructed instance immediately after instantiation overrides the default last. I applied this identically in both `careamist.py` and `careamist_v2.py`. 

### New features or files

none

- `src/careamics/careamist.py`: `ModelCheckpoint` instantiation extracted to a named variable and `CHECKPOINT_NAME_LAST` set on the instance and `{val_loss:.4f}` appended to the filename pattern
- `src/careamics/careamist_v2.py`: Same changes as above in the `_define_callbacks` method
- `src/careamics/config/lightning/callbacks/callback_config.py`: Updated docstrings for `save_last` and `auto_insert_metric_name` 
- `tests/conftest.py`: Updated `pre_trained` and `pre_trained_v2` fixtures to expect `LevitatingFrog_last.ckpt` and `test_last.ckpt` respectively
- `tests/test_careamist.py`: Updated all last checkpoint path assertions from `last.ckpt` to `LevitatingFrog_last.ckpt` 
- `tests/test_careamist_v2_train.py`: Updated all last checkpoint path assertions from `last.ckpt` to `test_last.ckpt` 

### Removed features or files

none

## How has this been tested?

Existing tests were updated to assert the new checkpoint filenames and then tested locally:pytest tests/test_careamist.py tests/test_careamist_v2_train.py -v -k "checkpoint or train_array or train_export or train_tiff or pre_trained"
**Result: **22 passed, 48 deselected** — all relevant tests pass.**

## Related Issues

- Resolves #792 

## Breaking changes

I think that existing checkpoints saved under the old naming convention (`last.ckpt`) will not be found if code or workflows hardcode that filename (not sure about this).

---

- [x] Pre-commit passes


src/careamics/config/lightning/callbacks/callback_config.py
jdeschamps, line 1: I think we should actually walk away from explicit Callback config parameters as it is done for the other lightning classes (i.e. only a `checkpoint_callback_parameters: dict`), that would have different default set of parameters for CARE and N2V.

But that should be another PR.

src/careamics/careamist.py
```diff
@@ -254,15 +254,16 @@ def _define_callbacks(
                     "and should not be passed as callbacks."
                 )
 
-        # checkpoint callback saves checkpoints during training
+        checkpoint_callback = ModelCheckpoint(
+            dirpath=self.work_dir / Path("checkpoints"),
```
jdeschamps: Maybe we should also encode the name of the experiment in the folder where they are created. That may make the whole set up cleaner.

jdeschamps: But not in the old `careamist`.

src/careamics/careamist_v2.py
```diff
@@ -167,12 +167,14 @@ def _define_callbacks(
                     "internally and should not be passed as callbacks."
                 )
 
+        checkpoint_callback = ModelCheckpoint(
+            dirpath=work_dir / "checkpoints",
```
jdeschamps: Repeating the comment here where we may change it: Maybe we should also encode the name of the experiment in the folder where they are created. That may make the whole set up cleaner.

jdeschamps (approved):
I'd save the checkpoints in `workdir / checkpoints / f"{experiment_name}"`.

src/careamics/careamist_v2.py
jdeschamps: Maybe as `workdir / checkpoints / f"{experiment_name}"`

src/careamics/config/lightning/callbacks/callback_config.py
diyasrivas: okay will make a separate PR

jdeschamps:
@diyasrivas Could you test what happens when "experiment_name" in the configuration has spaces? Because we allow that and I foresee issues (folder creation, checkpoints).

Is experiment name used for logging as well?

Two possibilities:
- Replace spaces by underscores when using it in a f-string
- or remove spaces from allowed symbols in the `NGConfiguration`

PR #808, opened 2026-03-09
Feat: raise error when `batch_size` is in `dataloader_params`

### LINKED ISSUE #793: `batch_size` present in `NGDataConfig` dataloader params should raise a validation error

diyasrivas:
## Description

> [!NOTE]  
> **tldr**: 
Added a `field_validator` to `NGDataConfig` that raises a `ValueError` at config instantiation if `batch_size` is found in any of the three `*_dataloader_params` dicts.

 `NGDataConfig` exposes  `batch_size` as an explicit top-level field alongside three dataloader parameter dicts (`train_dataloader_params`, `val_dataloader_params`, `pred_dataloader_params`). In `CareamicsDataModule`, all three dataloader methods pass `batch_size` explicitly to `DataLoader` and then unpack the corresponding params dict:

```python
return DataLoader(
    self.train_dataset,
    batch_size=self.batch_size,
    **dataloader_params,  # TypeError if 'batch_size' is also here
)
```

If a user sets `batch_size` inside any of these dicts, Python raises a cryptic `TypeError: got multiple values for keyword argument 'batch_size'` deep at runtime, with no indication of where the fix should be made.

### Overview - what changed?

`NGDataConfig` now validates all  `*_dataloader_params` fields at instantiation time, rejecting any dict that contains `batch_size` with an error message.

I added a `@field_validator` covering all three fields with `mode="before"` to `NGDataConfig`.

### New features or files

none

- `batch_size_not_in_dataloader_params` validator added to NGDataConfig in `src/careamics/config/data/ng_data_config.py`
- New tests added to `tests/config/data/test_ng_data_config.py`

### Removed features or files

none

## How has this been tested?

- `test_batch_size_in_dataloader_params_raises` : parametrised over all three fields, verifies that passing `{"batch_size": 4}` to any of them raises `ValueError` at config construction time
- `test_valid_dataloader_params_accepted` : verifies that valid dicts without `batch_size` are accepted without error
- `test_batch_size_in_pred_dataloader_params_predicting_mode_raises` : verifies the validator fires correctly in `mode="predicting"` with a compatible patching strategy

All tests in `tests/config/data/` and `tests/lightning/` pass locally.

## Related Issues

- Resolves #793 
---

- [x] Pre-commit passes


src/careamics/config/data/ng_data_config.py
```diff
@@ -437,6 +437,44 @@ def validate_filters_against_mode(
             )
         return filter_obj
 
+    @field_validator(
+        "train_dataloader_params",
+        "val_dataloader_params",
+        "pred_dataloader_params",
+        mode="before",
```
jdeschamps: `mode="before"` means that the parameters might not be dictionaries and can lead to errors. In `after` mode, Pydantic made sure that they are dict.

```suggestion
        mode="after",
```

tests/config/data/test_ng_data_config.py
```diff
@@ -641,3 +641,58 @@ def test_removing_channels(self):
...
+    "field",
+    [
+        "train_dataloader_params",
+        "val_dataloader_params",
+        "pred_dataloader_params",
+    ],
+)
+def test_batch_size_in_dataloader_params_raises(field):
+    """Test that setting batch_size in any dataloader params dict raises ValueError."""
+    with pytest.raises(ValueError, match="batch_size"):
```
jdeschamps:
```suggestion
    with pytest.raises(ValueError, match="`batch_size` should not be set in the dataloader parameters."):
```

jdeschamps: Just to make it more strict and that the error does not relate to the `batch_size` main parameter.

tests/config/data/test_ng_data_config.py
```diff
@@ -641,3 +641,58 @@ def test_removing_channels(self):
...
+        data_type="array",
+        axes="YX",
+        patching={"name": SupportedPatchingStrategy.RANDOM, "patch_size": [16, 16]},
+        normalization=DEFAULT_NORM,
+        **{field: params},
+    )
+    assert getattr(config, field)["num_workers"] == 2
+
+
+def test_batch_size_in_pred_dataloader_params_predicting_mode_raises():
```
jdeschamps: The validation does not depend on the `mode`, so this test should be removed.

jdeschamps (changes requested):
Thanks! Some changes requested!

tests/config/data/test_ng_data_config.py
diyasrivas: removed it 👍

jdeschamps (approved):

PR #809, opened 2026-03-09
Refac: replace `transforms` by `augmentations` for clarity and coherence

### LINKED ISSUE #697: Refac: use `augmentation` throughout the code instead of `transformation`
## Problem

Parameters to the convenience functions are called `augmentations`, but the objects that need to be instantiated and passed live in the `transformation` module. We should stick to `augmentations`.

diyasrivas:
> [!NOTE]  
> **tldr**: 
Renames `config/transformations/` to `config/augmentations/` and renames the `transforms` field on `NGDataConfig` to `augmentations`.

The module `careamics.config.transformations` had to be renamed. It contains configs for spatial augmentations (flips, rotations) but not the full transform pipeline (which also includes normalisation, N2V manipulate, etc.). Parameters to the convenience functions are called augmentations, but the objects that need to be instantiated and passed live in the transformation module. Therefore we need to be using augmentations terminology. 

### Overview - what changed?

The changes can be explained in 2 parts:
1. `src/careamics/config/transformations/` renamed to `src/careamics/config/augmentations/`, with all import paths updated across `src/`, `tests/`, and `docs/` (with help from pylance).
2. The `transforms` field on `NGDataConfig` renamed to `augmentations`, with all downstream accesses updated. The equivalent field on `DataConfig` (old careamics) is deliberately left as `transforms`.

- The original careamics source code (`DataConfig`, `configuration_factories.py`, `dataset/` modules, `train_data_module.py`) uses `transforms` as a field name internally. Changing this is out of scope here and would be a larger breaking change with no benefit scoped to this issue, therefore I let it be.
- For the new careamics NG dataset (`NGDataConfig`, `ng_factories/`, `dataset_ng/`) where users interact with augmentation configs directly, the field rename from `transforms` to `augmentations` is clean and contained here.
- Class names inside `config/augmentations/` (`TransformConfig`, `XYFlipConfig`, `NormalizeConfig`, etc.) are **not** renamed because they refer to the transform objects themselves, not to the concept of augmentation (need to verify).

### New features or files

- `src/careamics/config/augmentations/` — renamed from `transformations/`, same 7 files with no content changes
- `tests/config/augmentations/` — renamed from `tests/config/transformations/`, same 4 test files

Part 1 aka import path updates:
- `src/careamics/config/algorithms/n2v_algorithm_config.py`, `pn2v_algorithm_config.py` import path updates
- `src/careamics/config/configuration_factories.py` import path + docstring/doctest updates; also contains two `"transforms"` → `"augmentations"` dict key fixes when constructing `NGDataConfig` (in pt.2)
- `src/careamics/config/data/data_config.py` relative import path update only; field name unchanged 
- `src/careamics/config/ng_factories/care_n2n_factory.py`, `data_factory.py`, `n2v_factory.py` import path updates + `"transforms"` → `"augmentations"` dict key when constructing `NGDataConfig`
- `src/careamics/dataset/` (6 files) import path updates only; these operate on `DataConfig`, field names unchanged
- `src/careamics/lightning/train_data_module.py` import path + doctest update; operates on `DataConfig`, field name unchanged
- `src/careamics/transforms/compose.py`, `n2v_manipulate_torch.py` import path updates only
- `docs/guides/careamist_api/configuration/build_configuration.py`, `convenience_functions.py` import path updates

Part 2: `NGDataConfig.transforms` → `NGDataConfig.augmentations`:
- `src/careamics/config/data/ng_data_config.py` field renamed from `transforms` to `augmentations`; model validator `propagate_seed_to_transforms` renamed to `propagate_seed_to_augmentations` with internal `self.transforms` updated accordingly
- `src/careamics/dataset_ng/dataset.py` - `self.config.transforms` → `self.config.augmentations`
- `tests/config/data/test_ng_data_config.py` constructor kwarg and field access updated
- `tests/config/ng_factories/test_care_factory.py`, `test_data_factory.py`, `test_n2v_factory.py` - `.transforms` → `.augmentations` on `NGDataConfig` instances

### Removed features or files

- `src/careamics/config/transformations/` replaced by `src/careamics/config/augmentations/`
- `tests/config/transformations/` replaced by `tests/config/augmentations/`
technically not removed just renaming shows it this way....

## How has this been tested?

- `pytest tests/config/augmentations/ -v` 8 passed (smoke check after committing Part 1)
- `pytest tests/config/ tests/dataset_ng/ tests/lightning/ -v` 466 passed, 2 skipped after committing Part 2

## Related Issues

- Resolves #697 

## Breaking changes

- `careamics.config.transformations` no longer exists so imports must be updated to `careamics.config.augmentations`
- `NGDataConfig.transforms` is renamed to `NGDataConfig.augmentations` so any code constructing or accessing this field by name will break

Notebooks in `src/careamics/dataset_ng/demos/` that reference `config.data_config.transforms` need to be updated at some point.
---

- [x] Pre-commit passes


src/careamics/config/configuration_factories.py
```diff
@@ -279,7 +279,7 @@ def _create_data_configuration(
         "axes": axes,
         "patch_size": patch_size,
         "batch_size": batch_size,
-        "transforms": augmentations,
+        "augmentations": augmentations,
```
jdeschamps: This looks like it introduces a bug: `DataConfig` does not have an `augmentations` field (and it shouldn't, we don't want to change that API), so `transforms` will be ignored.

src/careamics/config/augmentations/normalize_config.py
jdeschamps, line 1: Now it actually feels odd to have normalization and pixel manipulation (which are transforms), in a package called `augmentations`.

In the config, it makes sense to rename the field to `augmentations` since we are only using augmentations in there.

Since some clean up will be necessary, we can leave it at that for now. The normalization will any way not be used anymore, and we can update the n2v manipulate to not be a transform and move it away.

jdeschamps (changes requested):
This PR made me realized that further cleaning will be necessary in the augmentation configs module, but it can happen later.

One bug to fix and tests are failing:

```python
FAILED tests/config/test_configuration_factories.py::test_supervised_configuration_passing_transforms - AssertionError: assert 2 == 1
 +  where 2 = len([XYFlipConfig(name='XYFlip', flip_x=True, flip_y=True, p=0.5, seed=None), XYRandomRotate90Config(name='XYRandomRotate90', p=0.5, seed=None)])
 +    where [XYFlipConfig(name='XYFlip', flip_x=True, flip_y=True, p=0.5, seed=None), XYRandomRotate90Config(name='XYRandomRotate90', p=0.5, seed=None)] = DataConfig(data_type='tiff', axes='YX', patch_size=[64, 64], batch_size=8, image_means=None, image_stds=None, target_m...'shuffle': True, 'num_workers': 0, 'pin_memory': False}, val_dataloader_params={'pin_memory': False, 'num_workers': 0}).transforms
 +      where DataConfig(data_type='tiff', axes='YX', patch_size=[64, 64], batch_size=8, image_means=None, image_stds=None, target_m...'shuffle': True, 'num_workers': 0, 'pin_memory': False}, val_dataloader_params={'pin_memory': False, 'num_workers': 0}) = Configuration(version='0.1.0', experiment_name='test', algorithm_config=N2NAlgorithm(algorithm='n2n', loss='mae', mode...ic_name=False, every_n_train_steps=None, train_time_interval=None, every_n_epochs=None), early_stopping_callback=None)).data_config
FAILED tests/config/test_configuration_factories.py::test_n2v_configuration_no_aug - AssertionError: assert [XYFlipConfig...5, seed=None)] == []
  
  Left contains 2 more items, first extra item: XYFlipConfig(name='XYFlip', flip_x=True, flip_y=True, p=0.5, seed=None)
  
  Full diff:
  - []
  + [
  +     XYFlipConfig(name='XYFlip', flip_x=True, flip_y=True, p=0.5, seed=None),
  +     XYRandomRotate90Config(name='XYRandomRotate90', p=0.5, seed=None),
  + ]

```

jdeschamps (approved):

PR #812, opened 2026-03-10
Perf(Next-Gen): Improve Stratified Patching Speed

melisande-c:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: This PR improves the performance of the stratified patching strategy that was introduced in #710.

The first iteration of the patching strategy was not optimised.

### Overview - what changed?

All changes are contained within `careamics/dataset_ng/patching_strategies/stratified_patching.py`

Time savings on the initialisation and patch exclusion were found by improving the `_region_bin_packing` (which Claude originally wrote). This bin packing is a slight variation on the normal bin packing problem. We want to find an efficient bin packing but the bin capacity is allowed to expand (although we want to keep this to a minimum), instead of the usual problem where the bin capacity is fixed but more bins can be added. I think this is what confused Claude. Basically I had to remove a stupid binary search on the bin size and just increase the capacity as necessary when none of the bins could fit the next sampling region.

Time savings for the generation of patch specs were made by refactoring the way some values are stored. The main one being the region bins now contain indices instead of dictionary keys and the probabilities are stored as numpy arrays that can be directly accessed.

### New features or files

- Notebook to demonstrate patching strategy performance
  - `src/careamics/dataset_ng/demos/stratified_patching_perf.ipynb`

- `careamics/dataset_ng/patching_strategies/stratified_patching.py`

## How has this been tested?

- All tests pass
- additional test added for the bin packing algorithm 
- Performance tests using included notebook

#### Initialisation

The results below show that the random patching strategy has almost no initialisation overhead and performs with constant time relative to the number of patches.

In comparison the stratified patching strategy has some overhead for initialisation with what seems to be linear time. However for almost 10,000 patches this about 0.35s. 

(Patches were increased by increasing the number of samples or by increasing the image size to see if it had any effect.)

23b837ac-aba0-4689-83ae-ca3e1eb3733a

#### Patch specs generation

The results show that the stratified patching strategy is about 3x slower than the random patching strategy at generating patch specs. For 30,000 patches (roughly 30 epochs) the stratified patching will add about 1s of overhead whereas the random patching will add about 0.3s of overhead.

For 200 epochs of ~1000 patches the random patching strategy added an overhead of ~2.2s whereas the stratified patching strategy added an overhead of ~6.0s.

#### Patch distribution
dd49cba7-0ae6-478e-80c6-3aed9173585f

**Tested on**: MacBook Air, M3, 2024, 16GB Memory.

---

- [x] Pre-commit passes


src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -28,6 +27,23 @@
 # `_ImageStratifiedPatching` instance for each sample in each image stack.
 
 
+@dataclass
+class _RegionBins:
+    bins: list[NDArray[np.int_]]
+    """
+    A list of numpy arrays which correspond to indices of regions contained in the bin.
```
```suggestion
    A list of numpy arrays representing bins, each consisting of the indices of regions
    contained in the bin.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -28,6 +27,23 @@
...
+    """
+    probs: NDArray[np.floating]
+    """
+    The probability a region is selected per epoch.
+    The position corresponds to the region index.
```
```suggestion
    Probabilities to select each region bin each epoch.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -371,49 +392,34 @@ def sample_patch_coord(self, index: int) -> NDArray[np.int_]:
...
         # for index > no. of bins, it is effectively treated as an empty bin
         # all of sampling regions are selected proportionally to their area.
-
-        if index < len(self.bins):
-            bin_ = self.bins[index]
-        else:
-            bin_ = []
-
-        region = self._sample_region_from_bin(bin_)
+        region = self._sample_region_from_bin(index)
```
jdeschamps, line 388: The comment above that line probably needs to be updated (index > n_bins)? And moved to `self._sample_region_from_bin`?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -440,22 +446,16 @@ def exclude_patches(self, grid_coords: Sequence[tuple[int, ...]]):
             for d_idx in itertools.product(*[d for _ in range(self.ndims)]):
                 # q is the ID of the orthant to remove
                 q: tuple[Literal[0, 1], ...] = tuple(0 if i == 1 else 1 for i in d_idx)
-                grid_idx = tuple(
+                grid_coord_neighbor = tuple(
                     g - (1 - i) for g, i in zip(grid_coord, d_idx, strict=True)
                 )
-                if grid_idx not in self.regions:
+                if grid_coord_neighbor not in self.grid_coords:
```
jdeschamps, line 445: I am a bit confused by the "neighbor" here. In which sense?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -668,144 +685,97 @@ def clip(
...
-
-        # Add to best bin or create new one
-        if best_bin_idx is not None:
-            bins[best_bin_idx].append(key)
-        else:
-            bins.append([key])
-
-    return bins
-
+    items: dict[int, int],
```
jdeschamps: I actually liked `areas` better, because `items` makes it confusing to me (half way through, I forget what we are manipulating).

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -668,144 +685,97 @@ def clip(
...
     Parameters
     ----------
-    areas : dict[tuple[int, ...], int]
-        A dictionary of sampling region areas. The key is the grid coordinate of the
-        corresponding sampling region. These are the sampling regions created by
-        `_ImageStratifiedPatching`.
-    target_n_bins : int
-        The desired number of bins.
+    items : dict[int, int]
+        A dictionary where the keys correspond to an ID and the values correspond to
```
jdeschamps, line 725: Also if the description could be more specific about these volumes are that would help making sense of the function! (although I understand that the actual algorithm does not care)

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
jdeschamps: Just so I get it: here items is `{index of SamplingRegion : volume of the SamplingRegion}`?

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -668,144 +685,97 @@ def clip(
...
-            # Found exact match! Save it and keep searching for smaller bin size
-            best_bin_size = mid_size
-            best_num_bins = num_bins
-            max_size = mid_size - 1  # Try smaller bin size
-        elif num_bins > target_n_bins:
-            # Too many bins, need larger bin size
-            min_size = mid_size + 1
+    if n_bins == 0:
+        return 0, []
+    if len(items) <= n_bins:
```
jdeschamps: can you comment on when this would typically be the case? many excluded regions?

jdeschamps (approved):
Looks good! Only meaningless comments now that we spoke.

I think the docs need to be updated for _ImageStratifiedPatching, the `areas` attribute threw me off a bit:
https://github.com/CAREamics/careamics/blob/7b528aab4d09e3c84c23ac4c23857f2c7bc4765c/src/careamics/dataset_ng/patching_strategies/stratified_patching.py#L286

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: To exclude an area, we have to process the 4 neighbouring regions that cover the patch, so this is iterating over the 4 regions. It might not exist if we are excluding a patch at the edge

jdeschamps: Clear, thanks!

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: Ah I guess I changed the argument name to make the function more generic, but I can change it back, and yes that is what `items` is

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: Pretty sure the equality happens when the image size is perfectly divisible by the patch size; for less than, I'm not sure but it does provide a speed increase if it does happen

melisande-c: I think a particular pattern of excluded patches can cause it to be "less than"

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: done in [172820d](https://github.com/CAREamics/careamics/pull/812/commits/172820d4d887270ac715bd2ca835e694ebc0e17f)

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: added a comment in [172820d](https://github.com/CAREamics/careamics/pull/812/commits/172820d4d887270ac715bd2ca835e694ebc0e17f)

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
melisande-c: I removed most of the comments entirely because the bin packing function has been updated to return empty arrays for empty bins, instead of no bin. [8511779](https://github.com/CAREamics/careamics/pull/812/commits/851177965b28d0952d3dd2266b56c1fc40b2993a)

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
jdeschamps: With the updated description it is fine!

PR #815, opened 2026-03-12
Feat: explicit supervised error in `CAREamistV2`

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Raise an error in CAREamistV2.train` if `train_target_data` is missing for supervised algorithms.

Currently, if we forget to pass target data to CARE/N2N in CAREamistV2, then an obscure error is raised that will not make sense to users:

```python
    x, target = batch[0], batch[1]
                          ~~~~~^^^
IndexError: list index out of range
```

### Overview - what changed?

With this PR, CAREamistV2 will raise an error if `train_target_data` or `val_target_data` are missing.

Algorithm configs have now a method that returns whether the algorithm is supervised. These are called in turn in the `NGConfiguration`, so that any `config.is_supervised()` returns the correct value.

`CAREamistV2` calls `config.is_supervised()` and compares with the input data.

### New features or files

- `is_supervised` methods in `CAREAlgorithm`, `N2NAlgorithm`, `N2VAlgorithm` and `NGConfiguration`.

- `CAREamistV2.train` compares input and `config.is_supervised()`.

## How has this been tested?

Tests.
---

- [ ] Pre-commit passes

### DIFF
--- a/src/careamics/careamist_v2.py
+++ b/src/careamics/careamist_v2.py
@@ -311,6 +311,13 @@ class CAREamistV2:
                 f"ignored."
             )
 
+        if self.config.is_supervised() and val_data is not None and val_data_target is None:
+            raise ValueError(
+                f"Validation target data must be provided for supervised training (got "
+                f"{self.config.get_algorithm_friendly_name()} algorithm). Provide "
+                f"`val_data_target`."
+            )
+
         datamodule = CareamicsDataModule(
             data_config=self.config.data_config,
             train_data=train_data,
--- a/src/careamics/careamist_v2.py
+++ b/src/careamics/careamist_v2.py
@@ -298,6 +298,19 @@ class CAREamistV2:
         if train_data is None:
             raise ValueError("Training data must be provided. Provide `train_data`.")
 
+        if self.config.is_supervised() and train_data_target is None:
+            raise ValueError(
+                f"Training target data must be provided for supervised training (got "
+                f"{self.config.get_algorithm_friendly_name()} algorithm). Provide "
+                f"`train_data_target`."
+            )
+        elif not self.config.is_supervised() and train_data_target is not None:
+            logger.warning(
+                f"Training target data provided for self-supervised training (got "
+                f"{self.config.get_algorithm_friendly_name()} algorithm) will be "
+                f"ignored."
+            )
+
         datamodule = CareamicsDataModule(
             data_config=self.config.data_config,
             train_data=train_data,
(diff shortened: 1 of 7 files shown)


melisande-c (approved):
I guess LightningAPI users won't get this error, but only place it is possible to determine whether a target has been provided is in `training_step` of the lightning module where the original error is raised.

https://github.com/CAREamics/careamics/blob/7a3a2db4aa0799ca6fa427a41faf54965f05bb38/src/careamics/lightning/dataset_ng/lightning_modules/care_module.py#L110-L111

jdeschamps:
> I guess LightningAPI users won't get this error, but only place it is possible to determine whether a target has been provided is in `training_step` of the lightning module where the original error is raised.
> 
> https://github.com/CAREamics/careamics/blob/7a3a2db4aa0799ca6fa427a41faf54965f05bb38/src/careamics/lightning/dataset_ng/lightning_modules/care_module.py#L110-L111

Actually, @veegalinova solution to raise a warning in the N2V module is also applicable here:

https://github.com/CAREamics/careamics/blob/7a3a2db4aa0799ca6fa427a41faf54965f05bb38/src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py#L65-L76

[EDIT]: see https://github.com/CAREamics/careamics/pull/815/changes/fe69ae9c12e3e21be45d195aa12238da1dfda047

jdeschamps:
> > I guess LightningAPI users won't get this error, but only place it is possible to determine whether a target has been provided is in `training_step` of the lightning module where the original error is raised.
> > https://github.com/CAREamics/careamics/blob/7a3a2db4aa0799ca6fa427a41faf54965f05bb38/src/careamics/lightning/dataset_ng/lightning_modules/care_module.py#L110-L111
> 
> Actually, @veegalinova solution to raise a warning in the N2V module is also applicable here:
> 
> https://github.com/CAREamics/careamics/blob/7a3a2db4aa0799ca6fa427a41faf54965f05bb38/src/careamics/lightning/dataset_ng/lightning_modules/n2v_module.py#L65-L76
> 
> [EDIT]: see [fe69ae9](https://github.com/CAREamics/careamics/commit/fe69ae9c12e3e21be45d195aa12238da1dfda047)

The question now is whether to keep the `CAREamistV2` error handling because they are more efficient (fail earlier) and more specific (can state which parameter to pass), or we just let the Lightning modules take care of it and hope that the message is clear enough.

I'd still prefer to keep them, but there is value in not bloating the code.

melisande-c:
> The question now is whether to keep the `CAREamistV2` error handling because they are more efficient (fail earlier) and more specific (can state which parameter to pass), or we just let the Lightning modules take care of it and hope that the message is clear enough.
> 
> I'd still prefer to keep them, but there is value in not bloating the code.

happy to keep 👍

Issue #733, opened 2026-02-04
Next-Gen CAREamics: default model checkpointing parameters for CARE vs N2V


jdeschamps:
## Problem

To be discussed, my opinion is that we should save many checkpoints for N2V and only the best k (3?) for CARE. This behaviour, if wanted, should be implemented in the next-gen careamics.

Probably `training_config` should have different default values in `N2VConfig` and `CAREConfig`/`N2NConfig` (the latter don't exist as of yet).

CatEek:
I see the utility of that, but at the same time it might be annoying for some users that careamics will create many files. Especially if users aren't very familiar with how to use these many checkpoints. 2nd point is many chackpoints might be very very similar between each other in terms of produced results. So I'd say default 3 would be better. And we can explain in docs in which cases users can save more
Maybe we can first proceed with logging of intermediate images for N2V?

jdeschamps:
The problem is that choosing the best 3 in N2V does not insure anything about the quality of these checkpoints. I agree that it is not as much of a problem with other methods since N2V converges rapidly to something decent. Nonetheless, saving some random checkpoints because the value of the predicted validation pixels is closer to the original value sounds useless. You might as well just save the last one only.

For CARE, top 3 is useful.

jdeschamps:
Closed by https://github.com/CAREamics/careamics/pull/817

PR #817, opened 2026-03-12
Feat(Next-Gen): callbacks as dict in `NGTrainingConfig`, default preset by algorithm

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Define a new NG training configuration that does not define Lightning parameters for the callbacks, but rather use presets for behaviors tailored to the algorithm.

In the current version of the code, the callbacks are [fully-fledged configurations](https://github.com/CAREamics/careamics/blob/main/src/careamics/config/lightning/callbacks/callback_config.py), as opposed to the decision that has been made for Lightning related configurations to not explicitly maintain the parameters from Lightning API (to a few exceptions). In addition, different presets might be desirable for N2V and CARE (https://github.com/CAREamics/careamics/issues/733).

Following https://github.com/CAREamics/careamics/issues/733, this PR now introduces different checkpointing behaviours for N2V and CARE:
- CARE: save last checkpoint, save top 3 checkpoints according to `val_loss`
- N2V/N2N: save last checkpoint, save checkpoint every 10 epochs

Finally, the recently introduced `monitor_val` parameter was not passed to the early stopping callback. This led to issues when monitoring `train_loss_epoch` (e.g. when skipping validation with N2V), since the callback was monitor `val_loss`. This PR makes sure it is propagated as well to the early_stopping.

### Overview - what changed?

This PR introduces a NG training configuration that does not specify the Lightning callbacks parameters, but rather uses `dict`. It also introduces presets for CARE and N2V/N2N that govern the checkpointing behavior.

Note that `monitor_val` is now also passed to the early stopping callback.

A new `NGTrainingConfig` retains the `checkpoint_callback` and `early_stopping_callback` parameters, albeit as `dict` rather than Pydantic models. Then, presets `SupervisedCheckpointing` and `SelfSupervisedCheckpointing` are used in the `default_training_dict` to instantiate a valid checkpoint parameters dictionary.

Both `NGConfiguration` and `create_training_config` convenience function use the default factory for `NGTrainingConfig`.

### New features or files

- `NGTrainingConfig`, `SupervisedCheckpointing` and `SelfSupervisedCheckpointing`

Note that I took this opportunity to rename some parameters of the `NGTrainingConfig` to be coherent with the rest of the configurations:
- `lightning_trainer_config` -> `trainer_params`
- `checkpoint_callback` -> `checkpoint_params`
- `early_stopping_callback` -> `early_stopping_params`

- `create_training_configuration` now takes `monitor_val` (for the early stopping) and `algorithm` (to decide on the preset) 
- `CAREamistV2` use of the callback dicts
- Refactor the type adapter validation from `configuration_io` to `ng_config_discriminator` to help properly validate dictionaries as `NGConfiguration`. For reasons that I did not manage to understand, the type validation done in the checkpoint loading was not working (`algorithm_config` was valid but not passed to the default factory of the `NGTrainingConfig`), so I used the one that was already in the configuration io utils.

## How has this been tested?

Running a script, tests.

- Incoherence between `supervised_config` when it comes to "expecting a target" and `supervised_checkpoint` when it comes to the loss calculation (where N2N needs a target but its loss is self-supervised, meaning it does not hold much information about the model performances).
- Checkpoint creation explosion: with these presets, training N2V for 200 epochs will result in 20 checkpoints. I believe this is still acceptable, especially because the whole checkpoint folder might not be kept by users.

---

- [x] Pre-commit passes


jdeschamps:
Something seems to be off with the tests, they are hanging.

EDIT: found the problem.

src/careamics/config/utils/configuration_io.py
```diff
@@ -86,7 +53,7 @@ def load_configuration_ng(path: Union[str, Path]) -> NGConfigs:
 
     Returns
     -------
-    Configuration
+    NGConfigs
```
jdeschamps:
```suggestion
    Configuration
```

src/careamics/config/ng_configs/ng_training_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Because self-supervised algorithms are evaluating the loss against noisy pixels,
+    its value is not a good measure of performances after a few epochs. Therefore, it
+    cannot be used to evaluate the best performing models.
+
+    This presets saves checkpoints every 10 epochs, as well as the last one.
+    """
+
+    every_n_epochs: int = 10
+    """Save a checkpoint every 10 epochs."""
```
melisande-c, line 48: maybe we should consider `every_n_train_steps` because the length of an epoch can vary dramatically depending on the amount of data

src/careamics/config/ng_configs/ng_training_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+        Training configuration dictionary with the specified parameters.
+    """
+    # user parameters take precedence over defaults
+    # since resulting checkpointing behaviour depends on complex interactions between
+    # parameters, we keep either user defined or the defaults
+    if checkpoint_params is None:
+        # select default checkpointing preset based on algorithm
+        default_preset = (
+            SupervisedCheckpointing
+            if algorithm == "care"
```
melisande-c, line 99: so n2n uses SelfSupervised? it makes sense because like n2v, the minimal val loss doesn't necessarily mean the best model performance. I guess naming-wise it's fine that we just know why n2n is using self supervised

tests/unit/config/ng_configs/test_ng_training_configuration.py
melisande-c, line 1: should move to the new `unit` directory?

src/careamics/config/ng_configs/ng_configuration.py
melisande-c, line 89: does `monitor` get propagated to `ReduceOnPlateau` somewhere else, I don't remember

tests/conftest.py
```diff
@@ -477,17 +480,25 @@ def _checkpoint_trainer(request):
...
-    else:
-        info_callback = None
-        callbacks = []
-    return Trainer(max_epochs=1, callbacks=callbacks), info_callback
+
+    def _get_trainer_and_info(
+        algorithm: str,
+    ) -> tuple[Trainer, CareamicsCheckpointInfo | None]:
+        if request.param:
+            info_callback = CareamicsCheckpointInfo(
```
melisande-c, line 488: hmm maybe there's a simpler way than a closure to make the checkpoint trainer depend on both the algorithm and whether to save with the checkpoint with `CareamicsCheckpointInfo`.

Not sure how though, so feel free to ignore this

melisande-c (approved):
Looks good, just some questions and minor things

src/careamics/config/ng_configs/ng_training_configuration.py
jdeschamps: It would make sense. But I would not know what to use as default, because it depends on the patch size. At least `epochs` is what people change and they can easily relate to it.

src/careamics/config/ng_configs/ng_training_configuration.py
jdeschamps: I just added a comment, in case. https://github.com/CAREamics/careamics/pull/817/commits/ddad49657aa5d8af8b475c475928bb84560753be

tests/unit/config/ng_configs/test_ng_training_configuration.py
jdeschamps: Yes indeed, done in https://github.com/CAREamics/careamics/pull/817/commits/ddad49657aa5d8af8b475c475928bb84560753be

src/careamics/config/ng_configs/ng_configuration.py
jdeschamps: So currently there are two `monitor`, the parameter in `N2VAlgorithm` and the one passed to the early stopping callback.

- `monitor`in N2V algorithm is passed in `N2VModule` to `configure_optimizers` ([see](https://github.com/CAREamics/careamics/blob/64cae22623126d1235b2e61ef061b0965750e8d6/src/careamics/lightning/dataset_ng/lightning_modules/module_utils.py#L156))
- `monitor` key in the presets in the NG training configuration is passed to the early stopping callback in `CAREamistV2`.

The convenience function for Noise2Void (not CARE or N2N) uses the same `monitor` to create the NG training config and the algorithm config. So any specified `monitor` will be passed to both early stopping and optimizers. For N2V, the present ensures checkpointing without monitoring val_loss.

For CARE/N2N, the convenience function does not expose that parameter, which is also not defined in the algorithm configs. The consequence for them is that:
- model checkpointing is done without monitoring of val loss for N2N, but with monitoring of val_loss for CARE
- early stopping is done on val_loss
- optimizers (including learning rate scheduler) are using val_loss

It is a bit of a mess, I have to think about it a little more.

tests/conftest.py
jdeschamps: To be honest, I did not think it through here. I had a hard time understanding these fixtures and what was going on. So I settled for the first solution that made sense to me and worked.

jdeschamps: My personal opinion is that the `checkpoint` fixture should be refactored to be clearer, but I am not willing to do that now... Just reading it again, I am again confused about what is happening there. :)

src/careamics/config/ng_configs/ng_training_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+    cannot be used to evaluate the best performing models.
+
+    This presets saves checkpoints every 10 epochs, as well as the last one.
+    """
+
+    every_n_epochs: int = 10
+    """Save a checkpoint every 10 epochs."""
+
+    save_top_k: int = -1
+    """Do not save the checkpoints based on a monitored value."""
```
melisande-c: wait I also just noticed that -1 means "all models are saved"

https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html#modelcheckpoint

src/careamics/config/ng_configs/ng_training_configuration.py
melisande-c: I think you mean it depends on the batch size, but I would argue every 10 epochs could also be a bad default, if someone has a lot of data they might not even train more than like 20-30 epochs.

Maybe we should expose this parameter in the n2v config convenience functions, to encourage people to change this parameter to fit how much data they have

tests/conftest.py
melisande-c: yeah sorry about that 🙈

melisande-c: > I am again confused about what is happening there. :)

You have it so that the fixture returns a function which captures the variable `"w_ckpt_info"`, or `"w/o_ckpt_info"` and takes the type of algorithm as an argument. This is so that it can depend on both values since fixtures can only be parametrised with one parameter

src/careamics/config/ng_configs/ng_configuration.py
melisande-c: I mean it might be feasible to want to monitor two different metrics for the optimisers and the checkpointing, for care maybe we'll want to add the option to use ssim for saving checkpoints, for example

src/careamics/config/ng_configs/ng_training_configuration.py
jdeschamps: What I meant was with patch size 128 or 64, the number of steps will also change. 

Given that the best checkpoint will probably always be the last one with N2V, it is also fine if only 2 checkpoints are saved. In addition, for most users an epoch is something relatable (even though they might not understand that it has no particular meaning if you set the number of steps). I don't really see what a bad default would be (except saving two often) in the case of N2V.

We could expose it, I am not sure it will be often used though. I am tempted to simply make a tutorial in the docs using https://github.com/CAREamics/careamics/pull/823 and a post configuration modification of `every_n_epochs`. And see how it ends up being used (in the facility, by users asking questions).

If that is you definite opinion that it should be on the steps, we can also make it such.

src/careamics/config/ng_configs/ng_configuration.py
jdeschamps: That is a good point, I'd open an issue to explore this.

src/careamics/config/ng_configs/ng_training_configuration.py
melisande-c: > Given that the best checkpoint will probably always be the last one with N2V

I'm not sure this is necessarily true, but we currently have no good way to tell which is the best checkpoint.

But yeah maybe a tutorial on how to modify the config is fine

src/careamics/config/ng_configs/ng_training_configuration.py
jdeschamps: > If save_top_k == -1, all models are saved. Please note that the monitors are checked every every_n_epochs epochs.

Yes, it needs to be set to `-1` so that we can save `every_n_epochs`.

melisande-c: ah I see, that's confusing

jdeschamps: The docstring description could be improved

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/817/changes/321fb7cb3550e10130e07752986561c31d536489

PR #818, opened 2026-03-12
Feat(Next-Gen): various fixes

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Various small fixes encountered when writing the docs.

### Allow disabling validation

- allow passing `n_val_patches=0` to disable validation
- throw error in `create_advanced_n2v_config` if `n_val_patches=0` and `monitor_metric="val_loss"`, since it will otherwise trigger an error later on

Note: having no validation makes no sense for CARE, so we could enforce this in the configuration.

https://github.com/CAREamics/careamics/commit/eaf87ffd3df6952bf440e863b2992aaa32914280
https://github.com/CAREamics/careamics/commit/29e3c6192c1c869d65af9372eebcd7445318938b

### Configuration version

- set `NGConfiguration` version to `0.2.0` (`0.1.0` will be the old configuration)

https://github.com/CAREamics/careamics/commit/336491eee5116f50c01799374a10355b1249e556
https://github.com/CAREamics/careamics/commit/c78c6b1002f8380d4ed02828e3182fc9dc10b7a9

### Avoid unclear error due to wrong `data_type`

- avoid getting an unspecific (non-Pydantic) error in `NGDataConfig` when `data_type` is wrong (e.g. "arra")

https://github.com/CAREamics/careamics/commit/216ff8c93c31fe8460a6847de8dd00bf3c80326d

### Improve error message in `get_write_func`

- improve the error message users would get when passing `czi` as `data_type` in prediction, which is not possible for lack of a write function

https://github.com/CAREamics/careamics/commit/8b92829bed4215cf3f0be7aed3e15def3f6c3e16

### Catch error early in CAREamistV2

- raise a specific error for trying to predict to disk to zarr without tiling

https://github.com/CAREamics/careamics/commit/b4b0674379ab3805303e719bf137adc1f3d772ad

### Rename normalization for coherence

- rename `minmax` to `min_max` since `mean_std` is written with an underscore
- rename `quantile` parameters to be plural, for coherence with the parameters in the other normalization methods

https://github.com/CAREamics/careamics/commit/cb462c654a660e0095cd74e2a45e8dd166899cfc
https://github.com/CAREamics/careamics/commit/8406607df3c546e7a9235caeaf42a60033383637
https://github.com/CAREamics/careamics/commit/5e029d0584d6b9bb18c0e7560aeb36f2d4b45147
https://github.com/CAREamics/careamics/commit/653c6206cb051db94a7a3222861727dac45773fe

### Add `AdamW`

- add support for `AdamW`, it will be needed for comparisons with another package for segmentation

https://github.com/CAREamics/careamics/commit/653c6206cb051db94a7a3222861727dac45773fe

### Remove unused parameter in `N2VManipuation`

- remove `remove_center` since we never changed that parameter

https://github.com/CAREamics/careamics/commit/6252eca1f0e2e53fcd2741cafed54e359e552134


src/careamics/config/data/ng_data_config.py
```diff
@@ -242,6 +242,9 @@ def axes_valid(cls, axes: str, info: ValidationInfo) -> str:
         ValueError
             If axes are not valid.
         """
+        if "data_type" not in info.data:
+            raise ValueError("`data_type` must be provided to validate axes.")
```
melisande-c: Maybe "Validation for `data_type` failed prior to axes validation"

melisande-c (approved):
Looks good

> Note: having no validation makes no sense for CARE, so we could enforce this in the configuration.

Remember that if validation data is passed separately the `n_val_patches` parameter is ignored and the number of validation patches is just calculated from the provided data. But I guess if someone is going out of their way to change `n_val_patches` from the default to zero they should be warned

src/careamics/config/data/ng_data_config.py
jdeschamps: Oh yes, much nicer.

jdeschamps:
> Looks good
> 
> > Note: having no validation makes no sense for CARE, so we could enforce this in the configuration.
> 
> Remember that if validation data is passed separately the `n_val_patches` parameter is ignored and the number of validation patches is just calculated from the provided data. But I guess if someone is going out of their way to change `n_val_patches` from the default to zero they should be warned

I am more afraid of people trying to actively not have validation because they misinterpreted the docs (no validation for N2V -> applied to CARE).

But maybe it is not that critical.

EDIT: the parameter is not available in the convenience function of CARE, so it's all good.

src/careamics/config/data/ng_data_config.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/818/commits/4b768c7d015ab9607c7e597165971af33c63345d

PR #822, opened 2026-03-13
Fix: add validator to cover mismatching channels in CARE/N2N

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Add validator to prevent `independent_channel=True` with mismatching input and output channel numbers.

The following snippet triggers an error in the UNet creation:

```python
   # create a configuration
    config = create_advanced_care_config(
        experiment_name="na",  # unused in CAREamistV2
        data_type="array",
        axes="CYX",
        patch_size=(64, 64),
        batch_size=2,
        num_epochs=1,
        num_workers=0,
        n_channels_in=3,
        n_channels_out=5,
        independent_channels=True,
        n_val_patches=1,
    )

    # instantiate CAREamistV2
    careamist = CAREamistV2(config=config, work_dir=tmp_dir)

    train_data = np.random.rand(3, 128, 128)  # (C, Y, X)
    target_data = np.random.rand(5, 128, 128)  # (C, Y, X)

    careamist.train(train_data=train_data, train_data_target=target_data)

-->    raise ValueError("out_channels must be divisible by groups")
ValueError: out_channels must be divisible by groups
```

This is because CARE and N2N allow different input/output channel numbers, but this is not compatible with independent channels. 

Further than that, CARE should probably default to non-independent channels by default, while N2N should be independent channels.

This PR adds a validator to cover that case and change the convenience functions defaults for CARE.


melisande-c:
Should N2N be allowed a different number of input and output channels?

jdeschamps:
> Should N2N be allowed a different number of input and output channels?

In most cases it will indeed be `n_in=n_out` (Noise2Noise, DeepCAD).

But during the DL4MIA course, we did some conclusive experiments with (N-1) to 1 frame in a Noise2Noise setting for calcium imaging. Unfortunately, we don't have the data. We should at some point come back to that and make an example.

melisande-c:
> > Should N2N be allowed a different number of input and output channels?
> 
> In most cases it will indeed be `n_in=n_out` (Noise2Noise, DeepCAD).
> 
> But during the DL4MIA course, we did some conclusive experiments with (N-1) to 1 frame in a Noise2Noise setting for calcium imaging. Unfortunately, we don't have the data. We should at some point come back to that and make an example.

Ah ok so that wouldn't be classed as CARE?

melisande-c (approved):

jdeschamps:
> > > Should N2N be allowed a different number of input and output channels?
> > 
> > 
> > In most cases it will indeed be `n_in=n_out` (Noise2Noise, DeepCAD).
> > But during the DL4MIA course, we did some conclusive experiments with (N-1) to 1 frame in a Noise2Noise setting for calcium imaging. Unfortunately, we don't have the data. We should at some point come back to that and make an example.
> 
> Ah ok so that wouldn't be classed as CARE?

Noisy to noisy I would still call Noise2Noise!

PR #806, opened 2026-03-06
Feat: Add `monitor_metric` config option for Noise2Void

Callidior:
## Disclaimer

- [ ] I am an AI agent.
- [x] I have used AI and I thoroughly reviewed every line.
- [ ] I have not used AI extensively.

## Description

> [!NOTE]  
> Add `monitor_metric` config option to customize the metric used by `ReduceLROnPlateau` LR scheduler.

### Background - why do we need this PR?

Enables training without validation runs.

### Overview - what changed?

- Added option `monitor_metric` to factory functions

## How has this been tested?

- Unit tests still pass
- Training runs on own data still successful and yielding expected results, both with monitoring `val_loss` and monitoring `train_loss_epoch`

## Additional Notes and Examples

---

**Please ensure your PR meets the following requirements:**

- [x] Code builds and passes tests locally, including doctests
- [ ] New tests have been added (for bug fixes/features)
- [x] Pre-commit passes
- [ ] PR to the documentation exists (for bug fixes / features)

on-behalf-of: @ZEISS <bjoern.barz@zeiss.com>

PR #823, opened 2026-03-13
Feat(Next-Gen CAReamics): show users available checkpoints

### LINKED ISSUE #305: Show users available checkpoints
## Problem

Once a model is trained, users have to inspect the checkpoints folder to know what has been saved. There is no indication of the loss or metrics corresponding to the validation set and the particular checkpoint.

The `CAREamist` should be able to print this information to users, it has access to the loss (and metrics in the future) for each epoch. Then it could also allow users to choose a checkpoint and load it, via maybe an `epoch_number`, rather than have them load manually via its path.

diyasrivas:
## Description

> [!NOTE]  
> **tldr**: 
Adds `get_checkpoints()` to `CAREamistV2` to list available checkpoints with their validation losses, and a `checkpoint` parameter to `predict` and `predict_to_disk` to run inference from a specific checkpoint.

Previously after training, users had to manually inspect the `checkpoints/` folder to know what had been saved, with no indication of the validation loss associated with each checkpoint. Loading a specific checkpoint required knowing the full file path. 

### Overview - what changed?

~~Two new public methods added to both `CAREamist` and `CAREamistV2`: `get_checkpoints()`, which scans the checkpoint directory and returns a structured summary of available checkpoints with their epoch numbers and validation losses, and `load_checkpoint(epoch)`, which loads the model weights from the checkpoint corresponding to a given epoch number. Then `_scan_checkpoints()` encapsulates the scanning logic to avoid redundant work and output when `load_checkpoint` is called after `get_checkpoints`.~~

`get_checkpoints()` method added to careamist_v2 which scans the checkpoint directory and returns a list of checkpoint filenames with a printed summary table of epoch numbers and validation losses. The `predict` and `predict_to_disk` methods have an optional `checkpoint` parameter that, when provided, passes the resolved checkpoint path to `trainer.predict` via `ckpt_path`, letting PyTorch Lightning handle device placement automatically. Checkpoint scanning logic is encapsulated in a new utility module `checkpoint_utils.py`.

Checkpoint filenames already encode epoch number and val_loss via the `ModelCheckpoint` naming pattern (`{experiment_name}_{epoch:02d}_step_{step}_{val_loss:.4f}.ckpt`), so no changes to checkpoint saving were needed. The epoch number is parsed by stripping the experiment name prefix from the filename stem. Val losses are retrieved from the existing CSV logger via `read_csv_logger` and joined to checkpoints via a private helper `_epoch_to_val_loss` in `lightning_utils.py`.

Checkpoint-based prediction passes the resolved path directly to `trainer.predict(ckpt_path=...)`, consistent with how `trainer.fit` handles checkpoint resumption in `train`. This avoids manual weight loading and ensures correct device placement.

Example usage after training:
```python
ckpts: list[str] = careamist.get_checkpoints()
#    Epoch      Val Loss  Name
# ----------------------------
#        0      1.013745  my_experiment_00_step_4_1.0137.ckpt
#        1      0.977344  my_experiment_01_step_8_0.9773.ckpt
#        2      1.100690  my_experiment_02_step_12_1.1007.ckpt
#     last           N/A  my_experiment_last.ckpt

preds = []
for ckpt in ckpts:
    pred, _ = careamist.predict(pred_data=data, checkpoint=ckpt)
    preds.append(pred)
```

## Changes made: 

~~updated `src/careamics/careamist.py`~~  
~~updated `src/careamics/careamist_v2.py`~~  
~~`src/careamics/utils/lightning_utils.py` added `_epoch_to_val_loss` and removed unnecessary `print(metrics)`~~

- `src/careamics/utils/checkpoint_utils.py` — new module containing `CheckpointInfo` TypedDict and `get_checkpoint_info` utility
- `src/careamics/careamist_v2.py` — added `get_checkpoints()`, `_resolve_ckpt_path()`, and `checkpoint` parameter to `predict` and `predict_to_disk`
- `src/careamics/utils/lightning_utils.py` — added `_epoch_to_val_loss` helper and removed stray `print(metrics)`
- `tests/test_careamist_v2_predict.py` — added `test_get_checkpoints`, `test_predict_with_checkpoint`, and `test_predict_with_checkpoint_gpu`

## How has this been tested?

~~Tested careamist with a 3-epoch N2V training run on random array data. Verified that `get_checkpoints()` correctly enumerates all saved checkpoints with the right epoch numbers and val_losses, that `load_checkpoint(epoch)` silently loads weights without re-printing the table, and that a `ValueError` with a clear message is raised when an invalid epoch is requested.~~

2 tests added to `tests/test_careamist_v2_predict.py`:

- `test_get_checkpoints` — verifies that after a 2-epoch training run, `get_checkpoints()` returns a non-empty list of `.ckpt` filename strings
- `test_predict_with_checkpoint` — verifies that `predict` produces correctly shaped output when called with each checkpoint returned by `get_checkpoints()`, directly testing the loop-over-checkpoints use case
- `test_predict_with_checkpoint_gpu` — verifies that checkpoint-based prediction runs on GPU and that model parameters remain on the correct device after inference; skipped automatically when CUDA is unavailable

## Related Issues

- Resolves #305 

---

- [x] Pre-commit passes


src/careamics/careamist.py
jdeschamps, line 1: Let's not add features to the old `CAREamist`, otherwise we will never be able to write the docs.

src/careamics/careamist_v2.py
```diff
@@ -798,3 +794,91 @@ def stop_training(self) -> None:
         """Stop the training loop."""
         self.trainer.should_stop = True
         self.trainer.limit_val_batches = 0  # skip validation
+
+    def _scan_checkpoints(self) -> list[dict]:
```
jdeschamps: This should be refactored somewhere else to not bloat the module.

src/careamics/careamist_v2.py
```diff
@@ -798,3 +794,91 @@ def stop_training(self) -> None:
...
+        )
+        losses = read_csv_logger(
+            self.config.get_safe_experiment_name(), self.work_dir / "csv_logs"
+        )
+        epoch_to_loss = _epoch_to_val_loss(losses)
+
+        checkpoints = []
+        for ckpt_path in sorted(checkpoint_dir.glob("*.ckpt")):
+            if ckpt_path.stem.endswith("_last"):
+                continue
```
jdeschamps: why is last not included?

src/careamics/careamist_v2.py
```diff
@@ -798,3 +794,91 @@ def stop_training(self) -> None:
...
+        checkpoints = self._scan_checkpoints()  
+        matches = [c for c in checkpoints if c["epoch"] == epoch]
+        if not matches:
+            available = [c["epoch"] for c in checkpoints]
+            raise ValueError(
+                f"No checkpoint found for epoch {epoch}. "
+                f"Available epochs: {available}."
+            )
+        checkpoint = torch.load(matches[0]["path"], map_location="cpu")
+        self.model.load_state_dict(checkpoint["state_dict"])
```
jdeschamps: Why using `map_location="cpu"`? Wouldn't that force subsequent prediction to be on cpu?

We have a `get_device` method somewhere, but I think it is safe to assume that it will should be automatically loaded on the same device it was saved from (since users would have just trained).

src/careamics/careamist_v2.py
```diff
@@ -798,3 +794,91 @@ def stop_training(self) -> None:
...
+                continue
+            checkpoints.append({
+                "epoch": epoch,
+                "val_loss": epoch_to_loss.get(epoch),
+                "path": ckpt_path,
+            })
+
+        return sorted(checkpoints, key=lambda x: x["epoch"])
+
+    def get_checkpoints(self) -> list[dict]:
```
jdeschamps: This should also be refactored somewhere else.

jdeschamps (changes requested):
Could you add tests? In particular that prediction can be performed after loading a checkpoint (the use case for that feature)? We should also tests on the VDIs that prediction is performed on GPU (I find the mapping parameter to CPU suspicious).

While the loading per epoch is neat, I feel it adds complexity. We could just as well list all the checkpoints and load from names.

In addition, it makes more sense to me to make the "checkpoint_name" be a parameter in the prediction functions. These could first load a checkpoint if it is requested and then predict as usual. That would have the advantage that it'd be easy for users to loop over the checkpoints:

```python
ckpts: list[str] = careamist.get_checkpoints()

preds = []
for ckpt in ckps:
    preds.append(careamist.predict(pred_data=data, checkpoint=ckpt))
```

src/careamics/careamist_v2.py
melisande-c: probably a good idea to use `self.model.load_from_checkpoint` instead of `load_state_dict`.

melisande-c: Also, the `CAREamistV2.checkpoint_path` attribute should be updated to reflect the current loaded checkpoint

jdeschamps: Does `self.model.load_from_checkpoint` return a new model or the same isntance with the new weights? I am wondering because it is a class method, isn't it?

If it returns a new instance, it may not be updated in the `Trainer`.

melisande-c: Ah that's true but if we use the `ckpt_path` argument in `trainer.predict` like we do with `trainer.fit` it will use the weights from that checkpoint. 

How we do it in `CAREamistV2.train`:
https://github.com/CAREamics/careamics/blob/cbe022092ca3705afcc1096b0563e032c99a824c/src/careamics/careamist_v2.py#L331-L333

lightning predict API docs: https://lightning.ai/docs/pytorch/stable/common/trainer.html#predict

jdeschamps: That would make perfect sense! Then no need to even do the loading ourselves.

src/careamics/careamist.py
diyasrivas: removed the initial changes i made to careamist

src/careamics/careamist_v2.py
diyasrivas: i refactored into a new module src/careamics/utils/checkpoint_utils.py with a CheckpointInfo TypedDict and get_checkpoint_info utility function. careamist_v2.py imports from there now

src/careamics/careamist_v2.py
diyasrivas: i included it now, it should appear at the bottom of the table with epoch shown as last

src/careamics/careamist_v2.py
diyasrivas: The checkpoint parameter is now passed directly to trainer.predict(ckpt_path=...), which should be consistent with how trainer.fit handles self.checkpoint_path

src/careamics/utils/checkpoint_utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+from .lightning_utils import _epoch_to_val_loss, read_csv_logger
+
+
+class CheckpointInfo(TypedDict):
+    """Structured information about a single checkpoint file."""
+
+    name: str
+    epoch: int | None
+    val_loss: float | None
```
jdeschamps: Actually this may be something else than the val loss. https://github.com/CAREamics/careamics/pull/817 introduced a default that makes N2V save checkpoints every 10 epochs, and not based on validation loss (there may not be any validation, see https://github.com/CAREamics/careamics/pull/806).

Maybe this should be renamed `monitored_val`?

src/careamics/utils/lightning_utils.py
```diff
@@ -69,3 +68,20 @@ def read_csv_logger(experiment_name: str, log_folder: Union[str, Path]) -> dict:
...
+    losses : dict
+        Dictionary as returned by `read_csv_logger`, containing keys
+        "val_epoch" and "val_loss".
+
+    Returns
+    -------
+    dict of int: float
+        Mapping from epoch number to validation loss.
+    """
+    return dict(zip(losses["val_epoch"], losses["val_loss"], strict=False))
```
jdeschamps: What happens in the absence of `val_loss`?

jdeschamps (changes requested):
Can you test what happens when using Noise2Void with the new checkpoint preset (https://github.com/CAREamics/careamics/pull/817) and no validation (https://github.com/CAREamics/careamics/pull/806)?

Tip to not use validation: 
https://github.com/CAREamics/careamics/blob/1aeb51977908bad3c8b5381ae7a85eb4b825d647/docs/configuration_lightning.py#L22-L35

diyasrivas:
I tested with the N2V preset and no validation (n_val_patches=0, monitor_metric="train_loss_epoch") with 3 epochs of training, here only the last checkpoint is saved. The N2V preset saves every 10 epochs so no epoch-based checkpoints appear within 3 epochs. get_checkpoints() handled it without crashing and just resturned the last checkpoint:

```python
checkpoints = careamist.get_checkpoints()
   Epoch      Val Loss  Name
----------------------------
    last           N/A  test_no_val_last.ckpt
After the monitored_val rename and the _epoch_to_val_loss fix, the output became:
python>>> checkpoints = careamist.get_checkpoints()
   Epoch   Monitored Val  Name
------------------------------
    last           N/A  test_no_val_last.ckpt
```
Added a test `test_get_checkpoints_no_validation` to tests/test_careamist_v2_predict.py covering this case.

src/careamics/utils/checkpoint_utils.py
diyasrivas: renamed it (in checkpoint_utils.py and careamist_v2.py)

src/careamics/utils/lightning_utils.py
diyasrivas: okay so the `try/except Exception` in `get_checkpoint_info` was silently catching any failure from `_epoch_to_val_loss` when `val_loss` was absent from the CSV log, which returned an empty mapping. I think this means that it didn't crash but the handling wasn't explicit.  Now `_epoch_to_val_loss` in `lightning_utils.py` explicitly checks for the presence of both first
```python
if "val_loss" not in losses or "val_epoch" not in losses:
    return {}
```

tests/test_careamist_v2_predict.py
```diff
@@ -298,3 +298,99 @@ def write_numpy(file_path: Path, img: NDArray, *args, **kwargs) -> None:
...
+    assert len(checkpoints) > 0
+
+    predicted, _ = careamist.predict(
+        pred_data=train_array,
+        data_type="array",
+        checkpoint=checkpoints[0],
+    )
+
+    assert predicted[0].shape == train_array.shape
+    assert next(careamist.model.parameters()).device.type == "cuda"
```
jdeschamps: What is the point of that tests? Was it a rogue addition by Claude? Tests are never running on GPU.

tests/test_careamist_v2_predict.py
```diff
@@ -298,3 +298,127 @@ def write_numpy(file_path: Path, img: NDArray, *args, **kwargs) -> None:
...
+    """Test get_checkpoints works when training without validation."""
+    train_array = random_array((32, 32), seed=42)
+
+    config = create_advanced_n2v_config(
+        experiment_name="test",
+        data_type="array",
+        axes="YX",
+        patch_size=(8, 8),
+        batch_size=2,
+        num_epochs=3,
```
jdeschamps, line 397: If this is testing only with a `last` checkpoint, then it is missing the case when more checkpoints are created.

You need to change the checkpoint parameters to set `every_n_epochs` to 1 or 2.

src/careamics/careamist_v2.py
```diff
@@ -611,6 +635,7 @@ def predict(
         channels: Sequence[int] | Literal["all"] | None = None,
         in_memory: bool | None = None,
         loading: Loading = None,
+        checkpoint: str | None = None,
```
jdeschamps: It should also accept `Path` as type

src/careamics/careamist_v2.py
```diff
@@ -658,6 +683,9 @@ def predict(
             Additional keyword arguments to be passed to the read function.
         extension_filter : str, default=""
             Filter for the file extension.
+        checkpoint : str, optional
+            Checkpoint filename as returned by `get_checkpoints()`. If None,
```
jdeschamps, line 756: I think it could also be a path to a checkpoint, or a checkpoint name as returned by `get_checkpoints`

jdeschamps: And `_resolve_ckpt_path` should just check whether it exists

src/careamics/careamist_v2.py
```diff
@@ -765,6 +797,7 @@ def predict_to_disk(
         write_extension: str | None = None,
         write_func: WriteFunc | None = None,
         write_func_kwargs: dict[str, Any] | None = None,
+        checkpoint: str | None = None,
```
jdeschamps: Same here about accepting `Path`

src/careamics/careamist_v2.py
```diff
@@ -988,3 +1024,33 @@ def stop_training(self) -> None:
...
+            self.work_dir / "csv_logs",
+        )
+
+        header = f"{'Epoch':>8}  {'Monitored Val':>14}  Name"
+        print(header)
+        print("-" * len(header))
+        for ckpt in info:
+            epoch_str = str(ckpt["epoch"]) if ckpt["epoch"] is not None else "last"
+            val_str = f"{ckpt['monitored_val']:.6f}" if ckpt["monitored_val"] is not None else "N/A"
+            print(f"{epoch_str:>8}  {val_str:>12}  {ckpt['name']}")
```
jdeschamps: I don't think we want to print anything. I'd remove that, the names returned by this method should be sufficient.

src/careamics/careamist_v2.py
```diff
@@ -988,3 +1024,33 @@ def stop_training(self) -> None:
...
+
+        Returns
+        -------
+        list of str
+            Checkpoint filenames, sorted by epoch number.
+        """
+        checkpoint_dir = (
+            self.work_dir / "checkpoints" / self.config.get_safe_experiment_name()
+        )
+        info = get_checkpoint_info(
```
jdeschamps: This could directly return just the name of the checkpoints.

tests/test_careamist_v2_predict.py
```diff
@@ -298,3 +298,127 @@ def write_numpy(file_path: Path, img: NDArray, *args, **kwargs) -> None:
...
+    )
+
+    careamist = CAREamistV2(config=config, work_dir=tmp_path)
+    careamist.train(train_data=train_array, val_data=val_array)
+
+    checkpoints = careamist.get_checkpoints()
+
+    assert len(checkpoints) > 0
+    assert all(isinstance(name, str) for name in checkpoints)
+    assert all(name.endswith(".ckpt") for name in checkpoints)
```
jdeschamps: Isn't there only one checkpoint in that case? Then the test is not testing the right thing, it may be only returning a single one.

src/careamics/utils/checkpoint_utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+from .lightning_utils import _epoch_to_val_loss, read_csv_logger
+
+
+class CheckpointInfo(TypedDict):
+    """Structured information about a single checkpoint file."""
+
+    name: str
+    epoch: int | None
+    monitored_val: float | None
```
jdeschamps: `monitored_val` is ambiguous because of validation.

src/careamics/utils/checkpoint_utils.py
jdeschamps, line 1: Honestly, I'd just return the name of the checkpoints and that's it. They have epochs and potentially val loss in the name, do they not?

jdeschamps: In addition, we would need unit tests for these utility functions (mocking files on disk).

jdeschamps (changes requested):

src/careamics/careamist_v2.py
```diff
@@ -1058,3 +1096,29 @@ def stop_training(self) -> None:
         """Stop the training loop."""
         self.trainer.should_stop = True
         self.trainer.limit_val_batches = 0  # skip validation
+
+    def get_checkpoints(self) -> list[str]:
```
jdeschamps, line 1100: What happens if the training was done with `max_epochs=-1`? (or `num_epochs=-1` in our convenience functions?

src/careamics/utils/checkpoint_utils.py
diyasrivas: I deleted checkpoint_utils.py + the CheckpointInfo TypedDict, get_checkpoint_info, and the _epoch_to_val_loss helper in lightning_utils.py. Now get_checkpoints() now does glob("*.ckpt") on the checkpoint directory, separating epoch-based checkpoints from _last and returning their filenames as strings

src/careamics/careamist_v2.py
diyasrivas: yeah youre right, I've removed it. get_checkpoints() now should only return the list of filenames

tests/test_careamist_v2_predict.py
diyasrivas: sorry mistake! didn;t check through the tests properly to understand them before. Removed

tests/test_careamist_v2_predict.py
diyasrivas: removed

tests/test_careamist_v2_predict.py
diyasrivas: for test_get_checkpoints & test_get_checkpoints_no_validation, config.training_config.checkpoint_params["every_n_epochs"] = 1 is set after config creation overriding the N2V preset. Used len(checkpoints) >= 2

src/careamics/careamist_v2.py
diyasrivas: updated to accept Path

src/careamics/utils/checkpoint_utils.py
diyasrivas: removed this file

src/careamics/careamist_v2.py
diyasrivas: When I run:
```python
    experiment_name="test_inf",
    data_type="array",
    axes="YX",
    patch_size=(8, 8),
    batch_size=2,
    num_epochs=-1,
    roi_size=5,
    masked_pixel_percentage=5,
)
```
It runs indefinitely... Then I tried with early stopping:
```python
    experiment_name="test_inf",
    data_type="array",
    axes="YX",
    patch_size=(8, 8),
    batch_size=2,
    num_epochs=-1,
    roi_size=5,
    masked_pixel_percentage=5,
)
config.training_config.checkpoint_params["every_n_epochs"] = 1
config.training_config.early_stopping_params = {
    "monitor": "val_loss",
    "patience": 2,
    "mode": "min",
}
careamist.train(train_data=train_array, val_data=val_array)
print(careamist.get_checkpoints())
# ['test_inf_00_step_8_0.9098.ckpt', 'test_inf_01_step_16_0.9857.ckpt',
#  'test_inf_02_step_24_0.9422.ckpt', 'test_inf_last.ckpt']
```
With early stopping, get_checkpoints() returns the checkpoints saved during training

jdeschamps:
There is too much divergence with `v0.2`, so I have adapted your contribution in https://github.com/CAREamics/careamics/pull/918

PR #824, opened 2026-03-15
Feat(Next-Gen): Odd image size check during prediction

### LINKED ISSUE #741: Next-Gen CAREamics: fail early if odd image sizes and no tiling during prediction
## Problem

CatEek:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: Fail early during prediction when image spatial dimensions are not a power of 2 and tiling is not used, with a clear error telling users to use tiling.

Previously, if users ran `predict()` or `predict_to_disk()` on oddly shaped images and did **not** use tiling, the error could surface only inside the model, with an vague message. 

### Overview - what changed?

- **Dataset validation**: In prediction mode, when the patching strategy is not tiled, the dataset now checks that every image's spatial dimensions are a power of 2. If not, it raises a `ValueError` before the model runs.
- **Tests**: New tests ensure (1) `predict` and `predict_to_disk` raise this error for non–power-of-2 shapes without tiling, and (2) prediction without tiling still succeeds when shapes are power-of-2 (no regression).

- In `CareamicsDataset.__init__` (`dataset.py`), after the existing training shape check, an `else` branch for `data_config.mode == Mode.PREDICTING` was added. When `patching` is not `TiledPatchingConfig`, it checks that all `data_shapes` (from `input_extractor.image_stacks`) have spatial dimensions (indices 2+ in SC(Z)YX) that are positive powers of 2 via a small helper `_all_spatial_dims_power_of_2`. If any dimension fails, it raises a `ValueError` with message: *"Image spatial dimensions must all be a power of 2 when predicting without tiling. Please use tiling by passing `tile_size` and `tile_overlap` to the predict method."*

### New features or files

- `_all_spatial_dims_power_of_2()` added in `src/careamics/dataset_ng/dataset.py`.

- `CareamicsDataset.__init__` in `src/careamics/dataset_ng/dataset.py`:  new prediction-mode check.
- `tests/test_careamist_v2_predict.py`: three new tests (`test_predict_power_of_2_no_tiling_succeeds`, `test_predict_non_power_of_2_no_tiling_raises`, `test_predict_to_disk_non_power_of_2_no_tiling_raises`).

## How has this been tested?

- `test_predict_non_power_of_2_no_tiling_raises`: `predict(pred_data=array_30x30)` without tiling → `ValueError` with message containing `"tiling"`.
- `test_predict_to_disk_non_power_of_2_no_tiling_raises`: `predict_to_disk(pred_data=dir_with_30x30_tiff)` without tiling → same `ValueError`.

## Related Issues

- Resolves #741 

---

- [x] Pre-commit passes


CatEek:
The only thing is, I'm not sure pow2 is ok for Z dim

jdeschamps:
> The only thing is, I'm not sure pow2 is ok for Z dim

We can think of a way to automatically extend Z to the nearest acceptable size, the image stack will automatically add zero boundaries if we ask for a larger than image size.

In the case of tiling, the same mechanism applies, but we probably need to tell users (via for instance raising an error).

src/careamics/dataset_ng/dataset.py
```diff
@@ -135,6 +140,27 @@ def _patch_size_within_data_shapes(
...
+        everything after the first two (sample and channel) dimensions.
+
+    Returns
+    -------
+    bool
+        True if every spatial dimension of every image is a power of 2.
+    """
+    for data_shape in data_shapes:
+        for dim in data_shape[2:]:
+            if dim <= 0 or (dim & (dim - 1)) != 0:
```
jdeschamps: Why would dim be negative?

src/careamics/dataset_ng/dataset.py
```diff
@@ -159,6 +185,14 @@ def __init__(
                     "Not all images sizes are greater or equal than the patch size for "
                     "training and validation."
                 )
+        else:
+            if not isinstance(data_config.patching, TiledPatchingConfig):
+                if not _all_spatial_dims_power_of_2(data_shapes):
```
jdeschamps: When we discussed it during the internal meeting, the reason I was pushing for this solution was to avoid a hack-around accessing algorithm configuration to extract the model `depth`. Yet, power of 2 is indeed very restrictive compared to the reality of the constraint (multiple of `2**depth`).

I am starting top change my mind about this... Also because here we would be adding to the dataset a constraint that is not dependent on the dataset (e.g. what if people are using our dataset with a different model, or when we will be using this dataset with a LVAE, whose constraints I am currently not aware of), and that is a maintenance issue.

Alternative solutions I can think of:

1. As in careamics-ui: catch `Sizes of tensors must match except` error in `CAREamistV2`. problem is that we don't know which data is the problem.
2. Create a model constraints class that can be instantiated in `CAREamistV2` or by users in the LightningAPI, and passed to the `CAREamicsDataset`. I quite like this idea, as it would fit the modular system of the NGDataset, and would allow to pass constraints from other model by writing other instances of that same protocol.

``` python
from typing import Protocol

class ModelConstraints(Protocol):
    def is_shape_compatible(shape: Sequence[int]) -> bool:
        ...

    def get_constraints_messages() -> str:
        """Returns a concise error message indicating what the constraints of the model are."""
        ...
```

This model constraint could be automatically generated by a configuration object. And if people don't use it in the Lightning, that's fine, there is just no check.

3. Any way to access the aglorithm config? This would face the same issue outlined above, too tailored to a single architecture, and would lead to `if-else` structures.

melisande-c:
> > The only thing is, I'm not sure pow2 is ok for Z dim
> 
> We can think of a way to automatically extend Z to the nearest acceptable size, the image stack will automatically add zero boundaries if we ask for a larger than image size.
> 
> In the case of tiling, the same mechanism applies, but we probably need to tell users (via for instance raising an error).

Yes for prediction tile size can be greater than the image dimension. But for training we already enforce through the config that every patch dimension must be a power of two and greater than 8.

Btw, this is the function used for validation, maybe we can reuse it here:
https://github.com/CAREamics/careamics/blob/cbe022092ca3705afcc1096b0563e032c99a824c/src/careamics/config/validators/patch_validators.py#L10-L14

src/careamics/dataset_ng/dataset.py
CatEek: Why don't we go with even numbers or multiple of 2**depth here and spin off a new issue? Or even just put the explicit error closer to model class's forward ?

melisande-c: yeah maybe a more explicit error in the model's forward is easier

src/careamics/dataset_ng/dataset.py
```diff
@@ -137,6 +142,32 @@ def _patch_size_within_data_shapes(
     return all(smaller_than_shapes)
 
 
+def _check_all_spatial_dims(data_shapes: Sequence[Sequence[int]]) -> bool:
+    """Return True if all spatial dims comply with the requirements.
+
+    YX dimensions must be even and be greate or equal than 16. Z dimenstion must be
+    even and be greater or equal than 4.
```
jdeschamps:
```suggestion
    YX dimensions must be even and greater or equal than 16. Z dimension must be
    even and greater or equal than 4.
```

src/careamics/dataset_ng/dataset.py
```diff
@@ -191,6 +222,14 @@ def __init__(
                     "Not all images sizes are greater or equal than the patch size for "
                     "training and validation."
                 )
+        else:
+            if not isinstance(data_config.patching, TiledPatchingConfig):
+                if not _check_all_spatial_dims(data_shapes):
+                    raise ValueError(
+                        "Image spatial dimensions must all be a power of 2 when "
```
jdeschamps: Now the error says power of 2, but the function checks even and minimum size.

tests/test_careamist_v2_predict.py
```diff
@@ -298,3 +298,47 @@ def write_numpy(file_path: Path, img: NDArray, *args, **kwargs) -> None:
...
+    )
+
+    image_dir = tmp_path / "images"
+    image_dir.mkdir()
+    tifffile.imwrite(image_dir / "image.tiff", random_array((15, 30)))
+
+    careamist = CAREamistV2(config=config, work_dir=tmp_path)
+
+    with pytest.raises(ValueError, match="tiling"):
+        careamist.predict_to_disk(pred_data=image_dir)
```
jdeschamps: I don't see why these two tests could not be one.

```suggestion
def test_predict_invalid_spatial_dims_no_tiling_raises(tmp_path: Path):
    """Test that predict raises ValueError for invalid spatial dims without tiling."""
    config = create_advanced_n2v_config(
        experiment_name="test",
        data_type="array",
        axes="YX",
        patch_size=(8, 8),
        batch_size=1,
        num_epochs=1,
        roi_size=5,
        masked_pixel_percentage=5,
    )

    careamist = CAREamistV2(config=config, work_dir=tmp_path)

    pred_array = random_array((15, 30))

    with pytest.raises(ValueError, match="tiling"):
        careamist.predict(pred_data=pred_array)

    with pytest.raises(ValueError, match="tiling"):
        careamist.predict_to_disk(pred_data=pred_array)
```

src/careamics/dataset_ng/dataset.py
jdeschamps: Explicit error in the model sounds ok for now to me!

jdeschamps (changes requested):

jdeschamps (approved):

jdeschamps (changes requested):
I think it is important to be providing an exact validation. I will push an alternative PR.

jdeschamps:
Superseded by https://github.com/CAREamics/careamics/pull/869

PR #826, opened 2026-03-15
Feat(next-Gen): default num workers equal to cpu cores

CatEek:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: Default dataloader `num_workers` now auto-detects CPUs.

### Overview - what changed?

- Allow `num_workers=-1` in NG data/N2V factories to auto-resolve to CPU count while preserving explicit `num_workers >= 0`.

## How has this been tested?

- relevant methods in `TestNGDataConfiguration`


src/careamics/config/ng_factories/n2v_factory.py
```diff
@@ -329,8 +329,9 @@ def create_advanced_n2v_config(
         Axis along which to apply structN2V mask.
     struct_n2v_span : int, default=5
         Span of the structN2V mask.
-    num_workers : int, default=0
-        Number of workers for data loading. Unless explicitly overridden in
+    num_workers : int, default=-1
+        Number of workers for data loading. Use ``-1`` to automatically choose based
```
jdeschamps:
```suggestion
        Number of workers for data loading. Use `-1` to automatically choose based
```

to stick with the way the other docs are written.

tests/config/ng_factories/test_data_factory.py
```diff
@@ -86,3 +87,32 @@ def test_default_patching(self):
...
+            num_workers=4,
+        )
+
+        assert config.train_dataloader_params["num_workers"] == 4
+        assert config.val_dataloader_params["num_workers"] == 4
+        assert config.pred_dataloader_params["num_workers"] == 4
+
+    def test_num_workers_auto_detect(self):
+        """Test that num_workers=-1 resolves to get_default_num_workers() value."""
+        expected = get_default_num_workers()
```
jdeschamps: I would mock the os cpu count and test `get_default_num_workers`, rather than in the data factory.

jdeschamps (commented):
Have you tested it on local MPS mac and on the Windows VDI? We need to check whether it actually does the job in the context of Jupyter notebooks and python scripts.

In particular with Windows, we may need to see if it fails without `if __init__` clause (see [Windows multiprocessing tips](https://docs.pytorch.org/docs/stable/notes/windows.html#multiprocessing-error-without-if-clause-protection)). Otherwise we may need to set it to `0` by default for Windows.

mese79:
As we discussed, just wanted to mention maybe a more *safe* number for `num_workers` might be `num_cpu - 1`.  
This might become useful especially when using `careamics` in a notebook on a usual laptop/PC while other stuff are running as well. But I guess it won't be the case for training on HPC.

veegalinova:
After some discussion and testing, we decided this: 
- Linux: `num_workers` defaults to either `cpu_count - 1`, but at max 4 workers total
- Windows, MacOS, testing: 0 workers
- `persistent_workers` on in case `num_workers` > 0 (avoids re-initializing workers between epochs).

Reasoning for these choices:
_(note - everything was tested on BSD68, so something might not hold for larger datasets!)_
- MacOS - `mps` works well with several workers, but because of the initialization method (`spawn` vs `fork` in linux), initialization takes a lot of time (I observe ~1 min hang at the start of the training for 4 workers). Performance gain for multi-workers is quite small in comparison, so I decided that start-up time is more important
- Linux - `cpu_count` of workers is good when headless, but if training in napari we need one core free. We decided to limit the workers with 4 max cause there is limited performance gain after (again, this can be re-considered after testing with more complicated datasets and on more performant machines)

tests/conftest.py
```diff
@@ -536,6 +536,9 @@ def checkpoint(
             masked_pixel_percentage=0.4,
             normalization="mean_std",
             normalization_params={"input_means": [0.5], "input_stds": [0.3]},
+            num_workers=0,
+            num_steps=1,
+            n_val_patches=1,
```
jdeschamps: `n_val_patches` is not used if there is `val_data`, which seems to be passed on line 543.

tests/config/ng_factories/test_data_factory.py
```diff
@@ -86,3 +90,50 @@ def test_default_patching(self):
...
+
+        assert config.train_dataloader_params["num_workers"] == 4
+        assert config.val_dataloader_params["num_workers"] == 4
+        assert config.pred_dataloader_params["num_workers"] == 4
+
+    def test_get_default_num_workers_in_pytest(self):
+        """Test that get_default_num_workers returns 0 when running under pytest."""
+        assert get_default_num_workers() == 0
+
+    def test_get_default_num_workers_linux(self, monkeypatch: pytest.MonkeyPatch):
```
jdeschamps: All three platform tests could be joined into a single test with parametrization `platform, n_workers`.

tests/unit/config/ng_factories/test_data_factory.py
jdeschamps, line 1: We should take this opportunity to examine all data factory tests and move the unit tests to `tests/unit/`

src/careamics/config/ng_factories/n2v_factory.py
jdeschamps: Insisting on this comment, we should be consistent in how we represent variables/values in our docstrings.

jdeschamps (approved):
Can you add more comments to `get_default_num_workers` to explain the choices (the same way you justified them here)?

So that in the future we don;t have to wonder again.

tests/conftest.py
veegalinova: removed

tests/config/ng_factories/test_data_factory.py
veegalinova: done

tests/unit/config/ng_factories/test_data_factory.py
veegalinova: Good call. All of them are unit tests, I moved them there, and also moved the `get_default_num_workers` tests to `test_ng_data_config.py` as it is now part of the configuration and not `data_factory`.

PR #832, opened 2026-03-19
Feature(Next-Gen): Patch filtering rework

### LINKED ISSUE #741: Next-Gen CAREamics: fail early if odd image sizes and no tiling during prediction

melisande-c:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: This makes the patch filtering in the NG-Dataset more efficient by using the `StratifiedPatchingStrategy` to set different sampling regions in the data to have a reduced probability of being sampled per epoch.

The old patch filtering implementation used the `RandomPatchingStrategy` to keep resampling patches until the patch has signal. If there is a large amount of background in the data this is inefficient. In the new implementation the filtering process is happens once at the start of training and there is no need to define a "patience".

### Overview - what changed?

Quite a lot...

- The attributes `patch_filter` and `coord_filter` have been removed from the `CAREamicsDataset`, and it is also no longer initialised with masks. This also means all functionality for filtering within the dataset has been removed.
- Instead, the filtering process happens in `/ng_dataset/filter_bg.py`, there are two functions, one for filtering with a filter, one for filtering with a mask.
- The `create_train_val_datasets` and `create_val_split_dataset` factory functions have been refactored slightly to reduced code duplication. The `train_dataset` is first created with the new `create_trian_dataset` within which the data is filtered.
- The `StratifiedPatchingStrategy` has been modified to have a new method to set the probability that a sampling region will be sampled from during an epoch. This also necessitates that the number of patches are reduced if the probability of any regions is reduced.
- `filtered_patch_prob` and `filter_ref_channel` has been added to the ng data config.
  - The ref channel was added for the case of multiple channels.

- The idea is a filtering function can be applied to each sampling region in the `StratifiedPatchingStrategy`, if the region doesn't pass some threshold the probability of it being sampled from in an epoch can be reduced.
- The patching strategy instance only needs to store the probability of each region which doesn't take too much memory.
- When calculating the total probabilities and the sampling bins, the adjusted areas of the regions can be used, this is the area multiplied by the new probability.

### New features or files

- `src/careamics/dataset_ng/filter_bg.py`
  - `filter_background`
  - `filter_background_with_mask`

- `src/careamics/dataset_ng/factory.py`
- src/careamics/dataset_ng/patching_strategies/stratified_patching.py

### Removed features or files

- filtering from `CAREamicsDataset`

## How has this been tested?

- Added unit tests that test the number of patches are reduced in the `StratifiedPatchingStrategy` if the method `set_region_probs` is called.

- Added functional tests for filtering background, from a few different layers:
  - The patching strategy layer
  - The `filter_background` and `filter_background_with_mask` layer
  - The `create_train_dataset` layer

Still missing is a test for the lightning module layer or an e2e test with filtering.

## Related Issues

- Resolves #741 

## Breaking changes

- Data config has changed, `filter_patience` has been removed
- Passing masks to the dataset no longer happens

### Still TODO:

- Clean up patch filtering configs and filtering classes, we no longer need filtering patience or the probability.
- I am considering removing the option to have different "coord filters" (we don't have any others) because a mask filter config is required to pass masks to `create_train_dataset`.

---

- [x] Pre-commit passes


src/careamics/config/data/ng_data_config.py
```diff
@@ -305,9 +305,15 @@ class NGDataConfig(BaseModel):
...
     mode is `training`."""
 
-    patch_filter_patience: int = Field(default=5, ge=1)
-    """Number of consecutive patches not passing the filter before accepting the next
-    patch."""
+    filtered_patch_prob: float = Field(0.1, ge=0.0, le=1.0)
+    """The probability that each patch classed as background will be selected each epoch
+    during training. Patches can be classes as background by either using a
+    `patch_filter` or by supplying a mask during training. If neither is chosen this
+    parameter is ignored."""
```
jdeschamps:
```suggestion
    """The probability that each patch considered background will be selected each epoch
    during training. Patches can be considered background by either using a
    `patch_filter` or by supplying a mask during training. If neither is chosen this
    parameter is ignored."""
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -160,6 +160,33 @@ def exclude_patches(
             self.cumulative_image_samples,
         ) = self._calc_bins()
 
+    def set_region_probs(
+        self, data_idx: int, sample_idx: int, probs: dict[tuple[int, ...], float]
+    ) -> None:
+        """
+        Set the probability that regions will be sampled from, each epoch.
```
```suggestion
        Set the probability that regions will be sampled from at each epoch.
```

src/careamics/dataset_ng/patching_strategies/stratified_patching.py
```diff
@@ -506,6 +553,23 @@ def get_included_grid_coords(self) -> list[tuple[int, ...]]:
         grid_coords_all: set[tuple[int, ...]] = set(self.grid_coords.keys())
         return list(grid_coords_all.difference(self.excluded_patches))
 
+    def set_region_probs(self, probs: dict[tuple[int, ...], float]) -> None:
+        """
+        Set the probability that regions will be sampled from, each epoch.
```
```suggestion
        Set the probability that regions will be sampled from at each epoch.
```

src/careamics/dataset_ng/factory.py
```diff
@@ -270,6 +263,109 @@ def select_image_stack_loader(
...
+            patch_filter,
+            config.filter_ref_channel,
+            config.filtered_patch_prob,
+        )
+    # if both masks and patch_filter are present they are both applied
+    if mask_extractor is not None and config.coord_filter is not None:
+        if not isinstance(patching_strategy, StratifiedPatchingStrategy):
+            raise TypeError(
+                "Mask filtering is only compatible with stratified "
+                f"patching. Found {config.patching.name} patching in the configuration."
```
jdeschamps, line 347:
```suggestion
                f"patching. Found {config.coord_filter.name} patching in the configuration."
```

src/careamics/dataset_ng/filter_bg.py
```diff
@@ -0,0 +1,115 @@
...
+    Parameters
+    ----------
+    patching_strategy : StratifiedPatchingStrategy
+        A stratified patching strategy to filter.
+    input_extractor : PatchExtractor
+        A patch extractor holding the data to filtering.
+    patch_filter : PatchFilterProtocol
+        The function to filter patches with.
+    ref_channel : int
+        Which channel to filter the data with.
```
jdeschamps:
```suggestion
        On which channel to filter the data.
```

jdeschamps (commented):
I have not gone through the tests, but here is just meaningless comments on docs, and one error to correct I think.

src/careamics/dataset_ng/factory.py
melisande-c: No this was correct before because filtering is only compatible with stratified patching, this error is telling the user that a different patching strategy was found, `"random"` for example

jdeschamps (approved):

PR #827, opened 2026-03-15
Chore(Next-Gen): fix docstring for Next-Gen files

CatEek:
## Disclaimer

- [ ] I am an AI agent.
- [x] I have used AI and I thoroughly reviewed every line.
- [ ] I have not used AI extensively.

## Description

> [!NOTE]  
> **tldr**: Fix numpydoc and ruff issues in `dataset_ng` so pre-commit passes.

### Background - why do we need this PR?

Pre-commit was failing

### Overview - what changed?

Docstrings across `dataset_ng` now document parameters and returns where missing.

### Implementation - how did you implement the changes?

Added minimal Parameters/Returns blocks and descriptions to satisfy numpydoc,  reordered sections,  shortened overlong docstrings.

### Modified features or files

- `dataset_ng` folder

## Related Issues

- Resolves #718 

---

**Please ensure your PR meets the following requirements:**

- [x] Code builds and passes tests locally, including doctests
- [ ] New tests have been added (for bug fixes/features)
- [x] Pre-commit passes
- [ ] PR to the documentation exists (for bug fixes / features)

PR #833, opened 2026-03-20
Chore(Next-Gen): enable pre-commits for `CAREamistV2`

jdeschamps:
## Description

Following https://github.com/CAREamics/careamics/pull/827, this PR enables pre-commit hooks for `CAREamistV2` and attempts to solve the errors.

Note that there is something fishy with the use of `TypeVar` that `mypy` disagrees with:

```python
src/careamics/careamist_v2.py:128: error: Argument 1 to "_load_model" of "CAREamistV2" has incompatible type "NGConfiguration[CAREAlgorithm] | Path | None"; expected "NGConfiguration[AlgorithmConfig] | Path | None"  [arg-type]
src/careamics/careamist_v2.py:128: error: Argument 1 to "_load_model" of "CAREamistV2" has incompatible type "NGConfiguration[N2NAlgorithm] | Path | None"; expected "NGConfiguration[AlgorithmConfig] | Path | None"  [arg-type]
src/careamics/careamist_v2.py:128: error: Argument 1 to "_load_model" of "CAREamistV2" has incompatible type "NGConfiguration[N2VAlgorithm] | Path | None"; expected "NGConfiguration[AlgorithmConfig] | Path | None"  [arg-type]
Found 3 errors in 1 file (checked 1 source file)
```

One way to solve it to redefine `AlgorithmConfig` in the same module, rather than importing it. It seems that allows "bounding" the type to the functions it is used in.

 But it would be annoying to have to redefine the generics in every module. I decided to just ignore it for now for two reasons:
- https://github.com/CAREamics/careamics/pull/817 introduces a validator that may improve the situation
- Maybe pyright (https://github.com/CAREamics/careamics/pull/698) will be better suited


melisande-c:
hmm yeah I think maybe this is one of those things pyright is fine with but mypy complains, and now I don't remember why I made the configuration a generic, sorry for complicating things. But a solution here is to replace the typing of the config arg in the `__init__` to this:

```python
NGConfigurationType = (
    NGConfiguration[CAREAlgorithm]
    | NGConfiguration[N2NAlgorithm]
    | NGConfiguration[N2VAlgorithm]
)
```

jdeschamps:
> hmm yeah I think maybe this is one of those things pyright is fine with but mypy complains, and now I don't remember why I made the configuration a generic, sorry for complicating things. But a solution here is to replace the typing of the config arg in the `__init__` to this:
> 
> ```python
> NGConfigurationType = (
>     NGConfiguration[CAREAlgorithm]
>     | NGConfiguration[N2NAlgorithm]
>     | NGConfiguration[N2VAlgorithm]
> )
> ```

No need to apologize. It is a bit annoying that generics are so annoying in Python (at least for mypy). Maybe another nail in the coffin.

Implemented the change in https://github.com/CAREamics/careamics/pull/833/changes/68b39be5a770e1cd48ce9c49dca7be78d6974c3f

src/careamics/careamist_v2.py
```diff
@@ -495,21 +511,25 @@ def train(
...
-        datamodule = CareamicsDataModule(
+        datamodule = CareamicsDataModule( # type: ignore
             data_config=self.config.data_config,
             train_data=train_data,
             val_data=val_data,
             train_data_target=train_data_target,
             val_data_target=val_data_target,
             train_data_mask=filtering_mask,
-            loading=loading,
+            loading=loading, # type: ignore
```
melisande-c, line 532: damn it something else that works with pyright but not mypy, this is happening bec mypy can't resolve overloads that well

melisande-c (approved):
Looks good

src/careamics/careamist_v2.py
jdeschamps: Yep... Moving to pyright is just going to feel good I surmise.

PR #859, opened 2026-03-27
Feat(Next-Gen CAREamist): disable early stopping for Noise2Void and Noise2Noise

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: disable early stopping for N2V and N2N.

The following code stops after a few epochs:

```python
from careamics.careamist_v2 import CAREamistV2
from careamics.config.ng_factories import create_advanced_n2v_config
import numpy as np

data = np.random.rand(512, 512)  # dummy data

# create a configuration
    experiment_name="the_n2v_training_test",
    data_type="array",
    axes="YX",
    patch_size=[64, 64],
    batch_size=8,
    num_epochs=50,
)

# instantiate a careamist
careamist = CAREamistV2(config, work_dir="residuals")

# train the model
careamist.train(train_data=[data], val_data=[data])
```

The reason is that the early stopping kicks in and stops training.

This PR does the following:
- Use the default `NGTrainingConfig` constructor to either make a default early stopping parameters dict (CARE) or set it to `None`.
- Add `early_stopping_params` to `create_advance_care_config`, so that users can set it however they wish. Note that `monitor_metric` is only a parameter for N2V, so it does not appear in `create_advance_care_config`

- [x] Pre-commit passes


docs/configuration.md
```diff
@@ -689,6 +689,28 @@ If you want to overried the CAREamics defaults, set `checkpoint_params`.
     --8<-- "configuration_lightning.py:adv_config_care_checkpoint"
     ```
 
+!!! note "Default checkpointing behavior"
+
+    By default, CARE saves the top 3 best models according to the validation loss, and the last model at the end of training.
+
+    Noise2Void, on the other hand, saves a checkpoint every 10 epochs,
+    as well as the last one. The reaosn behind this choice is that
```
melisande-c: typo
```suggestion
    as well as the last one. The reason behind this choice is that
```

melisande-c (approved):
looks good

PR #861, opened 2026-03-27
Chore(Next-Gen CAREmics): improve wrong input type error message

jdeschamps:
## Description

I find the current error message raised by the data module input validation difficult to read:
- It contains a lot of info that may not be relevant to the case
- It does not cover lists (as in the error message may show `input type <list>` rather than what is in the list)
- It states something about validation, but it is not the source of the error in that case

https://github.com/CAREamics/careamics/blob/9e911f201d1939cb7369c03be72841f17236fd7e/src/careamics/lightning/dataset_ng/data_module_utils.py#L431-L453

```python
ValueError: Invalid argument combination for data initialization. data_type=SupportedData.TIFF, input type is <class 'list'>. For custom data, you must provide either a ReadFuncLoading or ImageStackLoading dataclass as instruction on how to load the data. If a training target is provided, a validation target must also be provided, unless automatic validation splitting is being used.
```

This PR improves the error message by:
- checking whether the input is a sequence (and not a string), then checking the type inside the sequence
- making the second part of the error message specific to the data type

Examples:
```python
>>> initialize_data_pair(data_type="zarr", input_data=np.array([1, 2, 3]))
ValueError: Invalid argument combination for data initialization. The data type is set to `zarr`, but received non-compatible input type `<class 'numpy.ndarray'>`. For `zarr` data type, the expected input is a str or a pathlib.Path (see documentation), or a sequence of either. Refer to the documentation for more details.

>>> initialize_data_pair(data_type="tiff", input_data = [ np.array([1, 2, 3]) ])
ValueError: Invalid argument combination for data initialization. The data type is set to `tiff`, but received non-compatible input type `<class 'numpy.ndarray'>`. For `tiff` data type, the expected input is a str, a pathlib.Path, or a sequence of either. Refer to the documentation for more details.

>>> initialize_data_pair(data_type="array", input_data=[""])
ValueError: Invalid argument combination for data initialization. The data type is set to `array`, but received non-compatible input type `<class 'str'>`. For `array` data type, the expected input is a numpy.ndarray or a sequence of numpy.ndarray. Refer to the documentation for more details.

>>> initialize_data_pair(data_type="custom", input_data=["data/test.tiff"])
ValueError: Invalid argument combination for data initialization. The data type is set to `custom`, but received non-compatible input type `<class 'str'>`. For `custom` data type, a ReadFuncLoading or an ImageStackLoading is expected to be passed alongside the data (which influences the expected input type). Refer to the documentation for more details.
```


src/careamics/lightning/dataset_ng/data_module_utils.py
```diff
@@ -19,6 +19,47 @@
 """Type of input data passed to the dataset."""
 
 
+def _compatible_input_types(data_type: SD) -> str:
+    """String representation of the compatible input types for a given data type.
+
+    Parameters
+    ----------
+    data_type : SD
```
melisande-c: In the docstring the type should be `SupportedData`, although I guess it doesn't matter v much since it is a private function

I think I gave it the alias so the cases would fit on one line in the `initialize_data_pair` func

src/careamics/lightning/dataset_ng/data_module_utils.py
```diff
@@ -19,6 +19,47 @@
...
+            )
+        case SD.TIFF | SD.CZI:
+            return (
+                "the expected input is a str, a pathlib.Path, or a sequence of either"
+            )
+        case SD.ZARR:
+            return (
+                "the expected input is a str or a pathlib.Path (see documentation), or "
+                "a sequence of either"
+            )
```
melisande-c: why a separate case for zarr, the full message in `_initialize_data_pair` already says to look at the documentation

src/careamics/lightning/dataset_ng/data_module_utils.py
melisande-c: I've realised I forgot to exclude `str` from the sequence check here

melisande-c (approved):

src/careamics/lightning/dataset_ng/data_module_utils.py
jdeschamps: oh yes makes sense

src/careamics/lightning/dataset_ng/data_module_utils.py
jdeschamps: I think it is because I was a lot more specific initially. Will update!

src/careamics/lightning/dataset_ng/data_module_utils.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/861/changes/21bade542d5b7b1bae3b2b7c5e6e05c2ea290475

src/careamics/lightning/dataset_ng/data_module_utils.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/861/changes/21bade542d5b7b1bae3b2b7c5e6e05c2ea290475

src/careamics/lightning/dataset_ng/data_module_utils.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/861/changes/21bade542d5b7b1bae3b2b7c5e6e05c2ea290475

PR #862, opened 2026-03-27
Chore: update comment

jdeschamps:
## Description

We identified the following snippet of code as not being clear:

https://github.com/CAREamics/careamics/blob/623f0975241442421d2b7d9e74e4dedcf2968c0d/src/careamics/careamist_v2.py#L538-L539

There is a good reason for it to be there, this PR updates the comments to make it clearer.


melisande-c (approved):
Out of curiosity why does `limit_val_batches` get altered when calling `stop_training`?

jdeschamps:
If I remember correctly it is to avoid the validation loop at the end of the epoch. It may be triggered when the epoch is stopped (or maybe to avoid unlucky timing).

Issue #863, opened 2026-03-27
Next-Gen CAREamist: upper and lower quantile for targets


jdeschamps:
## Describe the bug

Quantile normalization is not compatible with a different number of input and output channels since per channel upper/lower quantiles are applied to both input and target.

The solution would be to have `target_lower_quantiles` and `target_upper_quantiles` as well.

PR #864, opened 2026-03-27
Feat(Next-Gen CAREamics): validate normalization stats against number of channels

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: add a validator to clarify quantile normalization error.

Not passing the correct number of quantile parameters for the quantile normalization will result in an error not caught by the configuration. Two different aspects to that error:
- since `upper/lower_quantiles` when `per_channel=True` are applied to both input and target, then the number of input and output channels must be equal (until we implement `target_upper/lower_quantiles`, see https://github.com/CAREamics/careamics/issues/863)
- if the number of input channels is greater than 1, then the default `upper/lower_quantiles` of length 1 will cause an error.

This PR introduces a validator to cover these cases, and extends the validation to all normalizations via member methods. `NGConfiguration` now calls these methods during validation against the model parameters (number of input and output channels).

I also added tests for the normalizations.


src/careamics/config/architectures/unet_config.py
```diff
@@ -126,3 +126,25 @@ def is_3D(self) -> bool:
...
+    def get_num_input_channels(self) -> int:
+        """
+        Get the number of input channels.
+
+        Returns
+        -------
+        int
+            Number of input channels.
+        """
+        return self.in_channels
```
melisande-c, line 139: why the extra methods? why not just access the attribute

src/careamics/config/ng_configs/ng_configuration.py
```diff
@@ -157,6 +157,51 @@ def validate_3D(self: Self) -> Self:
...
+            output channels do not match, or if the number of quantiles does not match
+            the number of input channels.
+
+        Returns
+        -------
+        Self
+            Validated configuration.
+        """
+        if (
+            self.data_config.normalization.name == "quantile"
```
melisande-c: one way to move the validation into the normalisation config would be to have a method on the normalisation 

```python
def validate_against_channels(in_channels, out_channels) -> tuple[bool, str]:
    """Validate the input and output channels against the normalisation statistics.
    
    If validation fails an error message is also returned.
    """
```
(^ can't remeber if the `ValueError` has to be raised in the validator for Pydantic to work, so returning the error message instead)

Then in current `NGDataConfig` validator can do:
```python
n_in = self.algorithm_config.model.get_num_input_channels()
n_out = self.algorithm_config.model.get_num_output_channels()

validation_passed, err_msg = self.data_config.normalization.validate_against_channels(
    n_in, n_out
)
if not validation_passed:
    raise ValueError(err_msg)
```

tests/unit/config/algorithm/test_algorithm_configs.py
```diff
@@ -0,0 +1,38 @@
...
+    """Test that an algorithm config can be created for all UNet-based algorithms."""
+    algo_config_dict = unet_ng_algo_dict_testing(algorithm=algorithm)
+    cfg = validate_ng_algos(algo_config_dict)
+    assert isinstance(cfg, cfg_class)
+
+
+@pytest.mark.parametrize(
+    "algorithm, n_in, n_out", list(zip(ALGORITHMS, [1, 2, 3], [1, 2, 3], strict=True))
+)
+def test_unet_algorithm_config_channels(algorithm, n_in, n_out):
```
melisande-c, line 34: This is only testing matching input and output channels?

melisande-c (commented):
Is there validatation for the other normalisation stats, e.g. mean-std, against the number of channels?

src/careamics/config/architectures/unet_config.py
jdeschamps: In general, i find this tidier. Now in https://github.com/CAREamics/careamics/pull/869, these methods become part of the interface of model configs (since they will be used by the model constraints).

src/careamics/config/ng_configs/ng_configuration.py
jdeschamps: The secondary methods could raise the error directly!

tests/unit/config/algorithm/test_algorithm_configs.py
jdeschamps: Ah yeah, it was meant to be an itertools call...

jdeschamps:
> Is there validatation for the other normalisation stats, e.g. mean-std, against the number of channels?

No, right now the aim was to validate the parameters that are already there by default since they can easily be incompatible with the data.

Now you are right that validation should be extended to the `None` parameters (those that are calculated on the data), in case one passes them to the configuration to speed up the process. I like the idea to move the check to normalization config.

src/careamics/config/architectures/unet_config.py
melisande-c: Ok we can agree to disagree aha

melisande-c:
> > Is there validatation for the other normalisation stats, e.g. mean-std, against the number of channels?
> 
> No, right now the aim was to validate the parameters that are already there by default since they can easily be incompatible with the data.

Ah I see, I guess I missed the point that this is about default parameters

jdeschamps:
> > > Is there validatation for the other normalisation stats, e.g. mean-std, against the number of channels?
> > 
> > 
> > No, right now the aim was to validate the parameters that are already there by default since they can easily be incompatible with the data.
> 
> Ah I see, I guess I missed the point that this is about default parameters

I did not make it clear!

I added the validation for all the normalizations in https://github.com/CAREamics/careamics/pull/864/commits/d73e547a494b8faf131eecde45265cb49a8741c4 because it does indeed make sense when people pass them.

I will add the tests later.

jdeschamps:
Here is how we stand after https://github.com/CAREamics/careamics/pull/864/changes/b0cc6cd8b765f6d5de345f50ba58f0d112166d7d (added validation methods to all normalizations) and https://github.com/CAREamics/careamics/pull/864/changes/d73e547a494b8faf131eecde45265cb49a8741c4 (added tests):
- added `validate_size(n_input, n_output)` to all normalization configs
- removed tests from `NGConfiguration`, since the validation against model input/output channels just calls `norm.validate_size()`
- added tests for all configurations, and covered the other untested cases

tests/unit/config/algorithm/test_algorithm_configs.py
melisande-c: no wait that makes sense because n2v can't have mismatching in out channels

melisande-c (approved):
Looks good!

although ngl I am struggling a bit understanding some of the tests but seems like everythings covered

tests/unit/config/algorithm/test_algorithm_configs.py
jdeschamps: Yes I realised, I did not change it in the end, I just added another test for the mismatching. (as far as I remember the commit)

jdeschamps:
> although ngl I am struggling a bit understanding some of the tests but seems like everythings covered

In all honestly, I perfectly understand. It took me a while to come up with that structure... There are two blind spots that I pointed out though.

In https://github.com/CAREamics/careamics/pull/864/commits/19d3dde00c3d8b354d24623fc05e744c7b5beff3, I tried to clarify the names of the functions and added a lot of comments. In the hope that the tests become a tad more understandable.

PR #867, opened 2026-03-27
Refac(Next-Gen): Patch filtering tidy-up – remove redundant code and update configs

### LINKED ISSUE #714: Next-Gen Dataset: better background patch exclusion
## Description

Our current background masking relies on a while loop and is extremely inefficient. https://github.com/CAREamics/careamics/pull/710 provides a patch sampling that would enable a better implementation by excluding areas at the sampling level.

Once #710 is merged, background patch exclusion should be implemented using that patching strategy.

melisande-c:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: This PR cleans up the patching filtering. Following #832 there was some redundant code related to the patch filtering, including the patch filter classes no longer needed the probability mechanism to allow some background patches through since this has now been implemented in a different way.

- The patch filter classes no longer need to have a probability of filtering out patches.
- The concept of a coordinate filter is no longer necessary because the filtering happens in a first pass at the start of training.
- Some new parameters relevant to patch filtering were introduced in #832 in the `NGDataConfig` but now these have been moved inside `FilterConfig` to not clutter the main data config.

### Overview - what changed?

- All the patch filter classes have had the stochastic element removed.
- The `CoordinateFilterProtocol` has been removed
- The `MaskFilter` now behaves like the rest of the patch filters and expects to receive a mask patch as an argument to the method `filter_out`.
- `MaxPatchFilter` "threshold_ratio" was unintuitive and has been renamed to "coverage" with the condition for filtering out reversed.
- small refactoring in the `filter_background` functions: the patch filter is now created inside the function, so that `filter_background` is initialised with the patch filter config which contains the other necessary parameters.

- `NGDataConfig`
  - `coord_filter` renamed to `mask_filter`
  - some filtering parameters moved inside `FilterConfig`.
  - `mask_filter` has a default factory so that it is always created for training mode (masking is only triggered if a mask is passed during training)
- Patch Filters
  - removed `p` and `seed` attributes
  - removed stochasticity in `filter_out` method
- `MaxPatchFilter`
  - renamed `threshold_ratio` to `coverage`. Reversed condition for filtering out. 
- `filter_background` and `filter_background_with_mask`
  - now takes the filter config as and argument and creates the patch filter itself
- `create_train_dataset` in `dataset_ng/factory.py`
  - updated for change in `filter_background` and `filter_background_with_mask`

### Removed features or files

- `CoordFilterProtocol`
- `create_coord_filter`

## How has this been tested?

- tests pass (some have been modified for function signature changes)
- one tests added for `mask_filter` default factory in `NGDataConfig`:  `test_mask_filter_default_factory`

## Related Issues

- Resolves #714 

## Breaking changes

NGDataConfig changes related to patch filtering

Some further thought can be put into making filter maps more convenient (and representative) for users to work with to find good thresholds for their data.

---

- [x] Pre-commit passes


src/careamics/config/data/ng_data_config.py
```diff
@@ -278,6 +275,38 @@ def default_in_memory(validated_params: dict[str, Any]) -> bool:
...
+    axes = validated_params.get("axes", "")
+
+    # only create mask filter in training mode
+    if mode != Mode.TRAINING:
+        return None
+
+    # determine if data is 3D
+    is_3d = "Z" in axes
+    if data_type == "czi":
+        is_3d = is_3d or "T" in axes
```
jdeschamps: There is another part of the code where such a check exist (at least one I can think of, although it happens twice in the function). It would be nice to refactor them into a single call:

https://github.com/CAREamics/careamics/blob/623f0975241442421d2b7d9e74e4dedcf2968c0d/src/careamics/config/data/ng_data_config.py#L87-L92

tests/functional/dataset_ng/test_factory.py
```diff
@@ -75,11 +76,11 @@ def test_filter_background_w_mask(
         mode="training",
         axes=axes,
         patch_size=patch_size,
-        filtered_patch_prob=background_prob,
     )
-    coord_filter = coord_filter_dict_testing()
-    config_dict["coord_filter"] = coord_filter
+    # mask filter should always be in the training config by default
```
jdeschamps, line 80: Should that be a unit test?

jdeschamps (approved):
We should now add the background filtering to the `docs/` snippets. Can be a subsequent PR.

tests/functional/dataset_ng/test_factory.py
melisande-c: I did add this unit test to test the default `mask_filter`, I think the assert below was mainly so I didn't get the red squigglies

https://github.com/CAREamics/careamics/blob/376006940a0a5fa4041cd9cae2a024c95755650a/tests/unit/config/data/test_ng_data_config.py#L899-L901

jdeschamps: Sounds good

src/careamics/config/data/ng_data_config.py
melisande-c: done in [ee696e0](https://github.com/CAREamics/careamics/pull/867/commits/ee696e0cf56bffa277bc61915cb93ecf93c3a215)

PR #868, opened 2026-03-27
Fix(Next-Gen CAREamics): ensure that the normalization stats are saved in the hyperparameters

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Currently, the data config saved in the checkpoint does not contain the normalization statistics.

When loading a checkpoint for prediction, statistics are systematically recalculated. The reason is that the data module saves the data configuration in the hyperparameters before the normalization statistics are computed.

This PR refactors the saving of the hyperparameters in a private method of the data module, and call it in `setup` after the training dataset has been created, ensuring the presence of the statistics.

# What changed

In `CAREamicsDataModule`:
- `self.save_hyperparameters` call has been refactored to `self._save_hparams` method
- call to `self._save_hparams` happens now in `setup`
- assertion added to test to verify the presence of the stats


src/careamics/lightning/dataset_ng/data_module.py
```diff
@@ -336,6 +322,28 @@ def _sampler(self, dataset: Literal["train", "val", "predict"]) -> Sampler | Non
             sampler = None
         return sampler
 
+    def _save_hparams(self) -> None:
+        """Save configuration in hyperparameters."""
+        self.save_hyperparameters(
+            {"data_config": self.config.model_dump(mode="json")},
+            ignore=[
```
melisande-c, line 329: oops just noticed these ignored params weren't updated when we changed the signature of the data module initialisation

melisande-c (approved):

src/careamics/lightning/dataset_ng/data_module.py
jdeschamps: Oh good catch, I didn't even think about it when I refactored it.

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/868/commits/e07d86306e443a51cd77cfd09741ee80ef80d9a6

PR #869, opened 2026-03-30
Feat(Next-Gen CAREamics): add model constraints validation in configuration and dataset

### LINKED ISSUE #741: Next-Gen CAREamics: fail early if odd image sizes and no tiling during prediction

jdeschamps:
## Description

> [!NOTE]  
> **tldr**: Introduce a mechanism to validate patch size, tile size or whole image size, against the model constraints to ensure that the input tensor shape is not leading to unclear errors but rather inform the users on what to do.

This PR introduces `ModelConstraints` that are called in the configuration and in the dataset to validate the size of the input to the model (patch size, tile size or image size).

Errors thrown by the model itself due to mismatching tensors and weights will not be clear to users (https://github.com/CAREamics/careamics/issues/741). https://github.com/CAREamics/careamics/pull/824#pullrequestreview-4031872628 introduces a soft constraint that does not cover all cases and leads to multiple definitions of what is an acceptable input shape (patch/tile size vs whole image input).  

### Overview - what changed?

A `ModelConstraints` object is used in the configuration and in the dataset to validate input size.

- Defined `ModelConstraints` protocol
- Added implementation of the protocol `UNetConstraints`
- Added a factory
- `NGConfiguration` calls the model constraints to validate patch and tile shape
- `CAREamistv2` creates a model constraints instance and propagates it to the prediction dataset
- `Dataset` validate input shape when using whole image strategy

## How has this been tested?

New tests and scripts.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/741

---

- [x] Pre-commit passes


src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
```diff
@@ -0,0 +1,61 @@
...
+        """Constructor.
+
+        Parameters
+        ----------
+        model_config : UNetConfig
+            The UNet model configuration from which to derive constraints.
+        """
+        self.model_config = model_config
+
+    def validate_input_shape(self, input_shape: Sequence[int]) -> None:
```
melisande-c: I guess we could also check if input and output channels are correct, might be useful for microsplit. But we can just wait to implement it when it comes up

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
```diff
@@ -0,0 +1,61 @@
...
+                f"Spatial input shape to model constraints should have length 2 (YX) or"
+                f" 3 (ZYX), but got shape {input_shape}."
+            )
+
+        dim_label = "ZYX" if len(input_shape) == 3 else "YX"
+
+        # check spatial dims against model depth constraints
+        depth = self.model_config.depth
+        for i, dim in enumerate(input_shape):
+            if dim % (2**depth) != 0 or dim < 2**depth:
```
melisande-c: The `dim < 2**depth` is to catch the case that `dim == 0`?

src/careamics/dataset_ng/dataset.py
```diff
@@ -185,6 +200,15 @@ def __init__(
                     "Not all images sizes are greater or equal than the patch size for "
                     "training and validation."
                 )
+        else:
+            # model constraints not applied if tiled, since the tiles are validated
+            # in the configuration itself
+            if model_constraints is not None and not isinstance(
+                data_config.patching, TiledPatchingConfig
+            ):
```
melisande-c: maybe better to check `isinstance(data_config.patching, WholePatchingConfig)` because technically the model constraints could be passed with training patching

src/careamics/dataset_ng/dataset.py
```diff
@@ -185,6 +200,15 @@ def __init__(
                     "Not all images sizes are greater or equal than the patch size for "
                     "training and validation."
                 )
```
melisande-c, line 202: This check is somewhat related, in that it is validating the spatial shapes (but not dependent on the model). 

 (just an observation)

tests/test_careamist_v2_predict.py
melisande-c, line 1: Might be a good idea to test the validation at the configuration level as well with all the `_PatchedConfig` patching strategy types

melisande-c (commented):
Makes sense, some of the comments are just thoughts, feel free to ignore those

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
jdeschamps: Yes, I actually tried that briefly and pedaled back. I think the reason was that the `channels` parameter (which selects a subset of channels) makes it complicated. But now I don't really see how.

Let me have a crack at it.

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
jdeschamps: Yes, which may be unnecessary

src/careamics/dataset_ng/dataset.py
jdeschamps: That is true!

jdeschamps: Actually the clause is reached only when  `mode=="predicting"`, so it will be either tiling or whole.

src/careamics/dataset_ng/dataset.py
jdeschamps: true, I am not sure how to reconcile those though!

tests/test_careamist_v2_predict.py
jdeschamps: yes, this test is inherited from https://github.com/CAREamics/careamics/pull/824

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
jdeschamps: Speaking of, `channels` were not validated against model input size. Added that in https://github.com/CAREamics/careamics/pull/869/changes/c84e093d8de80713c7f91a210cb58918f1a442ae.

src/careamics/dataset_ng/dataset.py
jdeschamps: (nonetheless updated the if statement in https://github.com/CAREamics/careamics/pull/869/changes/c84e093d8de80713c7f91a210cb58918f1a442ae, because I had already changed it...)

tests/test_careamist_v2_predict.py
jdeschamps: Now I remember the reason why I have not tested that particular validation here. In https://github.com/CAREamics/careamics/pull/864 I introduce the first unit tests for the NG configuration, and they come with the relevant functions in `/tests/utils`. I was hoping it would be reviewed earlier...

I'll wait that it is merge to add them.

src/careamics/dataset_ng/dataset.py
jdeschamps: > Actually the clause is reached only when `mode=="predicting"`, so it will be either tiling or whole.

The reason being that patches and tiles are already validated in the configuration.

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
melisande-c: then it might be clearer to have `if dim % (2**depth) != 0 or dim == 0`

src/careamics/dataset_ng/dataset.py
melisande-c: Oh oops I guess I lost track of the nested if-else blocks

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
jdeschamps: > Yes, I actually tried that briefly and pedaled back. I think the reason was that the `channels` parameter (which selects a subset of channels) makes it complicated. But now I don't really see how.
> 
> Let me have a crack at it.

Added channel validation to the dataset in https://github.com/CAREamics/careamics/pull/869/commits/f23a9cddcdceef909372f905889b96c22f196393.

src/careamics/lightning/dataset_ng/lightning_modules/constraints/unet_constraints.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/869/changes/67db4067b09cb2cfd43312b0c254b317d9c94db2

tests/test_careamist_v2_predict.py
jdeschamps: Added tests in https://github.com/CAREamics/careamics/pull/869/changes/7da8bcfc381eecd25f1fcdf492d78428eec90e54, mostly using mocking

tests/unit/config/ng_configs/test_ng_configuration.py
```diff
@@ -29,3 +69,202 @@ def test_unet_configs(algorithm, config_class):
...
+        default_training_dict,
+    )
+
+    unet_config_dict = unet_ng_config_dict_testing(algorithm=algorithm)
+
+    with patch(DEFAULT_TRAINING_FACTORY_PATH, wraps=default_training_dict) as mock:
+        instantiate_config(unet_config_dict)
+
+        mock.assert_called_once()
+        assert mock.call_args.kwargs["algorithm"] == algorithm
```
melisande-c: mmm I kinda feel this is testing pydantic's default factory system more than anything else

tests/unit/config/ng_configs/test_ng_configuration.py
```diff
@@ -29,3 +69,202 @@ def test_unet_configs(algorithm, config_class):
...
+    ],
+)
+def test_validate_patch_against_model(mode, patching, n_calls):
+    """Test that `validate_patch_against_model` calls model constraints
+    `validate_spatial_shape`.
+
+    Actual `validate_spatial_shape` is unit tested separately in its own module
+    test file. This test simply check that if there is a `patch_size`, the validation
+    from model constraints is skipped, otherwise it is called with the correct
+    parameters.
```
melisande-c: mocking can be useful, but it can also make things a bit more confusing. In a way, for something that is quick to test, maybe it doesn't hurt to have some duplication and to just check that patch sizes divisible by `2**depth` are accepted. Similar comments for the other 3 model constraints tests here. What do you think?

melisande-c (approved):

tests/unit/config/ng_configs/test_ng_configuration.py
jdeschamps: I had back and forth about that one, but it annoyed me that there were no tests to show that training was instantiated with the correct algorithm (since the default constructor calls a downstream function). This test only checks that the correct algorithm is propagated all the way to that Training config dict utility. So it overlaps more with the testing of the factory function itself.

I can remove it and we can keep just the factory function tests themselves.

tests/unit/config/ng_configs/test_ng_configuration.py
jdeschamps: Agree that mocking is really hard to read. But I don't see why we should test something underconstrained or repeat the same constrains (which I guess we would import from the model constraints tests to avoid duplication of definitions).

Also testing the configuration is complex because there are a lot of dependencies between parameters that make the configuration of the tests parameter also complex.

I can give it a try.

tests/unit/config/ng_configs/test_ng_configuration.py
melisande-c: fair enough, I guess it does not hurt to keep it

jdeschamps: Removed in https://github.com/CAREamics/careamics/pull/869/commits/1c02cc591f9cd9ef223cbd63fb2d2f888ef2d631

tests/unit/config/ng_configs/test_ng_configuration.py
jdeschamps: Ok, it was not as bad as I thought, the utility functions are flexible enough. Done in https://github.com/CAREamics/careamics/pull/869/changes/ffe8633c0f9a2775597bcb6825e324dc8b9e11d4

PR #874, opened 2026-04-07
Fix(Next-Gen CAREamics): ignore adding early stopping callback if `early_stopping_params` is an empty dict

mese79:
## Disclaimer

- [ ] I am an AI agent.

By default, the `early_stopping_params` is an empty dict when there's no need for early stopping callback. But currently the condition only checks if the `early_stopping_params` is not `None`. So, it will raise an error trying to add an EarlyStopping callback with no value for the `monitor` param.

- `src/careamics/careamist_v2.py`

## How has this been tested?
I tested vit ia careamics napari ui.
---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist_v2.py
+++ b/src/careamics/careamist_v2.py
@@ -377,7 +377,7 @@ class CAREamistV2:
         if enable_progress_bar:
             internal_callbacks.append(ProgressBarCallback())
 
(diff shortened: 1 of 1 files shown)


jdeschamps (approved):

jdeschamps:
Thanks!

PR #871, opened 2026-03-31
Feat: Next-Gen CAREamics

jdeschamps:
This PR brings about the Next-Gen CAREamics, a completely refactored code-base that is more modular and maintainable. It comes with a powerful, extensible and flexible dataset, including Zarr support and background exclusion. It will make the inclusion of new algorithm much easier in the next versions.

### Changes

- Moving `v0.1` modules to `careamics.compat`
- Renaming all NG CAREamics modules to remove `ng`
- Clean up module hierarchy and imports
- Add checks, error catching and better validation

### Corresponding PRs

- #886
- #893
- #896
- #894 
- #898
- #892
- #895
- #906
- #897
- #905
- #900 
- #909
- #917
- #905
- #920
- #916
- #919
- #922
- #925
- #923
- #932
- #928 
- #927
- #935 
- #937 
- #942
- #933 
- #918 
- #928 
- #917 
- #947

PR #886, opened 2026-04-10
Refac(v0.2): move next-gen modules

jdeschamps:
## Description

This PR moves Next-Gen CAREamics modules out of the `ng` submodules in which they where developed to clarify the module structure.

It should be merged in `dev/v0.2` (current draft PR https://github.com/CAREamics/careamics/pull/871) so that we can release `v0.2`.


veegalinova:
Some small thoughts looking at file tree: 
1. Should the "ng demo" folder be removed? 
2. `lightning/data_*` could be in a separate folder 
3. Some files that are in the dataset main folder seem more like utils? 
4. There is a rouge `nm_training_placeholder.py` in the main folder

jdeschamps:
> Some small thoughts looking at file tree:
> 
>     1. Should the "ng demo" folder be removed?
> 
>     2. `lightning/data_*` could be in a separate folder
> 
>     3. Some files that are in the dataset main folder seem more like utils?
> 
>     4. There is a rouge `nm_training_placeholder.py` in the main folder

1. I was waiting a little bit on that one because these files might be useful for documentation. Maybe I move them to a different folder as it would be cleaner.
2. Maybe we should discuss more globally the content of `lightning`, because `lightning.lightning_modules` is also a bit annoying.
3. Yes that's true, let's discuss those.
4. I didn't want to touch that, but it can indeed be easily moved. Maybe to some `careamics.dev`.

jdeschamps:
I did 1 and 4 in recent commit, left 2 and 3 for discussions.

melisande-c (approved):

PR #892, opened 2026-04-14
Refac(v0.2): Use consistent naming for all protocols and implementations

jdeschamps:
## Description

Here I went back and forth and was not so convinced anymore... Given that in a lot of cases we do use the type of protocol in the module names, here is a proposal:

- Protocol: no `_protocol` or `*Protocol` in module or class name
- Implementations: `<name>_<protocol_name>.py` for the module, `<Name><ProtocolName>` for the class

Names will be long, but it lifts doubts about what the modules and classes are. I tried to apply this to the callbacks and the prediction write strategies.


melisande-c (approved):
Makes sense, yeah I prefer not to have `*Protocol` in the class names because without it the type hints look cleaner.

So now we need to make sure the following are consistent:
- normalisation
- patching strategies
- patch filters

Maybe we should start a style guide document with the conventions we have chosen, either in CONTRIBUTING.md or another file

jdeschamps:
> Makes sense, yeah I prefer not to have `*Protocol` in the class names because without it the type hints look cleaner.
> 
> So now we need to make sure the following are consistent:
> 
>     * normalisation
> 
>     * patching strategies
> 
>     * patch filters
> 
> 
> Maybe we should start a style guide document with the conventions we have chosen, either in CONTRIBUTING.md or another file

Went maybe a bit overboard... https://github.com/CAREamics/careamics/pull/892/changes/eb81989c55eb3c89713b90962a3fa667d1d7f5fc

Added style guide https://github.com/CAREamics/careamics/pull/892/changes/64ebdafecadca57d7a135d6b947963ba8fb42a17

src/careamics/dataset/patching/whole_sample.py
melisande-c, line 1: whole_sample_patching.py

melisande-c (approved):

src/careamics/dataset/patching/whole_sample.py
jdeschamps: Good catch!

jdeschamps:
Nice title, thanks!

PR #893, opened 2026-04-14
Refac(v0.2): refactor transforms

jdeschamps:
## Description

- move+rename `transforms` to `careamics.dataset.augmentation`
- move n2v transform to `careamics.lightning.lightning_modules.n2v_utils`


src/careamics/dataset/augmentation/transform.py
melisande-c, line 1: I think we should make a note to improve this base class, but it will not change the API so it can wait for after v0.2 release

melisande-c (approved):

src/careamics/dataset/augmentation/transform.py
jdeschamps: Left a TODO in https://github.com/CAREamics/careamics/pull/893/changes/36816f467459a29bfd3447f3c57a4705949bd257

PR #898, opened 2026-04-15
Refac(0.2): better optimizer validation

jdeschamps:
## Description

- Remove unused functions (`get_optimziers` and `get_schedulers`) and their corresponding tests
- Repurpose `filter_parametes` to check unknown parameters, raising an error during Pydantic validation of the optimizers config
- Add unit tests for the Pydantic models
- Rename `torch_utils` to `get_device`

Note: I had to remove `verbose` from `ReduceLROnPlateau` config in `create_microsplit_config` since it is deprecated and not valid in the latest torch we support.


src/careamics/config/lightning/optimizer_configs.py
```diff
@@ -57,7 +84,7 @@ class OptimizerConfig(BaseModel):
 
     @field_validator("parameters")
     @classmethod
-    def filter_parameters(cls, user_params: dict, values: ValidationInfo) -> dict:
+    def valid_parameters(cls, user_params: dict, values: ValidationInfo) -> dict:
```
melisande-c: nit: rename to `validate_parameters` ?

melisande-c (approved):
Looks good

we could use a similar method to validate the trainer params

jdeschamps:
> Looks good
> 
> we could use a similar method to validate the trainer params

That's a good idea!

src/careamics/config/lightning/optimizer_configs.py
jdeschamps: It is not a good example, but I rather like it when the validators name are what they validate. e.g. `power_of_2()`, `list_of_even()`. In this case it is whether the parameters are valid entries.

melisande-c: ah i see, my brain just didn't parse the meaning

jdeschamps:
> Looks good
> 
> we could use a similar method to validate the trainer params

Done in https://github.com/CAREamics/careamics/pull/898/changes/45987a7ad0e7707fa17409071265ea6f6b0bd5f0
/

Issue #829, opened 2026-03-16
Predicting with WholePatchingStrategy needs `batch_size=1` for differently sized arrays


melisande-c:
## Problem
If predicting with whole patching strategy, all of the input images need to have the same spatial dimensions, or the batch size needs to be 1, otherwise there will be an error when pytorch tries to stack the images to form the batch.

## Potential solution
A clear error raised in CAREamistV2.

melisande-c:
Not sure if this is true, but maybe `convert_mode` will keep the same batch_size as in training, which could cause this issue

jdeschamps:
Check should happens here: https://github.com/CAREamics/careamics/blob/9629661ad2d2eff8e2dd0ac3e27efc1c36964e11/src/careamics/dataset_ng/dataset.py#L208

(note link to code is pre-`dev/v0.2`)

jdeschamps:
Closed by #916

PR #916, opened 2026-04-22
Fix(v0.2): catch error when predicting with inhomogeneous shapes and batch size > 1

jdeschamps:
## Description

- Rename functions and types to remove "NG"
- Refactor the data shape validation in the dataset to make it easier to test it
- Add validation of prediction with whole image against batch size, to warn users against incompatible shapes when using batch size > 1.  

Closes https://github.com/CAREamics/careamics/issues/829


src/careamics/dataset/dataset.py
```diff
@@ -120,56 +225,37 @@ def __init__(
...
-                if target_extractor is not None:
-                    for image_stack in target_extractor.image_stacks:
-                        model_constraints.validate_target_channels(
-                            image_stack.data_shape[1]
-                        )
+            target_data_shapes = (
+                [
+                    image_stack.data_shape
+                    for image_stack in target_extractor.image_stacks
+                ]
```
melisande-c: pretty sure you can do `target_extractor.shapes`

melisande-c (approved):

src/careamics/dataset/dataset.py
jdeschamps: Ah nice!

Done in https://github.com/CAREamics/careamics/pull/916/changes/459de0b10081bf0e26e49c44b131ec9dbaef1e82

PR #918, opened 2026-04-22
Feat(v0.2): default checkpoint for CARE vs N2V/N2N

### LINKED ISSUE #305: Show users available checkpoints

### LINKED ISSUE #606: Next-Gen CAREamics: auto checkpoint loading, N2V vs CARE
## Problem

In Noise2Void, it is often preferable to load the last checkpoint (basically keep the latest weights), especially because the loss is not indicative of performances once it has "plateaued".

For CARE, on the other hand, the "best" checkpoint should be loaded.

We currently have a single default for all algorithms. But there should be a switch depending on the algorithm. This could spark a larger conversation on how to reconcile different algorithm within the configuration, before it becomes an unmanageable monster due to the LVAE algorithms.

jdeschamps:
## Description

For CARE, we would usually load the best checkpoint, while for N2V/N2N we would load the last checkpoint.

This PR introduces the following:
- Add `checkpoint` parameter to `predict` and `predict_to_disk`
- If `None`, `checkpoint` defaults to `last` for N2V and N2N, and `best` for the others (CARE)
- Removed best checkpoint loading utility from `model_utils` and `CAREModule`
- Following https://github.com/CAREamics/careamics/pull/823:
    - Added `get_checkpoints` and adapted to return all checkpoint Paths 
    - Refactored all checkpoint root path into a single method
- Added a `self.run_version` and a simple run version mechanism (`careamics.utils.get_run_version`): the reason is that otherwise checkpoints are all pulled into the same folder if the experiment name does not change. In the mean time, logs are versioned by Lightning. Here, `CAREamist` starts at `0` and inspect the checkpoint folder for any `<experiment_name>_X` and record the current run version as `X+1`. The version is also passed to the logger to maintain the same version between logs and checkpoints.

## Example

Multiple runs of experiment named `my_exp` in working directory `experiments`.

```python
from careamics.careamist import CAREamist
from careamics.config.factories import create_advanced_n2v_config
import numpy as np
from pathlib import Path

# create a configuration
    experiment_name="my_exp",
    data_type="array",
    axes="YX",
    patch_size=[64, 64],
    batch_size=16,
    num_epochs=2,
    num_steps=10,
)

# instantiate a careamist
careamist = CAREamist(config, work_dir=root / "experiments")
careamist.train(
    train_data=[np.random.rand(512, 512)],

)
```

``` bash
my_exp
├── checkpoints
│        ├── my_exp_0
│        ├── my_exp_1
│        ├── my_exp_2
│        └──my_exp_3
└── csv_logs
         ├── version_0
         ├── version_1
         ├── version_2
         └──version_3
```

Where each version number corresponds to each other.

Note that removing `my_exp_3` will cause the next run to re-create it and overwrite `csv_logs/version_3`.

Resolves #606
Resolves #305
Closes #823


src/careamics/lightning/modules/care_module.py
veegalinova: Idk if it suggests correctly - `self._best_checkpoint_loaded` is not used anymore and can be deleted
```suggestion
```

src/careamics/careamist.py
```diff
@@ -743,6 +819,11 @@ def predict(
             Loading strategy to use for the prediction data. May be a ReadFuncLoading or
             ImageStackLoading. If None, uses the loading strategy from the training
             configuration.
+        checkpoint : str or Path, optional
+            Checkpoint to load before making predictions. Can be "best", "last", or a
+            path to a specific checkpoint. If None, uses the last checkpoint from
+            training Noise2Void or Noise2Noise models, otherwise the last checkpoint.
```
veegalinova:
```suggestion
            training Noise2Void or Noise2Noise models, otherwise the best checkpoint.
```

src/careamics/careamist.py
```diff
@@ -907,6 +995,11 @@ def predict_to_disk(
             Loading strategy to use for the prediction data. May be a ReadFuncLoading or
             ImageStackLoading. If None, uses the loading strategy from the training
             configuration.
+        checkpoint : str or Path, default=None
+            Checkpoint to load before making predictions. Can be "best", "last", or a
+            path to a specific checkpoint. If None, uses the last checkpoint from
+            training Noise2Void or Noise2Noise models, otherwise the last checkpoint.
```
veegalinova:
```suggestion
            training Noise2Void or Noise2Noise models, otherwise the best checkpoint.
```

src/careamics/careamist.py
```diff
@@ -427,17 +447,70 @@ def _create_loggers(
...
+                raise ValueError(
+                    f"{self.config.algorithm_config.get_algorithm_friendly_name()} does"
+                    f" not have a well-defined best checkpoint based on validation "
+                    f"loss. Please specify an explicit checkpoint path."
+                )
+            elif checkpoint is None:
+                return "last"
+
+        if checkpoint is not None:
+            return checkpoint
```
veegalinova, line 511: This should be added, otherwise careamist loaded from the checkpoint will fail at prediction

```suggestion
        if checkpoint is not None:
            return checkpoint
            
        if self.checkpoint_path is not None:
            return self.checkpoint_path
```

veegalinova (changes requested):

src/careamics/lightning/modules/care_module.py
jdeschamps: Good catch

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/918/changes/118bf878cb1cde129026223580ebe207f27a3b33

veegalinova (approved):

jdeschamps:
Actually, I ran into an issue. Just pasting the code here for my future-self:

```python

    ckpt = Path("care_denoising/checkpoints/care_denoising/care_denoising_last.ckpt")

    data_paths = [
        path
        for path in root.glob("EXP278_Smed_fixed_RedDot1_sub_5_N7*.tif")
    ]
    assert len(data_paths) > 1, "Expected multiple tifs in the test data folder"

    # instantiate a careamist
    careamist = CAREamist(checkpoint_path=ckpt)

    # once trained, predict
    careamist.predict_to_disk(
        pred_data=data_paths, 
        tile_size=[32, 128, 128], 
        tile_overlap=[16, 48, 48],
        data_type="tiff",
       prediction_dir = "n2v_denoising/predictions",
    )


ValueError: `.predict(ckpt_path="best")` is set but `ModelCheckpoint` is not configured to save the best model.
```

This stems from CAREamist not being trained, loaded from checkpoint. Prediction function attempts to load a checkpoint. It is a bit counter intuitive that we need to set the checkpoint twice (in order to solve that issue). 

~~Does CARE really save best throughout? Or only at the end?~~ Need to check but probably it just memorizes the name of the checkpoint as best.

jdeschamps:
Could not reproduce on a clean script.

Issue #878, opened 2026-04-08
Next-Gen: stitch multi-sample prediction broken if axes has `"T"` instead of `"S"`


melisande-c:
## Describe the bug

The `stitch_single_prediction` function decides if the data is multi-sample or not by checking if `"S"` is in the axes. However this is incorrect because the axes stored in the `ImageRegionData` is the original axes and using T instead of S, e.g. "TYX" is perfectly valid.

https://github.com/CAREamics/careamics/blob/88235396d21fcac040d4b35223529411a0cd0dc7/src/careamics/lightning/dataset_ng/prediction/stitch_prediction.py#L109

## Solution

I am pretty sure we don't even need to have this if-else block and the multi-sample code will work for a single-sample prediction. 

Need to check what happens in `group_tiles_by_key` to be sure.

PR #920, opened 2026-04-22
Fix(v0.2): simplifies stitching and fix potential error

jdeschamps:
## Description

As correctly assessed in https://github.com/CAREamics/careamics/issues/878, the `if else` clause was not necessary and introduced a potential error.

This PR removes the clause.

Closes https://github.com/CAREamics/careamics/issues/878


src/careamics/dataset/image_region_data.py
```diff
@@ -43,7 +43,7 @@ class ImageRegionData(NamedTuple, Generic[RegionSpecs]):
     """Source of the data, e.g. file path, zarr URI, or "array" for in-memory arrays."""
 
     data_shape: Sequence[int]
-    """Shape of the original image in (SCZ)YX format and order. If channels are
+    """Shape of the original image in SC(Z)YX format and order. If channels are
```
melisande-c: I'm guessing this was originally meant to imply SCZ are optional but YX are mandatory

I've just noticed that there is also an `original_data_shape` field? I guess the docstring here is wrong and it is the reshaped shape?

melisande-c (approved):

src/careamics/dataset/image_region_data.py
jdeschamps: Maybe, just to be sure I had a second look. By looking at the implementation in the `InMemoryImageStack`, `data_shape` is `data.shape` so it would be SC(Z)YX since reshaping is supposed to be called when getting the array. Likewise for the other image stacks.

I think it is supposed to be read as "this is the original data shape but rearranged in a different format". But it is not clear.

Improved in https://github.com/CAREamics/careamics/pull/920/commits/2b95aaccc6abc0ec81dab163c97861ae80f1fa53

PR #922, opened 2026-04-22
Chore(v0.2): warn users when using batch norm and batch size is amll

### LINKED ISSUE #488: Warn users when using BatchNorm with small batch size
## Problem

Currently, BatchNorm layers are present in the models, and users can set whichever batch size they want. We should probably have the configuration throw a warning if batch size is `<32` or `<8` when using batch norm, and point to deactivating the batch norm layers (possible since a recent PR).

Opinions?

jdeschamps:
## Description

Small batch size when using batch norm can be unstable. This PR adds a warning during validation of the configuration.

Closes https://github.com/CAREamics/careamics/issues/488

### DIFF
--- a/src/careamics/config/configuration.py
+++ b/src/careamics/config/configuration.py
@@ -216,7 +216,7 @@ class Configuration(BaseModel, Generic[AlgorithmConfig]):
         """
         if (
             self.algorithm_config.model.uses_batch_norm()
-            and self.data_config.batch_size < 32
+            and self.data_config.batch_size < 8
         ):
             warnings.warn(
                 f"Warning: Batch normalization is used with batch size "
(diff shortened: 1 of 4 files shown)


src/careamics/config/configuration.py
```diff
@@ -204,6 +205,30 @@ def validate_norm_against_channels(self: Self) -> Self:
...
+        """Warn if batch normalization is used with small batch size.
+
+        Returns
+        -------
+        Self
+            Validated configuration.
+        """
+        if (
+            self.algorithm_config.model.uses_batch_norm()
+            and self.data_config.batch_size < 32
```
veegalinova:
```suggestion
            and self.data_config.batch_size < 8
```

veegalinova: I think 32 is too large, I'm not sure it's an issue with batch size greater than 8

veegalinova (approved):

PR #937, opened 2026-04-28
Feature: Set patch filtering through config factory functions

### LINKED ISSUE #901: Next-Gen(Configs): Add patch filtering params to convenience functions
## Problem

The patch filtering arguments are not accessible through the convenience config factory functions.

The only way to use patch filtering is to modify the config or to build it from scratch.

melisande-c:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: Adds the patch filtering to the "advanced" config factory functions.

### Overview - what changed?

New `patch_filter_config` argument in `create_advanced_n2v_config`, `create_advanced_care_config` and `create_advanced_n2n_config`. I decided to pass the entire patch filter configuration to the factory functions because I tried with dictionaries but it felt a bit clunky. The patch filters have required non-default arguments which are hard to explain all the cases for in the docs and users would get no hints with plain dictionaries.

Exported the patch filter configs at the root of the config package, to make it's use more convenient.

Additional validator in the `DataConfig` that will set the coverage in `MaxPatchFilterConfig` to 0.25 if the data is 2D and 0.125 if the data is 3D only if the coverage has not been explicitly set by the users.

### New features or files

- `set_default_max_patch_filter_coverage` in `DataConfig`

- `create_advanced_care_config`
- `create_advanced_n2n_config`
- `_create_advanced_supervised_config`
- `create_advanced_n2v_config`
- `create_ng_data_config` (I guess "ng" should be removed)

## How has this been tested?

Added unit tests to ensure patch filter config is in the created configuration.

Added unit tests to check the max filter coverage is updated only when not explicitly set.

## Related Issues

- Resolves #901

Creating n2v config with max filtering
```python
from careamics.config import create_advanced_n2v_config, MaxPatchFilterConfig

    "with-patch-filter",
    data_type="array",
    axes="YX",
    patch_size=(64, 64),
    batch_size=32,
    patch_filter_config=MaxPatchFilterConfig(threshold=100)
)
```

#### Additional thought

Maybe we could do a similar thing with the normalization and pass the `norm_config` removing the `normalization_params` argument? I think still allow just the name to be passed if the user wants automatic stats calculation, so something like:

```python
NormalizationConfig = MeanStdConfig | MinMaxConfig | QuantileConfig | NoNormConfig

    ...
    normalization: Literal["mean_std", "min_max", "quantile", "none"] | NormalizationConfig
    ...
```

---

- [x] Pre-commit passes


jdeschamps (approved):
Cool! I agree, we could add a similar mechanism for the normalization.

PR #942, opened 2026-04-29
Fix(v0.2): training reproducibility

### LINKED ISSUE #460: Refactoring: reproducibility of the training pipeline
## Description
Currently, there's no way to ensure full reproducibility of the training pipeline. By "full reproducibility", I mean being able to exactly reproduce training loss values across multiple runs.

At the moment, the pipeline has two main sources of randomness: patch extraction and data transforms. All random operations are correctly using numpy's `default_rng()`. But, since the seed is None by default, each run will initialize the RNG differently, leading to variability between runs.

### Examples

- [patch extraction example](https://github.com/CAREamics/careamics/blob/488255829a3e95acde4252752b2f793ef950cf71/src/careamics/dataset/patching/sequential_patching.py#L136)
- [xy_flip transform example](https://github.com/CAREamics/careamics/blob/488255829a3e95acde4252752b2f793ef950cf71/src/careamics/transforms/xy_flip.py#L46)

``` python
rng = np.random.default_rng()  # seed not set, nor passed from the init
...
rng.shuffle(patches, axis=0) # will be randomized between runs
```

## Potential solutions
1) add a `random seed` parameter to the data configuration (and maybe consider setting a value by default?)
2) propagate this parameter through the pipeline to the patching and to the transforms

## Additional context 
For clarity, to achieve complete reproducibility, two more settings need to be handled outside the repo's code:
1) setting the seeds for each of the workers, can be done with lightning:
```python
from pytorch_lightning import seed_everything
seed_everything(42)
```
2) enabling torch determinism (may slightly impact performance):
```
torch.use_deterministic_algorithms(True)
torch.backends.cudnn.benchmark = False
```

We also need to insure that it will not affect the performance of current algorithms.

veegalinova:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: added seed fixing into the CAREamist code

### Overview - what changed?

After fixing seeds in the config, there was one more necessary change for the reproducibility - fixing lightning seeds in the CAREamist code, with seeding individual workers enabled.

Since the seed is fixed from inside the code (and not outside as usual for lightning pipelines), there are two actual calls to `seed_everything` needed - one after the config is loaded, before model is created (for deterministic weight initialization) and one before training (after Trainer and cuda init consume the state).

- `careamist.py` - added `seed_everything` calls after config loading and before training start

## How has this been tested?

After fixing torch determinism, the losses were numerically identical for both zero and multiple workers on several runs 
```python
torch.use_deterministic_algorithms(True)
torch.backends.cudnn.benchmark = False
```

## Related Issues

- Resolves #460 
---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -136,8 +136,6 @@ class CAREamist:
         self.config: ConfigurationType
         self.config, self.model = self._load_model(config, checkpoint_path, bmz_path)
 
-        _ = seed_everything(self.config.data_config.seed, workers=True)
-
         self.config.training_config.trainer_params["enable_progress_bar"] = (
             enable_progress_bar
         )
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -231,7 +231,7 @@ class CAREamist:
             config = load_configuration(Path(config))
         assert not isinstance(config, (Path, str))
 
-        _ = seed_everything(config.data_config.seed, workers=True)
+        _ = seed_everything(config.data_config.seed, workers=True, verbose=False)
         model = create_module(config.algorithm_config)
         return config, model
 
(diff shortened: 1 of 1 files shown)


jdeschamps (approved):
Great, thanks!

PR #947, opened 2026-05-07
v0.2: prediction oom error

### LINKED ISSUE #941: Catch OOM during prediction
## Problem

We should probably catch OOM during prediction and indicate on which parameters users can play. For instance prediction from a CAREamist instnatiated from a checkpoint probably means that we keep training `batch_size` (128) for prediction.

```python
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 8.00 GiB. GPU 0 has a total capacity of 15.80 GiB of which 1.87 GiB is free. Including non-PyTorch memory, this process has 12.56 GiB memory in use. Of the allocated memory 9.38 GiB is allocated by PyTorch, and 2.88 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

veegalinova:
## Description

A small PR adding a custom error message where OOM is triggered. 

- Resolves #941

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -863,9 +863,8 @@ class CAREamist:
         except RuntimeError as e:
             if "out of memory" in str(e).lower():
                 raise RuntimeError(
-                    "Out of memory error during prediction. Try reducing batch size or "
-                    "tile size, e.g. predict(..., batch_size=1) or "
-                    "predict(..., tile_size=(64, 64))."
+                    "\nOut of GPU memory during prediction.\n"
+                    "Try reducing the batch size: predict(..., batch_size=1)\n"
                 ) from e
             raise
         tiled = tile_size is not None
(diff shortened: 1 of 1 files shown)


src/careamics/careamist.py
```diff
@@ -856,9 +856,17 @@ def predict(
...
 
-        predictions: list[ImageRegionData] = self.trainer.predict(
-            model=self.model, datamodule=datamodule, ckpt_path=checkpoint
-        )  # type: ignore[assignment]
+        try:
+            predictions: list[ImageRegionData] = self.trainer.predict(
+                model=self.model, datamodule=datamodule, ckpt_path=checkpoint
+            )  # type: ignore[assignment]
+        except RuntimeError as e:
+            if "out of memory" in str(e).lower():
```
jdeschamps, line 864: Here is an idea:
- If no tiling and batch size > 1: suggest decreasing batch size or using tiling
- If no tiling and batch size =1: suggest tiling
- If tiling and batch size > 1: suggest decreasing tile size or batch size
- If tiling and batch size = 1: suggest decreasing tile size

veegalinova: Yes I like this, added these comments

jdeschamps (approved):

PR #949, opened 2026-05-12
Feature: Select validation in evenly spaced blocks

### LINKED ISSUE #813: Default number of validation patches may cause issues for small number of patches
## Problem

Automatic validation splitting was introduced in #789 and made the default if no validation data is provided. In the configuration we set the default number of validation patches to be 8. However if there is a very limited amount of training data this may cause problems, including errors raised about nan probabilities (needs to be investigated).

Additionally, we might want to introduce a better validation patch sampling strategy than completely random. If validation patches have a spacing of only one patch size this makes it very unlikely for the data in-between the validation patches to be sampled for training. Ideally, the validation patches should have a spacing greater than 2 patches, or they should be touching.

Currently the validation patch selection happens here:
https://github.com/CAREamics/careamics/blob/5a500182e18bfa4361de560c461ad4da43c0bd4d/src/careamics/dataset_ng/val_split.py#L67-L76

melisande-c:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: Randomly selecting validation patches could result in a configuration that reduces the likelyhood of some regions being selected for training. Additionally the validation might not accurately represent the statistics of the data.

For smallish image sizes selecting the validation patches badly could result in a reduction in the probability that some regions are selected for validation. This happens when the validation patches are too close too each other or the edge, making it unlikely to select the region between the validation patches for training. 

### Overview - what changed?

Added validation selection functionality that groups validation patches together in evenly spaced blocks so that there is always a gap greater than two patch widths between validation blocks, (and between the edges of the image and the blocks).

If this results in too many patches then validation patches are randomly removed, only if there removal does not result in a gap of 1, until the desired number of validation patches.

- Find how many validation patches there should be in each dimension.
- Find the best validation block size (and gap size) in each dimension
- Create the validation blocks
- Randomly remove patches until we have the desired number.

### New features or files

- `select_validation` function + helper functions

- `create_val_split` function -> replaced random val selection with block val selection

## How has this been tested?

- Added tests for each new function.
- Ran training with validation splitting and results still look good.
- Included demo script.

## Related Issues

- Resolves #813 

The included demo script produces the following results:

You can see how the random validation selection is effectively reducing the training data for the larger percentages of validation.

The new validation patch selection
block_validation_selection

The previous random validation selection
random_validation_selection

---

- [x] Pre-commit passes


melisande-c:
Also the effect on the patch distribution for 1 epoch of patches:

Block:
block_validation_selection_1_epoch

Random:
random_validation_selection_1_epoch

src/careamics/dataset/factory/val_split.py
```diff
@@ -98,6 +110,378 @@ def create_val_split(
...
+        dim_size: int = grid_shape[i]
+
+        # --- calculate the number of coordinates in each dimension
+        # try to have the patches distributed evenly in each dimension
+        ratios: list[float] = [  # ratios between dimensions
+            grid_shape[i] / grid_shape[j]
+            for j in range(ndims)
+            if i != j and val_coords_1D[j] is None
+        ]
+        # remaining val patches for the dimensions not calculate yet
```
jdeschamps:
```suggestion
        # remaining val patches for the dimensions not calculated yet
```

jdeschamps (approved):
I did not have time to delve too much in the details, but looks good.

Let's improve the error message as mentioned in the chat, and see the benchmark results.

melisande-c:
Error message updated in [0ff607d](https://github.com/CAREamics/careamics/pull/949/commits/0ff607ded53c8da1754171790e13650d3ffc7372)

PR #956, opened 2026-05-18
fixed stitch prediction to take channel dim from predictions not original (input) data

### LINKED ISSUE #948: Stitching tiled predictions fails when prediction channels are different from input's channel
```bash
```

mese79:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: This will fix stitching tiled predictions back into original input shape *but* getting the `channel` dimension from the predictions and not the original data.

In some cases, like `microSplit` or `lambdaSplit`, the predicted output might have a different channel dimension than the input data. This PR will fix the issue (See #948) .

### Overview - what changed?

In `stitch_single_prediction` and `stitch_single_sample`, the `data_shape` will have the `channel` dimension of the predictions and not the original input data.
```python
num_tiles = tiles[0].data_shape[0]
out_channels = tiles[0].data.shape[0]
data_shape = (num_tiles, out_channels, *tiles[0].data_shape[2:])
```

- `stitch_single_prediction` and `stitch_single_sample` functions in `src/careamics/lightning/prediction/stitch_prediction.py` are modified.

## How has this been tested?
A test is added in `tests/lightning/prediction/test_ng_stitch_tiled_prediction.py`.  
All other `prediction` tests are passed as well.

## Related Issues

- Resolves #948 

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/lightning/prediction/stitch_prediction.py
+++ b/src/careamics/lightning/prediction/stitch_prediction.py
@@ -103,7 +103,9 @@ def stitch_single_prediction(
     numpy.ndarray
         Full image, with dimensions SC(Z)YX.
     """
-    data_shape = tiles[0].data_shape
+    num_tiles = tiles[0].data_shape[0]
+    out_channels = tiles[0].data.shape[0]
+    data_shape = (num_tiles, out_channels, *tiles[0].data_shape[2:])
     predicted_image = np.zeros(data_shape, dtype=np.float32)
 
     # stitch each sample separately
@@ -148,7 +150,9 @@ def stitch_single_sample(
     numpy.ndarray
         Full sample, with dimensions C(Z)YX.
     """
-    data_shape = tiles[0].data_shape  # SC(Z)YX
+    num_tiles = tiles[0].data_shape[0]
+    out_channels = tiles[0].data.shape[0]
+    data_shape = (num_tiles, out_channels, *tiles[0].data_shape[2:])  # SC(Z)YX
     predicted_sample = np.zeros(data_shape[1:], dtype=np.float32)
 
     for tile in tiles:
--- a/tests/lightning/prediction/test_ng_stitch_tiled_prediction.py
+++ b/tests/lightning/prediction/test_ng_stitch_tiled_prediction.py
@@ -1,0 +1,99 @@
+import numpy as np
+import pytest
+from numpy.typing import NDArray
+
+from careamics.config.data import (
+    DataConfig,
+    TiledPatchingConfig,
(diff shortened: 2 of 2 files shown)


jdeschamps (changes requested):
There are more places where this will be an issue:

- `convert_prediction.py`
- `zarr_tile_write_strategy.py`

The best would be that the dataset would write the `original_data_shape` with the correct channel number, unfortunately this is not known from the configuration. Currently, in the UNet case, a model constraints is passed and it could hold the number of output channels. So this could be used.

However, it can be `None` (Lightning API, LVAE until we implement model constraints for the LVAE).

I see two ways:
- implement the fix everywhere (annoying, harder to maintain)
- implement n_output channels in model constraints, LVAE model constraints, update `original_data_size`

@mese79 @melisande-c

mese79:
I tried another approach and it seems working (tests were passed). I suggest to update the `ImageRegionData` class like this:
```python
@dataclass
class ImageRegionData(Generic[RegionSpecs]):
    data: NDArray
    source: Union[str, Literal["array"]]
    dtype: str
    axes: str
    internal_data_shape: Sequence[int]
    original_data_shape: Sequence[int]
    region_spec: RegionSpecs
    additional_metadata: dict[str, Any]

    @property
    def data_shape(self) -> Sequence[int | Tensor]:
        num_patches: int | Tensor = self.internal_data_shape[0]
        num_channels: int | Tensor
        if isinstance(num_patches, Tensor):
            # batch mode
            num_channels = torch.tensor([self.data.shape[1]] * len(num_patches))
        else:
            num_channels = self.data.shape[0]

        return [num_patches, num_channels, *self.internal_data_shape[2:]]  # SC(Z)YX
```
The `data_shape` will be a read-only property and return the correct shape with respect to carrying data either in single instance or in *batch* modes. The `internal_data_shape` holds the ordered original data shape. 
Then we need to register a proper `collate_fn` for this class once:
```python
from torch.utils.data._utils.collate import default_collate, default_collate_fn_map
...
def collate_region_data(batch, collate_fn_map=None):
    # collate each field of ImageRegionData class as a batch.
    field_names = [f.name for f in fields(ImageRegionData)]
    return ImageRegionData(**{
        name: default_collate([getattr(region_data, name) for region_data in batch])
        for name in field_names
    })  # type: ignore


# register the collate function handler
default_collate_fn_map[ImageRegionData] = collate_region_data
```

jdeschamps:
Superseded by https://github.com/CAREamics/careamics/pull/993

PR #975, opened 2026-06-12
CI(docs): Reduce training time for testing snippet examples

melisande-c:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: CI for docs snippets was taking a long time because some snippets are training on a lot of data. This PR reduces the training data (without changing the snippets shown in the docs), with either, `limit_train_batches`, reducing the epochs or, swapping the input data with smaller dummy data.

All the changes happen between snippets, so they are not visible on the website.

In `docs/current/getting_started.py` it still trains for 30 epochs, once for n2v and once for care, but this all happens in one snippet so the configuration cannot be changed before training.

- [x] Pre-commit passes

### DIFF
--- a/docs/current/careamist_training.py
+++ b/docs/current/careamist_training.py
@@ -1,18 +1,22 @@
 #!/usr/bin/env python
 import numpy as np
-from careamics.config.factories import create_n2v_config, create_care_config
-from careamics.dataset.factory import ReadFuncLoading
+from careamics.config.factories import (
+    create_advanced_n2v_config,
+    create_advanced_care_config,
+)
 
 # create a configuration
-config_n2v = create_n2v_config(
+config_n2v = create_advanced_n2v_config(
     experiment_name="n2v",
     data_type="array",
     axes="YX",
     patch_size=[64, 64],
     batch_size=8,
     num_epochs=1,
+    n_val_patches=2,
+    trainer_params={"limit_train_batches": 1},
 )
-config_care = create_care_config(
+config_care = create_advanced_care_config(
     experiment_name="care",
     data_type="array",
     axes="YX",
@@ -20,6 +24,7 @@ config_care = create_care_config(
     batch_size=8,
     num_epochs=1,
     n_val_patches=2,
+    trainer_params={"limit_train_batches": 1},
 )
 config = config_n2v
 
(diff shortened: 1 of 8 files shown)


docs/current/careamist_training.py
```diff
@@ -1,25 +1,30 @@
...
-config_care = create_care_config(
+config_care = create_advanced_care_config(
     experiment_name="care",
     data_type="array",
     axes="YX",
     patch_size=[64, 64],
     batch_size=8,
     num_epochs=1,
     n_val_patches=2,
+    trainer_params={"limit_train_batches": 1},
```
jdeschamps: Why not using `num_steps`?

docs/current/getting_started.py
```diff
@@ -3,7 +3,8 @@
 # %%
 import numpy as np
 
-train_data = np.random.randint(0, 255, (512, 512)).astype(np.float32)
+# cannot be 1 patch because of n_val_patches
+train_data = np.random.randint(0, 255, (64 * 4, 64 * 8)).astype(np.float32)
```
jdeschamps, line 7: Can you try to use two snippets one after the other and inject a configuration in between?

jdeschamps (approved):

docs/current/careamist_training.py
melisande-c: oh I guess I forgot that we added that

docs/current/getting_started.py
melisande-c: I will have to edit the getting started mark down file

melisande-c: but ok can do

melisande-c: mm it might look a bit weird splitting the snippet in two, currently the snippets CI is running quicker than the pytest CIs, so maybe its ok to leave it like this.
Screenshot 2026-06-12 at 16 00 59

docs/current/careamist_training.py
melisande-c: done in [0f7ba45](https://github.com/CAREamics/careamics/pull/975/commits/0f7ba458e3e136c46f27a44fa62c7ab574d1f2fd)

PR #977, opened 2026-06-22
Fix(Normalization): Denormalize with target stats when present

### LINKED ISSUE #967: Normalization: target stats not used for denomalization
## Describe the bug

The denormalization method is used to transform the distribution of the model's output to the distribution of the target data. However in both the `MeanStdNormalization` and `RangeNormalization` the input stats are used. For methods that have a different number of input and output channels this will result in an error.

https://github.com/CAREamics/careamics/blob/a9033f51acff3c60a8725cc35fce977791160019/src/careamics/dataset/normalization/mean_std_normalization.py#L176-L179

## Suggestion

1. Have two methods `denormalize_input` and `denormalize_target`.

OR

2. If targets don't exist, use input stats otherwise use target stats for denormalization.

parallelArchitect:
## Disclaimer

- [ ] I am an AI agent.
- [X] I have used AI and I thoroughly reviewed every line.

## Description

> [!NOTE]  
> **tldr**: denormalize() now correctly uses target statistics instead of input statistics when both are provided.

Both denormalize() methods unconditionally used input statistics to reverse-normalize the model's output. The model predicts values in the target distribution, not the input distribution. When input and target have different channel counts, this causes a ValueError in broadcast_stats. When channel counts match, the bug is silent but produces wrong values — the output is scaled to the input range instead of the target range.

### Overview - what changed?

denormalize() in both classes now checks for target stats first and falls back to input stats only when none were provided, preserving existing behavior.

Added a conditional check at the top of each denormalize() method — if target_means/target_stds (or target_mins/target_maxes for RangeNormalization) are not None, use them; otherwise fall back to input stats. No changes to the normalization forward pass or any other methods. 

## Changes Made

 denormalize() in src/careamics/dataset/normalization/mean_std_normalization.py
denormalize() in src/careamics/dataset/normalization/range_normalization.py

### New features or files
- test_denormalize_uses_target_stats_not_input_stats in tests/dataset/normalization/test_mean_std_normalization.py
test_denormalize_uses_target_range_not_input_range in tests/dataset/normalization/test_minmax_normalization.py
...

## How has this been tested?

Both regression tests reproduce the exact failure case from the issue (mismatched input/target channel counts), confirm the fix produces correct values, and verify backward compatibility when no target stats are provided. All 63 normalization tests pass locally.

## Related Issues

- Resolves #967

## Breaking changes

None — existing behavior preserved when no target stats are provided. 

---

- [X]  Code builds and passes tests locally, including doctests
- [X] New tests have been added (for bug fixes/features)
- [X] Pre-commit passes

### DIFF
--- a/src/careamics/dataset/normalization/mean_std_normalization.py
+++ b/src/careamics/dataset/normalization/mean_std_normalization.py
@@ -157,10 +157,9 @@ class MeanStdNormalization(Normalization):
     def denormalize(self, patch: torch.Tensor) -> torch.Tensor:
         """Reverse the normalization operation for a batch of patches.
 
-        Uses target statistics when available, since the patch passed in
-        is the model's output (in the target's distribution), not the
-        original input. Falls back to input statistics when no target
-        statistics were provided, preserving prior behavior.
+        The data is denormalized using the target statistics, when available. When no
+        target statistics are available, which is the case for self-supervised algorithms
+        such as N2V, the data is denormalized using the input statistics.
 
         Parameters
         ----------
--- a/src/careamics/dataset/normalization/mean_std_normalization.py
+++ b/src/careamics/dataset/normalization/mean_std_normalization.py
@@ -157,9 +157,10 @@ class MeanStdNormalization(Normalization):
     def denormalize(self, patch: torch.Tensor) -> torch.Tensor:
         """Reverse the normalization operation for a batch of patches.
 
-        The data is denormalized using the target statistics, when available. When no
-        target statistics are available, which is the case for self-supervised algorithms
-        such as N2V, the data is denormalized using the input statistics.
+        The data is denormalized using the target statistics, when available.
+        When no target statistics are available, which is the case for
+        self-supervised algorithms such as N2V, the data is denormalized
+        using the input statistics.
 
         Parameters
         ----------
--- a/src/careamics/dataset/normalization/mean_std_normalization.py
+++ b/src/careamics/dataset/normalization/mean_std_normalization.py
@@ -55,7 +55,6 @@ class MeanStdNormalization(Normalization):
         self.input_stds = input_stds
         self.target_means = target_means
         self.target_stds = target_stds
-
         self.eps = 1e-6
 
     def __call__(
@@ -87,6 +86,7 @@ class MeanStdNormalization(Normalization):
 
         means = reshape_stats(input_means, patch.ndim, channel_axis=0)
         stds = reshape_stats(input_stds, patch.ndim, channel_axis=0)
+
         norm_patch = self._apply_normalization(patch, means, stds)
 
         if target is None:
@@ -157,6 +157,11 @@ class MeanStdNormalization(Normalization):
     def denormalize(self, patch: torch.Tensor) -> torch.Tensor:
         """Reverse the normalization operation for a batch of patches.
 
+        Uses target statistics when available, since the patch passed in
+        is the model's output (in the target's distribution), not the
+        original input. Falls back to input statistics when no target
+        statistics were provided, preserving prior behavior.
+
         Parameters
         ----------
         patch : torch.Tensor
@@ -168,12 +173,15 @@ class MeanStdNormalization(Normalization):
             Transformed array.
         """
         n_channels = patch.shape[1]
-        input_means = broadcast_stats(self.input_means, n_channels, "input_means")
-        input_stds = broadcast_stats(self.input_stds, n_channels, "input_stds")
+        if self.target_means is not None and self.target_stds is not None:
+            means_list, stds_list = self.target_means, self.target_stds
+        else:
+            means_list, stds_list = self.input_means, self.input_stds
+        means_list = broadcast_stats(means_list, n_channels, "means")
+        stds_list = broadcast_stats(stds_list, n_channels, "stds")
 
         patch = patch.to(dtype=torch.float32)
 
-        means = reshape_stats(input_means, patch.ndim, channel_axis=1)
-        stds = reshape_stats(input_stds, patch.ndim, channel_axis=1)
-
+        means = reshape_stats(means_list, patch.ndim, channel_axis=1)
+        stds = reshape_stats(stds_list, patch.ndim, channel_axis=1)
         return self._apply_denormalization(patch, means, stds)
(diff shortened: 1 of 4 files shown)


src/careamics/dataset/normalization/mean_std_normalization.py
```diff
@@ -157,6 +157,11 @@ def _apply_denormalization(
     def denormalize(self, patch: torch.Tensor) -> torch.Tensor:
         """Reverse the normalization operation for a batch of patches.
 
+        Uses target statistics when available, since the patch passed in
+        is the model's output (in the target's distribution), not the
+        original input. Falls back to input statistics when no target
+        statistics were provided, preserving prior behavior.
```
melisande-c: Do not reference "prior behaviour" in documentation since this will not make sense to people reading it later, especially in the API reference on the website.

The algorithm N2V is self-supervised and the input statistics are the target statistics which is why we want this behaviour.

```suggestion
    The data is denormalized using the target statistics, when available. When no target
    statistics are available, which is the case for self-supervised algorithms such as
    N2V, the data is denormalized using the input statistics.
```

src/careamics/dataset/normalization/range_normalization.py
```diff
@@ -113,6 +102,11 @@ def __call__(
     def denormalize(self, patch: torch.Tensor) -> torch.Tensor:
         """Reverse the normalization operation for a batch of patches.
 
+        Uses target range when available, since the patch passed in is the
+        model's output (in the target's distribution), not the original input.
+        Falls back to input range when no target range was provided, preserving
+        prior behavior.
```
melisande-c: Again, do not mention "prior behaviour" in documentation. Replace this segment with something similar suggested in my previous [comment](https://github.com/CAREamics/careamics/pull/977/changes#r3498437862)

tests/dataset/normalization/test_mean_std_normalization.py
```diff
@@ -226,3 +226,48 @@ def test_global_stats_pools_across_channels():
...
     assert sample.data[0].mean() < 0.0
     assert sample.data[1].mean() > 0.0
+
+
+def test_denormalize_uses_target_stats_not_input_stats():
+    """Regression test for #967: denormalize() must use target stats when
+    they exist, not input stats. Real failure case: input and target have
+    a different number of channels (e.g. denoising task), so reusing
+    input's per-channel stats on the model's (target-shaped) output either
+    crashes or silently applies the wrong values."""
```
melisande-c: Do not mention the issue number or PR number in the test docstring or any of the comments. Let's not call this a regression test, we are just testing new behaviour, so let's rename it to something like `test_denormalize_with_target_stats`.

tests/dataset/normalization/test_mean_std_normalization.py
```diff
@@ -226,3 +226,48 @@ def test_global_stats_pools_across_channels():
...
+    )
+    dataset = create_dataset(config=config, inputs=[input_data], targets=[target_data])
+
+    # Real failure scenario from the issue: model output has a different
+    # channel count than the input. Force target stats to be multi-channel
+    # while input stats stay single-channel (global), the exact mismatch
+    # described in #967.
+    norm = dataset.normalization
+    norm.target_means = [150.0, 300.0]
+    norm.target_stds = [30.0, 50.0]
```
melisande-c: set the target stats when creating the configuration above, not here.

tests/dataset/normalization/test_mean_std_normalization.py
```diff
@@ -226,3 +226,48 @@ def test_global_stats_pools_across_channels():
...
+
+    # Simulate model output: batch=1, 2 channels (matches target, not input)
+    model_output = torch.zeros((1, 2, 8, 8))
```
melisande-c: can you make the model output `torch.ones`? So that the result should be `target_std` + `target_mean` not just the `target_mean`

tests/dataset/normalization/test_minmax_normalization.py
```diff
@@ -221,3 +221,40 @@ def test_global_stats_pools_across_channels():
     sample, *_ = dataset[0]
     assert np.isclose(sample.data[0].mean(), 0.0, atol=0.01)
     assert np.isclose(sample.data[1].mean(), 1.0, atol=0.01)
+
+
+def test_denormalize_uses_target_range_not_input_range():
+    """Regression test for #967: denormalize() must use target range when
+    it exists, not input range. Real failure case: input and target have
+    a different number of channels, so reusing input's per-channel range
+    on the model's (target-shaped) output gives silently wrong values."""
```
melisande-c: Again, let's make this docstring a simple one-liner to describe what the test does without mentioning the PR number or the issue number, and rename the test to `test_denormalize_with_target_range`.

tests/dataset/normalization/test_minmax_normalization.py
```diff
@@ -221,3 +221,40 @@ def test_global_stats_pools_across_channels():
...
+            "input_maxes": [100.0],
+            "target_mins": [0.0],
+            "target_maxes": [1000.0],
+        },
+    )
+    dataset = create_dataset(config=config, inputs=[input_data], targets=[target_data])
+
+    norm = dataset.normalization
+    norm.target_mins = [0.0, 0.0]
+    norm.target_maxes = [1000.0, 2000.0]
```
melisande-c: Again, set the target stats when creating the configuration instead of patching it in here.

melisande-c (changes requested):
Hi @parallelArchitect, thank you for your contribution, however next time please read our contribution guidelines which states that people outside the core team who wish to contribute should get in contact with us first, you can request to work on an existing issue by commenting on it and waiting for our response on how to proceed. 

I have requested some changes, mostly about documentation and the tests.

Thanks again!

parallelArchitect:
Thanks for the thorough review @melisande-c — addressed everything:

- Removed "prior behavior" language from both docstrings, used your suggested phrasing
- Renamed tests to `test_denormalize_with_target_stats` / `test_denormalize_with_target_range`, simplified docstrings, removed issue/PR references
- Moved target stats/range into config creation instead of patching after dataset creation
- Changed mean_std test's model output from zeros to ones so the assertion validates target_mean + target_std being applied, not just the mean

All normalization tests pass (63/63).

melisande-c:
Great thanks for the changes, can you fix the last pre-commit "line too long" error and then we are good to merge.

melisande-c (approved):

PR #985, opened 2026-06-29
Feat(norm): skip target normalization flag

### LINKED ISSUE #680: Next-Gen Dataset: Add option to skip target normalization
## Problem

For segmentation tasks, we have targets but might not want to normalize them (e.g., binary masks or integer labels). Currently, the normalization system assumes that if targets exist, they should be normalized with the same strategy as inputs. 

## Potential solution

Add a `normalize_target` flag normalization configs. When set to `False`, targets would be passed through unchanged and statistics computation would skip the target data entirely. As suggested in #655.

jdeschamps:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: For segmentation tasks it will be necessary to allow skipping target normalization, this PR adds such a mechanism.

Add a `skip_target` flag to all normalization.

Segmentation tasks will require to skip target normaloization to avoid hacky approaches such as setting target stats to identity transformations.

### Overview - what changed?

`skip_target` parameter added to the normalization configuration and implementations.

### New features or files

- `skip_target` added all normalization configurations, alongside a warning when the target stats are nonetheless set.
- `skip_target` is used in the normalization validation to avoid errors when the number of output channels may not fit the stats, even when the target normalzation is skipped.

## How has this been tested?

New tests.

## Related Issues

- Resolves #680

Note that the doc has not been explicitly updated with a section to skip target normalization because it is not clear where the segmentation will end up.

---

- [x] Pre-commit passes


tests/unit/dataset/normalization/_init__.py
veegalinova, line 1: Typo in file name

veegalinova (approved):
A new init file is misnamed, otherwise all good!

tests/unit/dataset/normalization/_init__.py
jdeschamps: Good catch, fixed in https://github.com/CAREamics/careamics/pull/985/changes/a9d6da64a17e1baaeb621ae0eb14441b52568fa4

PR #993, opened 2026-07-02
Feat(prediction): restore predictions with different channels than input

### LINKED ISSUE #948: Stitching tiled predictions fails when prediction channels are different from input's channel

### LINKED ISSUE #963: Allow input and ouput with different axes
## Problem

While most of our applications so far had inputs and targets with same axes, it is not always the case and can be blocking for the addition of new algorithms / tasks. For instance one of the CARE original experiment was ZYX -> YX (projection).  

The error is raised when reshaping the data in the image stacks:

```
line 243, in reshape_array
    transform = AxesTransform(original_axes, array.shape)
  File "<string>", line 5, in __init__
  File "/src/careamics/utils/reshape_array.py", line 53, in __post_init__
    raise ValueError(
    ...<3 lines>...
    )
ValueError: Axes 'YX' length (2) does not match shape (2, 128, 128) length (3).
```

Similarly to https://github.com/CAREamics/careamics/issues/948, targets with different axes will also create issues when reshaping before writing to disk (prediction conversion, zarr prediction writer).

jdeschamps:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: This PR removes fixes the error faced when restoring a prediction that has different channel dimension than the input.

Currently, any prediction with C dimension different from input will lead to an error during array reshaping. This PR solves the issue by engineering the array restoration to handle the channel differently. Because of the implementation of the Zarr tile writer, some modifications had to be carried out there as well.

It introduces the following:
- `target_axes` in ImageRegionData, set to `axes` if `target_axes` is `None`
- Zarr tile writing is only possible with canonical axes order (STCZYX)
- restoration functions have been refactored to avoid logic duplication
- restoration functions now take the `target_axes` (never `None`, can be equal to `axes`) as reference for restoration, current channel size is used
- Validation in data configuration enforces that if S/T are present in `axes`, so should they be in `target_axes`
- CAREamist now also allow to change `target_axes` to allow for them to match `axes` (e.g. no S)
- Zarr tile writing has been refactored, as first step to lower technical debt.

Multiclass segmentation, CARE, MicroSplit, LambdaSplit, all have different channel dimensions in input and output. Current issue raised by mismatching channel dimensions (https://github.com/CAREamics/careamics/issues/948) is currently blocking the implementation or support of these algorithms.

### Overview - what changed?

- CAREamist takes an optional `new_target_axes`
- Data configuration validates `target_axes` S/T axes against those in `axes`
- `ImageRegionData` propagate a `target_axes`, which may be equal to `axes` when the configuration has no `target_axes`
- Restoration takes the current channel size, and reshape the array according to `target_axes`
- Zarr tile writing is no longer supported for non-canonical axes order

To allow different C dimensions, I implemented a `RestoredAxesTransform` class as a counterpart to `AxesTransform`. This class allows transforming axes, shapes and stitch slices from the canonical SCZYX axes order into the original input order.

In addition, that class allows refactoring common logic between `restore_array` and `restore_tile` (now removed) into a single testable class.

Further refactoring of the tile manipulation logic from the Zarr tile writer into a single `ZarrTileHandler` class that holds a `RestoredAxesTransform` and handle all the tile restoration and indexing. In particular, it is necessary for the write strategy to know what the receiving array shape is (channel-adjusted) and what the new chunks/shards are. Previously, it was assumed that array shape and chunks/shards could just be carried over from the input.

### New features or files

- `_validate_axes` additional validation in `DataConfig`
- `target_axes` in `ImageRegionData`, set to `config.target_axes or config.axes` in every instantiation
- Refactored instantiation of `ImageRegionData` from model outputs into `ImageRegionData.from_model_output`
- `new_target_axes` in `CAREamist`
- `RestoredAxesTransform` in `reshape_array`: this class is responsible for all matters linked to restoring original axes order, that is to say reshaping an array or a tile, or a shape (e.g. chunks), and restoring axes and stitch slices order.
- `ZarrTileHandler` in `zarr_tile_write_strategy` wraps all the logic around interaction and restoration of the data.

- `ZarrTileWriteStrategy` has been slightly refactored to move some logic to the new `ZarrTileHandler` or to private functions.
- `AxesTransform`: shared validation logic with `RestoredAxesTransform` has been refactored in a private function.
- Some tests have been moved to `tests/integration` or `tests/e2e`
- Many tests have been adjusted to reflect the new behaviour

### Removed features or files

- `test_zarr_tile_write_strategy`: tests on non-canonical axes order have been removed.

## How has this been tested?

Additional tests have been implemented, others have been adjusted.

## Related Issues

- Resolves https://github.com/CAREamics/careamics/issues/948
- Resolves https://github.com/CAREamics/careamics/issues/963
- Closes https://github.com/CAREamics/careamics/pull/956

This is an alternative to https://github.com/CAREamics/careamics/pull/956 proposed solution (see comments in the PR rather than the code diff).

Zarr tile writing will need to be refactored further.

Additional things to potentially discuss:
- Renaming the axes transform, maybe systematically using "canonical" to refer to the reference order STCZYX.
- "restoring" the tile is still used in the Zarr tile writer because that allows ensuring correct channels, but the naming is a bit abusive since non-canonical axes order are not supported anymore.

## TODOs

- [x] Fix e2e and `tests/lightning/callbacks/prediction/test_zarr_tile_write_strategy.py` tests
- [ ] ZarrHandler tests
- [x] Docs update, document `new_target_axes`
- [ ] Tests for data config new validator

---

- [x] Pre-commit passes


jdeschamps:
On hold until target loading is merged (#1005 )

jdeschamps:
Continuing live discussion with @melisande-c: I still can't convince myself that for most users it will not be more confusing to have the following:

```python
cfg = create_config(
    ...,
    axes="YXC",
    target_axes="YX",
    n_channels_in=3,
    n_channels_out=1,
)

...

prediction, source = careamist.predict(...)

prediction[0].shape
>>> (128, 128, 1)
```

Rather than:

```python
prediction[0].shape
>>> (128, 128)
```

Hence the decision (after several back and forth commits) in this PR to remove any singleton "C" from restored prediction regardless of other parameters.

melisande-c:
My suggestion was alway that the prediction axes should be those of the target axes (if present)! Therefore in the above example the prediction should be restored to have the shape `"YX"` and have the shape `(128, 128)`. I haven't looked at the code yet to know if this would remove any complexity.

Also, for tasks where we go from `YX` -> `YXC`, e.g. segmentation, I think it is important to reshape to the target axes (either `YXC` or `CYX`) to make it easier for users to compare their ground truth annotations to the predictions. I can imagine that the position of the `C` axis in ground truth annotations will depend on the software used to create them.

(Equally, if the target originally has a singleton `C` axis I think that should remain present in the predictions to make it easier to compare ground truths to predictions at test time.)

jdeschamps:
Currently, this PR does not have yet the `target_axes`, waiting for the other PR to be merged. Then I will need to find a place to substitute/add it to the `ImageRegionData`. Then the idea was anyway to have the `target_axes` (if present) as reference for the order, otherwise the applicable logic/decision would be that of this PR.

melisande-c:
> Currently, this PR does not have yet the `target_axes`, waiting for the other PR to be merged. Then I will need to find a place to substitute/add it to the `ImageRegionData`. Then the idea was anyway to have the `target_axes` (if present) as reference for the order, otherwise the applicable logic/decision would be that of this PR.

Ah I see, I was confused because `target_axes` is present as an argument to `create_config` in the code snippet your previous comment.

But I am still against removing any singleton dimension if they were already present in the data.

jdeschamps:
> Ah I see, I was confused because `target_axes` is present as an argument to `create_config` in the code snippet your previous comment.
> 

Ah yes, too many PRs in parallel... No, this PR will have to be updated to make it use `target_axes`.

> But I am still against removing any singleton dimension if they were already present in the data.

To be fair, I don't remember now what happens if "CYX" has singleton channel as input (e.g. to N2V). I will double check.

jdeschamps:
## Use cases

Just to clarify the use-cases (assuming `target_axes` PR merged):

- CARE: model output and targets are strictly the same shape and dims, if they are different from `axes` then `target_axes` need to be specified. So `target_axes` needs to be the reference for restoring if it is specified, otherwise `axes` Previous restoring implementation can be used as is.
- Noise2Noise: C channel might differ, `target_axes` would be specified and would then be the same as CARE.
- LambdaSplit: There is no target, output has different `C` shape than `axes` but should keep same order. `original_data_shape` should be modified to accommodate new channel shape.

These cases tend to speak in favour of `target_axes | None` being the only thing required, and this PR being simplified. What about MSplit?

- MSplit? TODO
 
## Zarr restoring

Currently, this PR introduces a raised error when the restoring shape is not canonical (STCZYX order). While it makes no sense to write non-canonical order to Zarr, raising an error is not the right solution. We should just not restore the array and simply take the canonical `data_shape` (not `original_data_shape`), adjusted for channels, and use that to create the array.

## Splitting refactoring

This PR mixes refactoring and new features, that makes it hard to review and understand. If the refactoring is useful, it could be split into a different PR.

mese79:
There might be a problem with `denormalization` as well ([here](https://github.com/CAREamics/careamics/blob/d1425d3ed56ca54618652fa87d07d83c7da02413/src/careamics/dataset/normalization/mean_std_normalization.py#L195)):
```python
n_channels = patch.shape[1]
if self.target_means is not None and self.target_stds is not None:
    means_list, stds_list = self.target_means, self.target_stds
else:
    means_list, stds_list = self.input_means, self.input_stds
means_list = broadcast_stats(means_list, n_channels, "means")
stds_list = broadcast_stats(stds_list, n_channels, "stds")
```
If the number of channels in the target differs from the input, the target `denormalization` might fail.

jdeschamps:
> There might be a problem with `denormalization` as well ([here](https://github.com/CAREamics/careamics/blob/d1425d3ed56ca54618652fa87d07d83c7da02413/src/careamics/dataset/normalization/mean_std_normalization.py#L195)):
> 
> ```python
> n_channels = patch.shape[1]
> if self.target_means is not None and self.target_stds is not None:
>     means_list, stds_list = self.target_means, self.target_stds
> else:
>     means_list, stds_list = self.input_means, self.input_stds
> means_list = broadcast_stats(means_list, n_channels, "means")
> stds_list = broadcast_stats(stds_list, n_channels, "stds")
> ```
> 
> If the number of channels in the target differs from the input, the target `denormalization` might fail.

After a quick look, it seems it can only happen if:
1) There is no target (but output channels != input channels), which is only the case for LambdaSplit
2) Target normalization is skipped (and output channels != input channels)

Number 2 should be fixed, it was a blind spot in #985.

How is LambdaSplit dealing with denormalization? A solution would be to always turn off target normalization for LambdaSplit.

mese79:
> How is LambdaSplit dealing with denormalization? A solution would be to always turn off target normalization for LambdaSplit.

I think in LambdaSplit, the target denormalization is off by default. Fede was using scale-invariant metrics, so he can just use the raw model output. Also, not sure if we could always have the target stats, since LambdaSplit is unsupervised.

mese79:
> How is LambdaSplit dealing with denormalization? A solution would be to always turn off target normalization for LambdaSplit.

I think in LambdaSplit, the target denormalization is off by default. Fede was using scale-invariant metrics, so he can just use the raw model output. Also, not sure if we could always have the target stats, since LambdaSplit is unsupervised.

src/careamics/lightning/callbacks/prediction/zarr_tile_write_strategy.py
```diff
@@ -1,39 +1,218 @@
...
+    """Coordinates where to crop the tile."""
+
+    stitch_coords: Sequence[int]
+    """Coordinates in the array for stitching the tile."""
+
+    sample_idx: int
+    """Sample index of the tile."""
+
+    tile: NDArray
+    """Tile data."""
```
melisande-c: why are these class attributes? If more than one instance of the class is created these will change for every instance

src/careamics/lightning/callbacks/prediction/zarr_tile_write_strategy.py
```diff
@@ -76,6 +255,58 @@ def _add_output_key(dirpath: Path, path: str | Path) -> Path:
     return dirpath / new_name
 
 
+def _decipher_destination(
```
melisande-c: I feel like "decipher" is the wrong word because we are not decoding anything, maybe "get" is more broad

melisande-c (commented):
Seems like everything is ok apart from my comment about class attribute, although I did find trying to follow the data flow in `RestoredAxesTransform` kind of confusing so I might have missed something.

Also, did you test what happens when using CZI input with the axes SCTYX for 3D with T?

src/careamics/lightning/callbacks/prediction/zarr_tile_write_strategy.py
jdeschamps: I probably started from a dataclass. That being said, I don't think these are class attributes since they are just type annotations and set in `__init__`:

```python
from careamics.lightning.callbacks.prediction.zarr_tile_write_strategy import ZarrTileHandler
from careamics.dataset.image_region_data import ImageRegionData
from careamics.dataset.patching import TileSpecs
import numpy as np

z1 = ZarrTileHandler(
    ImageRegionData(
        data=np.ones((1, 2, 3)),
        source="ahahah",
        data_shape=(1, 2, 3),
        dtype="float",
        axes="YXC",
        target_axes="CYX",
        original_data_shape=(2, 3, 1),
        region_spec=TileSpecs(
            data_idx=0,
            sample_idx=0,
            coords=(0, 0),
            patch_size=(8, 8),
            crop_coords=(0,0),
            crop_size=(4, 4),
            stitch_coords=(0,0),
            total_tiles=2,
        ),
        additional_metadata={}
    )
)
z2 = ZarrTileHandler(
    ImageRegionData(
        data=np.ones((1, 2, 3)),
        source="glaaaaaah",
        data_shape=(1, 2, 3),
        dtype="float",
        axes="YXC",
        target_axes="CYX",
        original_data_shape=(2, 3, 1),
        region_spec=TileSpecs(
            data_idx=0,
            sample_idx=28,
            coords=(0, 0),
            patch_size=(8, 8),
            crop_coords=(0,0),
            crop_size=(4, 4),
            stitch_coords=(0,0),
            total_tiles=2,
        ),
        additional_metadata={}
    )
)

z1.sample_idx = 4

print(z1.sample_idx)
>> 4

print(z2.sample_idx)
 >> 28
 ```
 
 But I agree that it is totally unnecessary...

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/993/changes/0556fa67a363937810b5834b085c09f15e9c9bba

src/careamics/lightning/callbacks/prediction/zarr_tile_write_strategy.py
jdeschamps: Makes sense!

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/993/changes/1404144c207043ee3baa0fea49ddbef1c428ec5b

src/careamics/lightning/callbacks/prediction/zarr_tile_write_strategy.py
melisande-c: oh oops, I was too tired reading that I guess, type annotations are fine

jdeschamps: I honestly had doubts as well once you pointed it out...

jdeschamps:
> Also, did you test what happens when using CZI input with the axes SCTYX for 3D with T?

Testing axes restoration for CZI with T as 3rd dimension, or a pipeline with different input/target axes?

[EDIT]: I had an explicit flag for "SCTYX for 3D with T" that would raise an error upon restoring. Removing that now.

jdeschamps:
Changes since the review:

- https://github.com/CAREamics/careamics/pull/993/changes/e710eec2c5ae16680872176f2229734397cb3a0d: fix doc
- https://github.com/CAREamics/careamics/pull/993/changes/3df845889d206c034d3f4815d24c84c79de05a4c: rename a function because it was one letter away from another one (mode vs model)
- https://github.com/CAREamics/careamics/pull/993/changes/050807ba5b15c801e0745e26d385d1991096bc91: remove failure upon "SCTYX" CZI input, and implemented a work around by changing axes/target_axes to replace "T" by "Z".

I still want to implement some more tests.

jdeschamps:
Actually the `ZarrTileHandler` class is simply a wrapper, I don't really see what to test there. So I would consider this PR done.

I will open an issue for further refactoring of the reshape array classes for clarity.

melisande-c (approved):

PR #996, opened 2026-07-03
Fix (prediction): file naming truncation and overwriting 

### LINKED ISSUE #991: Bug: possible file name truncation and overwriting during prediction to disk
## Describe the bug

> With careamist.predict_to_disk now, I was getting a FileExistsError because all of my prediction images are being saved with the same name. The source images all have the same the file name header (the end of the file name differentiates them) so I think the names were getting truncated and that's what was causing the error. I changed the names to just be numbers and that fixed it.

Reported by a user [here](https://forum.image.sc/t/how-to-save-view-prediction-images-using-careamics/121667/4).

diyasrivas:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: 
Prevents `predict_to_disk` from truncating source file names

When writing predictions to disk, the output file name is derived from the source file name. For source files that shared a common prefix and only differed in a dotted suffix, the differentiating part was being stripped, so multiple predictions resolved to the same output name and writing failed.

The output name was built in `create_write_file_path` using `Path.stem` followed by `Path.with_suffix()`. Both of these strip a trailing dotted segment, so any dot in the meaningful part of the name was lost.

### Overview - what changed?

`create_write_file_path` now only strips the source extension when it is a recognized input extension, and otherwise keeps the full name. It no longer uses `with_suffix`, so interior dots are preserved and outputs remain unique.

Added a `KNOWN_INPUT_EXTENSIONS` set (`.tif`, `.tiff`, `.czi`, `.zarr`, matching `SupportedData`). In `create_write_file_path`, the source extension is stripped via `Path.stem` only when `file_path.suffix` is one of these; otherwise the full `file_path.name` is kept. The write extension is then appended by string concatenation rather than `with_suffix`, which avoids stripping a second dotted segment. This keeps the common single-extension case unchanged while preserving any interior dots.

- `create_write_file_path` in `src/careamics/lightning/callbacks/prediction/file_path_utils.py`: only strips recognized input extension and no longer uses `with_suffix`; added the `KNOWN_INPUT_EXTENSIONS` constant.

- `tests/lightning/callbacks/prediction/test_file_path_utils.py` : updated `test_create_write_file_path` to use a recognized extension (the previous version asserted an unrecognized extension would be swapped).

## How has this been tested?

Ran `test_file_path_utils.py` locally which passed. Verified against the reported failure cases so source names such as `experiment.0001`/`experiment.0002` and `cells.pos0.tif`/`cells.pos1.tif` now produce distinct output names instead of colliding, while ordinary single-extension names (`scan_0001.tif` → `scan_0001.tiff`) are unchanged.

## Related Issues

- Resolves #991 

## Breaking changes

None

### DIFF
--- a/src/careamics/lightning/callbacks/prediction/file_path_utils.py
+++ b/src/careamics/lightning/callbacks/prediction/file_path_utils.py
@@ -2,9 +2,6 @@
 
 from pathlib import Path
 
-# Extensions that CAREamics recognises as input image files (see `SupportedData`).
-KNOWN_INPUT_EXTENSIONS = frozenset({".tif", ".tiff", ".czi", ".zarr"})
-
 
 def create_write_file_path(
     dirpath: Path, file_path: Path, write_extension: str, postfix: str = ""
@@ -14,11 +11,6 @@ def create_write_file_path(
 
     Takes the original file path, changes the directory to `dirpath` and changes
     the extension to `write_extension`.
-
-    Only a recognised input extension (see `KNOWN_INPUT_EXTENSIONS`) is replaced;
-    any other dotted segments in the name are preserved. This avoids truncating
-    names such as ``experiment.0001`` or ``cells.pos0.tif``, which would otherwise
-    collide and raise a ``FileExistsError`` when written.
 
     Parameters
     ----------
@@ -37,9 +29,5 @@ def create_write_file_path(
         The output file path.
     """
     file_path = Path(file_path)  # as a guard against str input
-    if file_path.suffix.lower() in KNOWN_INPUT_EXTENSIONS:
-        stem = file_path.stem
-    else:
-        stem = file_path.name
-    file_name = f"{stem}{postfix}{write_extension}"
+    file_name = f"{file_path.stem}{postfix}{write_extension}"
     return dirpath / file_name
--- a/src/careamics/lightning/callbacks/prediction/file_path_utils.py
+++ b/src/careamics/lightning/callbacks/prediction/file_path_utils.py
@@ -12,7 +12,7 @@ def common_source_base(sources: Sequence[Path | str]) -> Path | None:
     Parameters
     ----------
     sources : sequence of pathlib.Path or str
-        Source paths of the predictions. Array sources (``"array"``) are ignored.
+        Source paths of the predictions. Array sources (`"array"`) are ignored.
 
     Returns
     -------
--- a/src/careamics/lightning/callbacks/prediction/file_path_utils.py
+++ b/src/careamics/lightning/callbacks/prediction/file_path_utils.py
@@ -2,6 +2,9 @@
 
 from pathlib import Path
 
+# Extensions that CAREamics recognises as input image files (see `SupportedData`).
+KNOWN_INPUT_EXTENSIONS = frozenset({".tif", ".tiff", ".czi", ".zarr"})
+
 
 def create_write_file_path(
     dirpath: Path, file_path: Path, write_extension: str, postfix: str = ""
@@ -11,6 +14,11 @@ def create_write_file_path(
 
     Takes the original file path, changes the directory to `dirpath` and changes
     the extension to `write_extension`.
+
+    Only a recognised input extension (see `KNOWN_INPUT_EXTENSIONS`) is replaced;
+    any other dotted segments in the name are preserved. This avoids truncating
+    names such as ``experiment.0001`` or ``cells.pos0.tif``, which would otherwise
+    collide and raise a ``FileExistsError`` when written.
 
     Parameters
     ----------
@@ -29,6 +37,9 @@ def create_write_file_path(
         The output file path.
     """
     file_path = Path(file_path)  # as a guard against str input
-    file_name = f"{file_path.stem}{postfix}"
-    file_path = dirpath / Path(file_name).with_suffix(write_extension)
-    return file_path
+    if file_path.suffix.lower() in KNOWN_INPUT_EXTENSIONS:
+        stem = file_path.stem
+    else:
+        stem = file_path.name
+    file_name = f"{stem}{postfix}{write_extension}"
+    return dirpath / file_name
--- a/src/careamics/lightning/callbacks/prediction/file_path_utils.py
+++ b/src/careamics/lightning/callbacks/prediction/file_path_utils.py
@@ -1,16 +1,47 @@
 """Module containing file path utilities for `WriteStrategy` to use."""
 
+import os
+from collections.abc import Sequence
 from pathlib import Path
 
 
+def common_source_base(sources: Sequence[Path | str]) -> Path | None:
+    """
+    Find the common parent directory of the source paths.
+
+    Parameters
+    ----------
+    sources : sequence of pathlib.Path or str
+        Source paths of the predictions. Array sources (``"array"``) are ignored.
+
+    Returns
+    -------
+    Path or None
+        The common parent, or None if it cannot be determined (fewer than two file
+        sources, or no shared base such as mixed absolute and relative paths).
+    """
+    paths = [Path(source) for source in sources if str(source) != "array"]
+    if len(paths) < 2:
+        return None
+    try:
+        return Path(os.path.commonpath([str(path) for path in paths]))
+    except ValueError:
+        # paths share no common base (e.g. mixing absolute and relative paths)
+        return None
+
+
 def create_write_file_path(
-    dirpath: Path, file_path: Path, write_extension: str, postfix: str = ""
+    dirpath: Path,
+    file_path: Path,
+    write_extension: str,
+    postfix: str = "",
+    source_base: Path | None = None,
 ) -> Path:
     """
     Create the file name for the output file.
 
-    Takes the original file path, changes the directory to `dirpath` and changes
-    the extension to `write_extension`.
+    If `source_base` is given, the source path relative to it is preserved under
+    `dirpath`, keeping outputs unique for same-named files in different directories.
 
     Parameters
     ----------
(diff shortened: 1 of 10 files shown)


src/careamics/lightning/callbacks/prediction/file_path_utils.py
```diff
@@ -12,6 +15,11 @@ def create_write_file_path(
     Takes the original file path, changes the directory to `dirpath` and changes
     the extension to `write_extension`.
 
+    Only a recognised input extension (see `KNOWN_INPUT_EXTENSIONS`) is replaced;
+    any other dotted segments in the name are preserved. This avoids truncating
+    names such as ``experiment.0001`` or ``cells.pos0.tif``, which would otherwise
+    collide and raise a ``FileExistsError`` when written.
+
```
jdeschamps, line 45: Can we remove this comment? Claude likes useless comments.

src/careamics/lightning/callbacks/prediction/file_path_utils.py
```diff
@@ -29,6 +37,9 @@ def create_write_file_path(
         The output file path.
     """
     file_path = Path(file_path)  # as a guard against str input
-    file_name = f"{file_path.stem}{postfix}"
-    file_path = dirpath / Path(file_name).with_suffix(write_extension)
-    return file_path
+    if file_path.suffix.lower() in KNOWN_INPUT_EXTENSIONS:
+        stem = file_path.stem
+    else:
+        stem = file_path.name
```
jdeschamps: I don't understand why we can't just do:

```suggestion
    stem = file_path.stem
```

The problem with `.stem.with_suffix()` is that the `stem` has a `.` in there, but `f"{stem}{post_fix}{write_extension}"` should work.

The one would need to remove the known extension dict. Also the source might be a custom, unknown, extension.

jdeschamps (changes requested):
Thanks for taking care of that! I think it can be even simpler, see my comments.

melisande-c:
Not related to the original request, but I guess we do now support passing individual file paths, which may be in different directories but have the same name, which also would not be supported.

jdeschamps:
> Not related to the original request, but I guess we do now support passing individual file paths, which may be in different directories but have the same name, which also would not be supported.

That could be taken care of in this PR, if you feel like it @diyasrivas. Otherwise we need to open an issue.

diyasrivas:
> but I guess we do now support passing individual file paths, which may be in different directories but have the same name, which also would not be supported.

I am just going to implement it here

src/careamics/lightning/callbacks/prediction/file_path_utils.py
diyasrivas: removed

diyasrivas:
> Not related to the original request, but I guess we do now support passing individual file paths, which may be in different directories but have the same name, which also would not be supported.

I updated it so predictions actually get written properly instead of colliding. Each prediction now gets saved to a path relative to the common parent folder of all the sources, so files keep their folder structure and same-named files don't overwrite each other.

For example:

["dir1/image.tif", "dir2/image.tif"]
    → pred/dir1/image.tiff
    → pred/dir2/image.tiff

["/abs/dir1/subdir/image.tif", "/abs/dir1/image.tif", "/abs/dir2/image.tif"]
    → pred/dir1/subdir/image.tiff
    → pred/dir1/image.tiff
    → pred/dir2/image.tiff

And if you just point it at a single folder of files, nothing changes

src/careamics/lightning/callbacks/prediction/file_path_utils.py
```diff
@@ -1,16 +1,47 @@
...
 
 
+def common_source_base(sources: Sequence[Path | str]) -> Path | None:
+    """
+    Find the common parent directory of the source paths.
+
+    Parameters
+    ----------
+    sources : sequence of pathlib.Path or str
+        Source paths of the predictions. Array sources (``"array"``) are ignored.
```
jdeschamps:
```suggestion
        Source paths of the predictions. Array sources (`"array"`) are ignored.
```

src/careamics/lightning/callbacks/prediction/image_write_strategy.py
```diff
@@ -68,6 +68,9 @@ def __init__(
 
         self.image_cache: dict[int, list[ImageRegionData]] = defaultdict(list)
 
+        # common parent of the sources, used to preserve their directory structure
+        self.source_base: Path | None = None
```
jdeschamps, line 76: It seems to me that it will always be `None` here, so there is no point in having this variable. Did I miss something?

jdeschamps: Oh, it is modified from outside the class. I don't like rogue modifications of attributes. It would be clearer if it has a setter `set_common_path`, maybe even defined in the `Protocol`

src/careamics/lightning/callbacks/prediction/image_write_strategy.py
```diff
@@ -68,6 +68,9 @@ def __init__(
 
         self.image_cache: dict[int, list[ImageRegionData]] = defaultdict(list)
 
+        # common parent of the sources, used to preserve their directory structure
```
jdeschamps:
```suggestion
        # common parent of the sources, used to preserve their directory structure
        # modified by the prediction writer callback
```

src/careamics/lightning/callbacks/prediction/tile_write_strategy.py
```diff
@@ -74,6 +74,9 @@ def __init__(
         # where tiles will be cached until a whole image has been predicted
         self.tile_cache: dict[int, list[ImageRegionData]] = defaultdict(list)
 
+        # common parent of the sources, used to preserve their directory structure
```
jdeschamps:
```suggestion
        # common parent of the sources, used to preserve their directory structure
        # modified by the prediction writer callback
```

jdeschamps (changes requested):

src/careamics/lightning/callbacks/prediction/zarr_tile_write_strategy.py
```diff
@@ -89,6 +89,20 @@ def __init__(self) -> None:
...
+        This strategy writes all predictions into a single zarr store rather than as
+        individual files, so there is no per-source directory structure to preserve
+        and `source_base` is ignored.
+
+        Parameters
+        ----------
+        source_base : pathlib.Path or None
+            Common parent of all prediction sources; ignored by this strategy.
+        """
+
```
jdeschamps, line 102: Sorry, just to be a bit more pythonic

```suggestion
    def set_source_base(self, source_base: Path | None) -> None:
        """
        No-op.

        Parameters
        ----------
        source_base : pathlib.Path or None
            Ignored.
        """
        pass

```

jdeschamps:
Seems like there is an issue coming up in the tests:

```python
 src/careamics/lightning/callbacks/prediction/prediction_writer_callback.py:147: in setup
    source_paths = getattr(trainer.datamodule, "predict_source_paths", [])
                           ^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Mock spec='Trainer' id='5683250864'>, name = 'datamodule'

    def __getattr__(self, name):
        if name in {'_mock_methods', '_mock_unsafe'}:
            raise AttributeError(name)
        elif self._mock_methods is not None:
            if name not in self._mock_methods or name in _all_magics:
>               raise AttributeError("Mock object has no attribute %r" % name)
E               AttributeError: Mock object has no attribute 'datamodule'. Did you mean: 'init_module'?

```

src/careamics/lightning/callbacks/prediction/file_path_utils.py
diyasrivas: yeah agreed because `with_suffix()` asks pathlib to guess where the extension starts, which is exactly what broke (any dot in the name chops the wrong part off). Building the name from stem + postfix + extension doesn't guess at all and dropping the dict means unknown/custom extensions work for free instead of us maintaining a list of formats we happen to know about.

The file changed a bit since (added source_base to preserve the source dir structure so same-named files in different folders don't collide), but that only affects which path the stem comes from, the name construction concept is still there which you can see in the latest version https://github.com/CAREamics/careamics/blob/6bfdd6494a15589affa03c038c3a4134ba181309/src/careamics/lightning/callbacks/prediction/file_path_utils.py

diyasrivas:
> Seems like there is an issue coming up in the tests:
> 
> ```python
>  src/careamics/lightning/callbacks/prediction/prediction_writer_callback.py:147: in setup
>     source_paths = getattr(trainer.datamodule, "predict_source_paths", [])
>                            ^^^^^^^^^^^^^^^^^^
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> self = <Mock spec='Trainer' id='5683250864'>, name = 'datamodule'
> 
>     def __getattr__(self, name):
>         if name in {'_mock_methods', '_mock_unsafe'}:
>             raise AttributeError(name)
>         elif self._mock_methods is not None:
>             if name not in self._mock_methods or name in _all_magics:
> >               raise AttributeError("Mock object has no attribute %r" % name)
> E               AttributeError: Mock object has no attribute 'datamodule'. Did you mean: 'init_module'?
> ```

the latest commit should fix that, Lightning sets datamodule at runtime, so it's not there and the mock refuses to hand it out. The mock just wasn't a complete stand-in for a real trainer once setup started reading .datamodule before, so now the test does:

```python
 trainer = mocker.Mock(spec=Trainer)
 trainer.datamodule = None
```

`None` mimics predicting straight from dataloaders with no datamodule which would fall back to [] and `common_source_base([])` returns `None` - and now everything passes..

jdeschamps (approved):

PR #999, opened 2026-07-06
Fix(checkpoints): allow checkpoints exported after prediction to be loaded

### LINKED ISSUE #939: Saving checkpoint after prediction wouldn't save the data config in `datamodule_hyper_parameters`
The lightning `trainer` has a `datamodule` property which will set on `train` and `predict` methods by the passed datamodule:
`trainer.fit(model, datamodule=data_module)`
or 
`trainer.predict(model, datamodule=inf_data_module)`
When saving the checkpoint it will save data config with respect to the **current `datamodule`** in the trainer. In `careamics`, the data config will be saved in `hparams` *only* for train/validation: [here](https://github.com/CAREamics/careamics/blob/0d3d6c545658a1b9c52c6ce95b0eea529ca29d1f/src/careamics/lightning/data/data_module.py#L296).  

A basic example to reproduce the issue (works with `dev/v0.2`):
```python
from pathlib import Path

import numpy as np
import torch

from careamics.config.factories import create_advanced_n2v_config
from careamics.lightning import (
    CareamicsDataModule,
    N2VModule,
)


my_path = Path(".")

    experiment_name="na",
    data_type="array",
    axes="YX",
    patch_size=(32, 32),
    batch_size=2,
    num_epochs=1,
    num_workers=0,
)

# create lightning modules
model = N2VModule(config.algorithm_config)

# training data
train_array = rng.integers(0, 255, (64, 64)).astype(np.float32)
val_array = rng.integers(0, 255, (64, 64)).astype(np.float32)

data_module = CareamicsDataModule(
    data_config=config.data_config,
    train_data=train_array,
    val_data=val_array,
)

trainer = Trainer(
    enable_progress_bar=True,
    **config.training_config.trainer_params,
)

# train
trainer.fit(model, datamodule=data_module)

# saving the checkpoint
trainer.save_checkpoint(my_path / "after_training.ckpt")

# create an inference data config
pred_config = config.data_config.convert_mode(
    new_mode="predicting",
    new_patch_size=(64, 64),
    overlap_size=(32, 32),
    new_batch_size=1,
)

inf_data_module = CareamicsDataModule(
    data_config=pred_config,
    pred_data=val_array,
)

# run inference
tiled_predictions = trainer.predict(model, datamodule=inf_data_module)

# saving the checkpoint
trainer.save_checkpoint(my_path / "after_prediction.ckpt")

# load checkpoints
ckpt1 = torch.load(my_path / "after_training.ckpt")
ckpt2 = torch.load(my_path / "after_prediction.ckpt")

data_hparam_key = "datamodule_hyper_parameters"

print(f"\nafter training: {data_hparam_key in ckpt1}, {ckpt1.keys()}\n")
print(f"after prediction: {data_hparam_key in ckpt2}, {ckpt2.keys()}")

```

If we call the `self._save_hparams()` for the prediction datamodule as well, then the saved *data config* will be the prediction config and not the training one.

diyasrivas:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: 
Persist the training data config into every checkpoint saved through the CAREamics trainer, so a checkpoint saved after prediction no longer loses data_config and stays loadable.

PyTorch Lightning's Trainer populates checkpoint["datamodule_hyper_parameters"] from whichever datamodule is active on the trainer when the checkpoint is saved. In CAREamics, CareamicsDataModule only writes data_config into its hyperparameters during setup("fit"/"validate") — never during setup("predict"). After a predict() call the trainer's active datamodule is the prediction datamodule, whose hyperparameters do not contain data_config. Any checkpoint saved at that point has an empty/incomplete datamodule_hyper_parameters, and load_config_from_checkpoint then fails because it requires checkpoint["datamodule_hyper_parameters"]["data_config"].

### Overview - what changed?

The ConfigSaverCallback now persists the training data_config into the checkpoint, independently of which datamodule is currently active on the trainer. This guarantees that any checkpoint saved through the CAREamics trainer, whether after training or after prediction, always carries the training data config and remains loadable.

## Changes Made

`ConfigSaverCallback` takes an optional data_config parameter and writes it into checkpoint["datamodule_hyper_parameters"]["data_config"], overriding whatever the active datamodule left behind. CAREamist passes config.data_config when constructing the callback. Because normalization statistics are resolved in place into that same DataConfig object during setup("fit"), the callback's reference already carries the computed means/stds by the time a checkpoint is saved.

- `ConfigSaverCallback` (src/careamics/lightning/callbacks/config_saver_callback.py): added optional data_config parameter; on_save_checkpoint now persists the training data_config into datamodule_hyper_parameters when provided
- `CAREamist._define_callbacks (src/careamics/careamist.py)`: passes `config.data_config` to `ConfigSaverCallback`
- `docs/current/lightning_api.py`: updated both ConfigSaverCallback constructions to pass config.data_config
- `tests/conftest.py`: the checkpoint fixture now supplies data_config when constructing ConfigSaverCallback

### Removed features or files

None

## How has this been tested?

- All tests in `tests/lightning/utils/test_load_checkpoint.py` and `tests/test_careamist_predict.py` pass locally
- Manually verified the issue scenario: after `train()` → `predict(..., tile_size=...)` → `trainer.save_checkpoint()`, the checkpoint contains `datamodule_hyper_parameters["data_config"]`, is loadable, and reports mode `"training"` (not `"predicting"`)
- The documented Lightning API guide test (`docs/test_guides.py::test_script_execution[Scripts.lightning_api]`) passes locally

## Related Issues

- Resolves #939 

## Breaking changes

None. ConfigSaverCallback gains an optional `data_config` argument (default None) so existing constructions are fine unchanged and only skip persisting the data config when it is not supplied.

---

- [x] Pre-commit passes

### DIFF
--- a/docs/current/lightning_api.py
+++ b/docs/current/lightning_api.py
@@ -50,7 +50,10 @@ callbacks = [
         **config.training_config.checkpoint_params,
     ),
     ConfigSaverCallback(
-        config.version, config.experiment_name, config.training_config
+        config.version,
+        config.experiment_name,
+        config.training_config,
+        config.data_config,
     ),  # (6)!
 ]
 
@@ -100,7 +103,12 @@ callbacks = [
         filename=f"{config.experiment_name}_{{epoch:02d}}_step_{{step}}",
         **config.training_config.checkpoint_params,
     ),
-    ConfigSaverCallback(config.version, config.experiment_name, config.training_config),
+    ConfigSaverCallback(
+        config.version,
+        config.experiment_name,
+        config.training_config,
+        config.data_config,
+    ),
     pred_writer,  # (2)!
 ]
 
--- a/docs/current/lightning_api.md
+++ b/docs/current/lightning_api.md
@@ -29,8 +29,11 @@ are setting up Noise2Void, but the same can be done for CARE and N2N using the `
 For more details, refer to the [data documentation]().
 5. Careamics configuration has its own default set of parameters for the `ModelCheckpoint`
 callback (and `EarlyStopping`). You can either use those or set up your own.
-6. The `ConfigSaverCallback` callback is used to log the configuration in the 
-checkpoints.
+6. The `ConfigSaverCallback` callback stores the CAREamics configuration in the
+checkpoints. It is optional in the Lightning API, add it if you want to load the
+checkpoint with the CAREamist API and recover the training data and training configs.
+Without it, the checkpoint can still be loaded, but the training configuration falls back
+to its defaults.
 7. Similarly, the configuration create training parameters configuration. You can set your
 own rather than reusing those.
 8. As in any Lightning script, pass the model and datamodule to the trainer and call `fit` to start training.
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -385,6 +385,7 @@ class CAREamist:
                 config.version,
                 config.get_safe_experiment_name(),
                 config.training_config,
+                config.data_config,
             ),
         ]
 
--- a/src/careamics/lightning/callbacks/config_saver_callback.py
+++ b/src/careamics/lightning/callbacks/config_saver_callback.py
@@ -5,6 +5,7 @@ from typing import Any
 from pytorch_lightning import LightningModule, Trainer
 from pytorch_lightning.callbacks import Callback
 
+from careamics.config.data.data_config import DataConfig
 from careamics.config.lightning.training_configuration import TrainingConfig
 
 
@@ -14,6 +15,8 @@ class ConfigSaverCallback(Callback):
 
     This callback automatically stores CAREamics version, experiment name,
     and training configuration in the checkpoint file for reproducibility.
+    It also persists the training data configuration into the checkpoint,
+    independently of the datamodule currently active on the trainer.
 
     Parameters
     ----------
@@ -23,6 +26,8 @@ class ConfigSaverCallback(Callback):
         Name of the experiment.
     training_config : TrainingConfig
         Training configuration to store in checkpoint.
+    data_config : DataConfig
+        Training data configuration to store in checkpoint.
 
     Attributes
     ----------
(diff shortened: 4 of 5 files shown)


docs/current/lightning_api.py
```diff
@@ -50,7 +50,10 @@
         **config.training_config.checkpoint_params,
     ),
     ConfigSaverCallback(
-        config.version, config.experiment_name, config.training_config
+        config.version,
+        config.experiment_name,
+        config.training_config,
+        config.data_config,
```
melisande-c, line 56: Maybe not totally part of the scope of the PR but I realised we don't mention why one might want to use the `ConfigSaverCallback` with the lightning API, technically it is optional. It would be so that you can later load the checkpoint with the CAREamist API instead of the Lightning API and have the training config saved, otherwise it will be loaded with the default training config.

melisande-c (approved):
I've thought about this a little bit because to me it doesn't feel quite right the way this overwrites the lightning checkpointing behaviour. Especially since lightning do not document `trainer.save_checkpoint`, even though it is not a private method, so I guess they do not expect people to be calling it manually after prediction.

A reason not to overwrite lightning behaviour is that now we will be coupled to their API. For example, say they decide to change the name of the `"datamodule_hyper_params"` key or something else, we would also have to update this accordingly and then maybe also raise the lower bounds of our lightning dependency which could cause us to be incompatible with other libraries.

With all that being said, I feel it is very unlikely that they will change the `datamodule_hyper_params` key. It is a relatively simple fix and I couldn't think of anything else

docs/current/lightning_api.py
diyasrivas: I can also update the `lightning_api.md` to make it explicit that `ConfigSaverCallback` is optional in the Lightning API

jdeschamps:
Ignore the pre-commit errors, they will be fixed elsewhere. Is it good to merge?

PR #1000, opened 2026-07-06
fix: defaults for predict-to-disk write type and in-memory

### LINKED ISSUE #877: Next-Gen: Predict to disk quality-of-life improvements
## Problem

There are a few things that make using the predict to disk functionality a bit annoying:

### Default TIFF write_type in CAREamistV2

We can debate this, but maybe if the data_type is `zarr` the `write_type` should also default to `zarr`. 

Maybe this is also partly that the `write_type` argument is near the end of the function signature and a user might miss it, so we should also make it more promenent.

https://github.com/CAREamics/careamics/blob/88235396d21fcac040d4b35223529411a0cd0dc7/src/careamics/careamist_v2.py#L822

### Switching to zarr without updating `in_memory` causes an error

If the training data type was in memory but we change the write type to zarr then there will be the error that "in memory is not supported for zarr" because the default `in_memory` used is the same as training.

Maybe we can just document this better.

### Print statement in `WriteTilesZarr` called for every tile

There is a print statement in the zarr write strategy that ends up getting called for every tile. This polutes the terminal output

https://github.com/CAREamics/careamics/blob/88235396d21fcac040d4b35223529411a0cd0dc7/src/careamics/lightning/dataset_ng/callbacks/prediction_writer/write_tiles_zarr_strategy.py#L108

diyasrivas:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: 
`predict_to_disk` now defaults its write type from the prediction data type (zarr → zarr) and no longer lets a stale in-memory training setting break zarr/czi predictions.

Two remaining issues when predicting on zarr data:
1. `predict_to_disk(..., write_type=...)` defaulted to `"tiff"` regardless of the input `data_type`. A user predicting on a zarr and forgetting `write_type` would silently get TIFF output.
2. When training with in-memory data (tiff/array, where `in_memory` defaults to `True`) and then predicting on zarr, the `in_memory=True` value carried over from the training configuration and tripped the validator with `` `in_memory` not supported for `data_type` zarr `` — even though the user never asked for in-memory zarr.

### Overview - what changed?

`predict_to_disk` now infers `write_type` from the effective prediction `data_type`, and both `predict`/`predict_to_disk` no longer carry a stale `in_memory` value into a zarr/czi prediction when the user did not explicitly set it.

in `src/careamics/careamist.py`:

- Write type inference: the `write_type` default changed from `"tiff"` to `None` (across the overloads and implementation). When `None`, it resolves to `"zarr"` if the effective data type (`data_type or self.config.data_config.data_type`) is `"zarr"`, else `"tiff"`. Explicit values behave exactly as before, so this is backward-compatible.

- In-memory carry-over: in `_build_predict_datamodule`, when in_memory is not explicitly provided and the effective data type is zarr/czi, it is set to False before building the prediction config. An explicit in_memory=True still raises, as before.

if in_memory is None and (
    (data_type or self.config.data_config.data_type) in ("zarr", "czi")
):
    in_memory = False

### New features or files

None

- `predict_to_disk` in `src/careamics/careamist.py`: `write_type` default changed to None with inference from the effective data type
- `_build_predict_datamodule` in `src/careamics/careamist.py`: skip carrying over in_memory for zarr/czi prediction data when not explicitly set
- tests/test_careamist_predict.py: added tests for the new defaulting behaviour

### Removed features or files

None

## How has this been tested?

 All test_careamist_predict.py tests pass + 2 new tests added:
- test_predict_to_disk_zarr_input_defaults_to_zarr: trains in-memory on tiff, then predicts on zarr input with write_type and in_memory both omitted, and asserts a .zarr output is produced 
- test_predict_to_disk_zarr_explicit_in_memory_raises: an explicit in_memory=True with zarr data still raises

## Related Issues

- Resolves #877 

## Breaking changes

None

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -224,6 +224,60 @@ class CAREamist:
             return self._from_bmz(bmz_path)
 
     @staticmethod
+    def _default_write_type(
+        write_type: Literal["tiff", "zarr", "custom"] | None,
+        data_type: Literal["array", "tiff", "zarr", "czi", "custom"],
+    ) -> Literal["tiff", "zarr", "custom"]:
+        """Default the write type from the prediction data type.
+
+        Parameters
+        ----------
+        write_type : {"tiff", "zarr", "custom"} or None
+            The write type requested by the user, or None to infer it.
+        data_type : {"array", "tiff", "zarr", "czi", "custom"}
+            The (effective) prediction data type.
+
+        Returns
+        -------
+        {"tiff", "zarr", "custom"}
+            The resolved write type.
+        """
+        if write_type is not None:
+            return write_type
+
+        # config data type can be custom or czi, we support only zarr and tiff
+        # writing if no writing function is passed
+        return "zarr" if data_type == "zarr" else "tiff"
+
+    @staticmethod
+    def _predict_in_memory(
+        in_memory: bool | None,
+        data_type: Literal["array", "tiff", "zarr", "czi", "custom"],
+    ) -> bool | None:
+        """Resolve the `in_memory` setting for prediction.
+
+        An explicit user value is always respected, so the incompatibility of an
+        explicit `in_memory=True` with zarr/czi is still surfaced by configuration
+        validation.
+
+        Parameters
+        ----------
+        in_memory : bool or None
+            The `in_memory` value requested by the user, or None to infer it.
+        data_type : {"array", "tiff", "zarr", "czi", "custom"}
+            The (effective) prediction data type.
+
+        Returns
+        -------
+        bool or None
+            The resolved `in_memory` value. None defers to the training configuration.
+        """
+        # overwrite data cfg in_memory with `False` if czi or zarr
+        if in_memory is None and data_type in ("zarr", "czi"):
+            return False
+        return in_memory
+
+    @staticmethod
     def _from_config(
         config: ConfigurationType | Path | str,
     ) -> tuple[ConfigurationType, CAREamicsModule]:
@@ -698,11 +752,8 @@ class CAREamist:
         if num_workers is not None:
             dataloader_params = {"num_workers": num_workers}
 
-        # overwrite data cfg in_memory with `False` if czi or zarr
-        if in_memory is None and (
-            (data_type or self.config.data_config.data_type) in ("zarr", "czi")
-        ):
-            in_memory = False
+        effective_data_type = data_type or self.config.data_config.data_type
+        in_memory = self._predict_in_memory(in_memory, effective_data_type)
 
         pred_data_config = self.config.data_config.convert_mode(
             new_mode="predicting",
@@ -1065,12 +1116,8 @@ class CAREamist:
             write_func_kwargs = {}
 
         # default the write type from the prediction data type
-        if write_type is None:
-            effective_data_type = data_type or self.config.data_config.data_type
-
-            # config data type can be custom or czi, we support only zarr and tiff
-            # writing if no writing function is passed
-            write_type = "zarr" if effective_data_type == "zarr" else "tiff"
+        effective_data_type = data_type or self.config.data_config.data_type
+        write_type = self._default_write_type(write_type, effective_data_type)
 
         if Path(prediction_dir).is_absolute():
             write_dir = Path(prediction_dir)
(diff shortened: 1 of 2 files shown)


src/careamics/careamist.py
```diff
@@ -1054,6 +1064,11 @@ def predict_to_disk(
         if write_func_kwargs is None:
             write_func_kwargs = {}
 
+        # default the write type from the prediction data type
+        if write_type is None:
+            effective_data_type = data_type or self.config.data_config.data_type
+            write_type = "zarr" if effective_data_type == "zarr" else "tiff"
```
jdeschamps: I found this bit a bit counterintuitive at first, but I understand that `cfg...data_type` can be `custom` or `czi`, which we do not support as write type. Maybe an additional comment?

```suggestion
        if write_type is None:
            effective_data_type = data_type or self.config.data_config.data_type
            
            # config data type can be custom or czi, we support only zarr and tiff
            # writing if no writing function is passed
            write_type = "zarr" if effective_data_type == "zarr" else "tiff"
```

src/careamics/careamist.py
```diff
@@ -698,6 +698,12 @@ def _build_predict_datamodule(
         if num_workers is not None:
             dataloader_params = {"num_workers": num_workers}
 
+        # do not let a stale training value carry over
```
jdeschamps: Clarify comment:

```suggestion
        # overwrite data cfg in_memory with `False` if czi or zarr
```

tests/test_careamist_predict.py
```diff
@@ -322,3 +323,89 @@ def test_predict_invalid_spatial_dims_no_tiling_raises(tmp_path: Path):
...
+        tile_size=(16, 16),
+        tile_overlap=(4, 4),
+    )
+
+    outputs = list((tmp_path / "predictions").glob("*.zarr"))
+    assert len(outputs) == 1
+
+
+def test_predict_to_disk_zarr_explicit_in_memory_raises(tmp_path: Path):
+    """An explicit `in_memory=True` with zarr data still raises."""
```
jdeschamps:
```suggestion
    """Test that an explicit `in_memory=True` with zarr data raises an error."""
```

tests/test_careamist_predict.py
```diff
@@ -322,3 +323,89 @@ def test_predict_invalid_spatial_dims_no_tiling_raises(tmp_path: Path):
...
+
+
+@pytest.mark.mps_gh_fail
+def test_predict_to_disk_zarr_input_defaults_to_zarr(tmp_path: Path):
+    """Predicting on zarr input defaults the write type to zarr.
+
+    Also checks that a stale in-memory training value does not carry over and
+    conflict with the zarr prediction data type.
+    """
+    careamist = _train_in_memory_tiff_careamist(tmp_path)
```
jdeschamps: Better than such a function, we should check if we can use the "checkpoint" fixture defined in `conftest.py`.

tests/test_careamist_predict.py
```diff
@@ -322,3 +323,89 @@ def test_predict_invalid_spatial_dims_no_tiling_raises(tmp_path: Path):
...
+        batch_size=2,
+        num_epochs=1,
+        roi_size=5,
+        masked_pixel_percentage=5,
+    )
+    careamist = CAREamist(config=config, work_dir=tmp_path)
+
+    zarr_path = _write_zarr_input(tmp_path / "pred.zarr", random_array((32, 32)))
+
+    with pytest.raises(ValueError, match="in_memory"):
```
jdeschamps: Where does the error come from? Was it raised before?

tests/test_careamist_predict.py
```diff
@@ -322,3 +323,89 @@ def test_predict_invalid_spatial_dims_no_tiling_raises(tmp_path: Path):
...
+    # training on tiff defaults `in_memory` to True
+    assert config.data_config.in_memory is True
+
+    careamist = CAREamist(config=config, work_dir=tmp_path)
+    careamist.train(train_data=image_dir, val_data=val_file)
+    return careamist
+
+
+@pytest.mark.mps_gh_fail
+def test_predict_to_disk_zarr_input_defaults_to_zarr(tmp_path: Path):
```
jdeschamps: I am tempted to remove this test, the problem is that it is a lengthy one (training) for testing a very small feature. That feature should be extracted to a private function and that function itself tested.

jdeschamps (changes requested):

jdeschamps (approved):

PR #1005, opened 2026-07-07
Feat: introduce `target_axes` to allow loading targets with different axes 

### LINKED ISSUE #963: Allow input and ouput with different axes

jdeschamps:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: Add `target_axes` configuration parameter to be able to load targets with different axes than the input data.

Currently, targets have to have the same axes as input otherwise an error is raised in the dataset. This leads to the need of having singleton channels for instance for MicroSplit. 

This PR is a first step towards fixing all the issues in having different input and target axes.

To be able to load targets with different axes as the input.

### Overview - what changed?

New `target_axes` parameter in the `DataConfig`.

Additional changes are some refactoring motivated by the need to implement unit tests for the configuration factories to verify that the validation is done properly.

Regarding `target_axes`:
- Introduced `target_axes` in `DataConfig`, with validation similar to axes and a validation for spatial dimensionality
- Added `target_axes` to CARE configuration factory
- Added `target_axes` to data configuration factory
- Substituted `target_axes` to `axes` in the target patch extractor calls in dataset and microsplit dataset factories
- Removed `channels` adjustment to the target's `ImageRegionData` in the dataset
- Added a validation to ensure that `target_axes` is `None` in N2V configuration

Additional refactoring:
- `factory_utils.py` now contains the handling of the augmentations for both care/n2n and n2v convenience functions, and corresponding unit tests have been added
- `care_n2n_factory`: channel validation has been moved to _validate_channel_dim and tests for this function have been added
- `n2v_factory`: same and an additional refactoring of the N2VManipulateConfig creation so as to be able to test it directly (corrsponding tests have been implemented).
- `SupportedTransform` renamed `SupportedAugmentation`
- `create_ng_data_configuration` renamed `create_data_configuration`

Unit tests:
- a test module for all convenience factories has been added that tests orthogonal parameters for all three factories (advanced n2v, care and n2n)
- specific tests for parameters that are exclusive to the convenience functions have been added to the respective test modules

## How has this been tested?

New tests.

## Resolves

- Resolves https://github.com/CAREamics/careamics/issues/963

## Notes

- Note that the use of `channels` to adjust the "C" channel in the target `ImageRegionData` data shape (in the Dataset) has been removed, that is a departure from previous behavior.
- I left `target_axes` in N2N as well, in case it is a single plane vs (N-1) stack N2N training (DeepCAD). 
- An error will still occur if there is a mismatch between axes and model output dimensions. It will be a follow-up PR. 
- Currently, this PR introduces a validation to ensure that spatial axes are the same. That is an early-failure validation that can be removed the day we want to support that.

---

- [x] Pre-commit passes


jdeschamps:
## Catching non-channel axes mismatch

One feature that probably needs to be implemented: if `target_axes` is `None` and an error is raised because of mismatching between shape and axes (e.g. axes "YX" but target is "ZYX"), then the message should encourage users to set a `target_axes` value.

The reason is that only channels differences can be caught in the convenience functions (due to the specification of the number of model inputs / outputs).

## `target_axes` as a reference for restoring shape

Out of scope here, but it will need to be another PR. Model output should be restored according to `target_axes` if the latter is specified. For that, `target_axes` needs to be forwarded via the `ImageRegionData`, either as a new entry or replacing `axes` and `original_data_shape` with the target's.

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -4,21 +4,105 @@
...
+    n_channels_out: int | None,
+) -> tuple[int, int]:
+    """Validate channel dimensions and adjust model input and output channel numbers.
+
+    Parameters
+    ----------
+    axes : str
+        Axes of the data (e.g. SYX).
+    target_axes : str or None
+        Axes of the target data. Required when the output has more than one channel.
```
melisande-c: Below, you actually also allow `target_axes` to be `None` if `n_channels_in==n_channels_out`.

But maybe it is simpler to assume that if `target_axes` is not specified it is equal to `axes` (this was the previous behaviour). Now we only need to guard for 
```python
 if not target_channels_present and resolved_n_channels_out > 1:
```

I haven't read into the dataloading yet but if the loaded target data cannot be reshaped with `axes` maybe we can catch the error and hint that they should check the `target_axes`.

tests/config/factories/test_care_factory.py
melisande-c, line 1: I guess these tests have been refactored to `unit/config/factories/test_care_n2n_factory.py`?

tests/unit/config/data/test_microsplit_data_config.py
```diff
@@ -99,3 +99,17 @@ def test_convert_mode_preserves_ms_options(
     assert converted.padding_mode == config.padding_mode
     assert converted.alpha_ranges is None
     assert converted.uncorrelated_channel_prob == 0.0
+
+
+def test_convert_mode_preserves_target_axes() -> None:
+    """Test MicroSplit mode conversion preserves inherited target axes."""
+    config = MicroSplitDataConfig(
```
melisande-c: If this is already tested in `DataConfig` then this test is not super necessary since there is nothing in `MicroSplitDataConfig` that specifically interacts with the `target` axes.

tests/unit/config/factories/test_care_n2n_factory.py
```diff
@@ -0,0 +1,131 @@
...
+# --- List of test parameters
+
+VALID_CASES = [
+    # axes, target_axes, channels, in_channels, out_channels, expected
+    # no channels
+    ("YX", None, None, None, None, (1, 1)),
+    ("YX", None, None, 1, None, (1, 1)),
+    ("YX", None, None, 1, 1, (1, 1)),
+    ("YX", None, None, None, 1, (1, 1)),
+    ("YX", "ZYX", None, None, None, (1, 1)),
```
melisande-c: `YX` -> `ZYX` is allowed?

tests/unit/config/factories/test_care_n2n_factory.py
```diff
@@ -0,0 +1,131 @@
...
+    # with channels
+    ("CYX", None, [0], None, None, (1, 1)),
+    ("CYX", None, [0, 2], None, None, (2, 2)),
+    ("CYX", None, None, 3, None, (3, 3)),
+    ("CYX", "CYX", [0, 2], 2, 4, (2, 4)),
+    ("YX", "CYX", None, 1, 2, (1, 2)),
+    ("YX", "CYX", None, None, 2, (1, 2)),
+    ("CYX", "YX", None, 2, 1, (2, 1)),
```
melisande-c: Maybe we should also add `CYX` -> `YXC` just in case

tests/unit/config/factories/test_care_n2n_factory.py
```diff
@@ -0,0 +1,131 @@
...
+        None,
+        "does not match length of `channels`",
+    ),
+    (
+        "YX",
+        None,
+        None,
+        1,
+        2,
+        "`target_axes` must be specified",
```
melisande-c: I think the case:

`axes="CYX", n_channels_out=2` will also result in the `"target_axes must be specified"` error

tests/unit/config/factories/test_care_n2n_factory.py
melisande-c: oh I guess the `_validate_channel_dim` does not check the dimensions.

tests/unit/config/factories/test_data_factory.py
```diff
@@ -88,7 +88,7 @@ def test_train_dataloader_params(self):
 
     def test_default_patching(self):
         """Test that the default patching strategy is random."""
```
melisande-c:
```suggestion
        """Test that the default patching strategy is stratified."""
```

src/careamics/config/factories/n2v_factory.py
```diff
@@ -19,12 +15,122 @@
...
+        raise ValueError(
+            f"C is not present in the axes, but number of channels is specified "
+            f"(got {n_channels} channel)."
+        )
+
+    if n_channels is not None and channels is not None and n_channels != len(channels):
+        raise ValueError(
+            f"Number of channels ({n_channels}) does not match length of "
+            f"`channels` ({len(channels)}). Only specify `channels`."
+        )
```
melisande-c: I guess it's ok but this section is pretty much a duplicate of the start of `care_n2n_factory._validate_channel_dim`. If it was one utility function some tests could be removed.

Also there might be a way to reuse it for the `target_axes` if we try my suggestion of assuming `target_axes=axes` if `target_axes is None`.

melisande-c (commented):
Looks good but a few things in the tests and one suggestion about when require the `target_axes` to be supplied for care/n2n, currently I am pretty sure it will throw an error for:
```python
  axes="CYX",
  target_axes=None,
  n_channels_in=3,
  n_channels_out=2,
```
when maybe in this case it is safe to assume both the target and input have the same channel order.

Also, it would be nice if we had some e2e tests to make sure all the moving parts are assembled correctly and model training works for different input and output axes.

src/careamics/config/factories/care_n2n_factory.py
jdeschamps: This situation:
```python
from careamics.careamist import CAREamist
from careamics.config.factories import create_advanced_care_config
from matplotlib import pyplot as plt
import numpy as np

train_data = [rng.integers(0, 255, (2, 128, 128, 3)).astype(np.float32) for _ in range(4)]
train_target = [rng.integers(0, 255, (2, 2, 128, 128)).astype(np.float32) for _ in range(4)]

# create a configuration
config = create_advanced_care_config(
    experiment_name="api_example",
    data_type="array",
    axes="SYXC",
    target_axes=None,
    patch_size=[64, 64],
    # batch size, num_epochs and num_steps reduced for the sake of example
    batch_size=4,
    num_epochs=2,
    num_steps=10,
    channels=[0, 1],
    n_channels_out=2,
)

# instantiate a careamist
careamist = CAREamist(config)

# train the model
careamist.train(train_data=train_data, train_data_target=train_target)
```

leads to the target to be misshaped and the padding actually results in a patch of the correct shape, failing silently (because now targets is mostly 0s).

I agree that in principle this situation should not happen, but it will also be super hard to debug... I guess it could happen when there is a singleton channel in input for some reason, and multiple channels in output (e.g. segmentation).

I'd rather force users to specify `target_axes`, rather than have a silent failure such as this one. 

An alternative is to show a warning saying that if the input and target have different axes, a silent failure with training not converging is expected.

tests/config/factories/test_care_factory.py
jdeschamps: Not quite. This PR introduces some refactoring, so for instance tests that pertain to the augmentations are now performed on the utility function that actually deals with the augmentations.

Then a new `test_config_factories` generalizes a simple test for the shared parameters of CARE, N2V and N2N configuration factories.

So whatever these previous tests was testing has roughly been reimplemented from scratch in theses tests.

tests/unit/config/data/test_microsplit_data_config.py
jdeschamps: That's a good point, removed in https://github.com/CAREamics/careamics/pull/1005/changes/5ee901498fef8d5289c6a9742bff19df6bcd4945.

tests/unit/config/factories/test_care_n2n_factory.py
jdeschamps: Added in https://github.com/CAREamics/careamics/pull/1005/changes/324b699682c568def501d0bc287b6f0d7ae5d943. I added a few situations including singleton axes with these axes.

tests/unit/config/factories/test_care_n2n_factory.py
jdeschamps: Added in https://github.com/CAREamics/careamics/pull/1005/changes/324b699682c568def501d0bc287b6f0d7ae5d943

tests/unit/config/factories/test_data_factory.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/1005/changes/360a83affb140483e3dd134a6f0f8d4fc1923f39

src/careamics/config/factories/n2v_factory.py
jdeschamps: Yeah that would make sense to do further refactoring.

src/careamics/config/factories/care_n2n_factory.py
jdeschamps: Nevermind, your suggestion obviously covers that case (n_channels_out>1). I had stashed the changes when testing something else and forgot to pop them...

jdeschamps: Alright, so after running the example with the suggestion, it leads to the same issue. I am fine with having the users be responsible for having the axes in the same order, especially given that it is a very unlikely situation. I would however add a warning.

jdeschamps: Implemented in https://github.com/CAREamics/careamics/pull/1005/changes/abb20fbe17542a3d5b88ae39c6d268507519ae99

jdeschamps:
https://github.com/CAREamics/careamics/pull/1005/changes/0ebb3999d05e52f27f83424f58b10e79309ea7a6: I had to cover yet another edge case: if `target_axes` is specified without "C" (e.g. "YX"), `n_channels_out` was set to `n_channels_in` (e.g. size 3 with `axes="CYX"`)...

src/careamics/config/factories/n2v_factory.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/1005/changes/de387f3749da040e0b4736b6550cfbc164101333.

Test cases relevant for the new utility function have been moved from the factory tests to the utility test.

jdeschamps:
https://github.com/CAREamics/careamics/pull/1005/changes/01f27179fd242a42b529075530b1b1976b10482f: an e2e test (not implemented here) showed that we will run into issues if using `channels` and if `channels` is not compatible with `target_axes`.

Regarding the last point, either we limit `channels` to input or we only allow to use it if it is compatible with the both input and output. Currently, I simply limited it to input (simpler) and probably would need to extend the docs and docstring.

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -4,20 +4,98 @@
...
+        axes=axes, channels=channels, n_channels=n_channels_in, attr_name="n_channels"
+    )
+
+    # resolve number of input channels
+    if n_channels_in is None and channels is None:
+        resolved_n_channels_in = 1
+    elif n_channels_in is not None:
+        resolved_n_channels_in = n_channels_in
+    else:
+        assert channels is not None
```
melisande-c, line 67: I assume that `channels` cannot be None at this point due to `validate_input_channels` catching the other cases?

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -4,20 +4,98 @@
...
+        resolved_n_channels_in = len(channels)
+
+    resolved_n_channels_out = (
+        n_channels_out
+        if n_channels_out is not None
+        else (
```
melisande-c: maybe would be a little more legible as a if else block instead of a nested ternary statement

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -78,8 +157,16 @@ def create_care_config(
...
     n_channels_in : int or None, default=None
         Number of input channels.
-    n_channels_out : int or None, default=None
-        Number of output channels.
+    n_channels_out : int | None, default=None
+        Number of output channels. If not specified, it will default to `n_channels_in`
+        if the latter is specified and `target_axes` includes "C" (note that if
+        `target_axes` is not specified, it is assumed to be equal to `axes`), then
+        `n_channels_out` will be set to `n_channels_in`. Otherwise, it will be set to
+        1.
```
melisande-c:
```suggestion
        Number of output channels. If `target_axes` contains `C` then it will default to
        `n_channels_in` otherwise it will default to 1. Note that if `target_axes` is 
        `None` it will default to `axes`.
```

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -78,8 +157,16 @@ def create_care_config(
...
+    target_axes : str or None, default=None
+        Axes of target data. Required when the output has more than one channel, unless
+        `axes` includes "C", in which case `target_axes` is assumed to be equal to
+        `axes`.
```
melisande-c:
```suggestion
        Axes of target data. If not specified it will default to `axes`. Pay extra
        attention if you are training a task that has a different number of input and 
        output channels.
```

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -148,8 +236,16 @@ def create_n2n_config(
...
```
melisande-c:
```suggestion
        Number of output channels. If `target_axes` contains `C` then it will default to
        `n_channels_in` otherwise it will default to 1. Note that if `target_axes` is 
        `None` it will default to `axes`.
    target_axes : str or None, default=None
        Axes of target data. If not specified it will default to `axes`. Pay extra
        attention if you are training a task that has a different number of input and 
        output channels.
```

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -242,8 +339,15 @@ def create_advanced_care_config(
...
-        it will default to the same number as n_channels_in.
+        Number of output channels. If not specified, it will default to `n_channels_in`
+        if the latter is specified and `target_axes` includes "C" (note that if
+        `target_axes` is not specified, it is assumed to be equal to `axes`), then
+        `n_channels_out` will be set to `n_channels_in`. Otherwise, it will be set to
+        1.
+    target_axes : str or None, default=None
+        Axes of target data. Required when the output has more than one channel, unless
+        `axes` includes "C", in which case `target_axes` is assumed to be equal to
+        `axes`.
```
melisande-c:
```suggestion
        Number of output channels. If `target_axes` contains `C` then it will default to
        `n_channels_in` otherwise it will default to 1. Note that if `target_axes` is 
        `None` it will default to `axes`.
    target_axes : str or None, default=None
        Axes of target data. If not specified it will default to `axes`. Pay extra
        attention if you are training a task that has a different number of input and 
        output channels.
```

src/careamics/config/factories/care_n2n_factory.py
```diff
@@ -401,6 +506,8 @@ def create_advanced_n2n_config(
     n_channels_out : int | None, default=None
         Number of output channels. If not specified, but n_channels_in is specified,
         it will default to the same number as n_channels_in.
+    target_axes : str or None, default=None
+        Axes of target data. Required when the output has more than one channel.
```
melisande-c:
```suggestion
        Number of output channels. If `target_axes` contains `C` then it will default to
        `n_channels_in` otherwise it will default to 1. Note that if `target_axes` is 
        `None` it will default to `axes`.
    target_axes : str or None, default=None
        Axes of target data. If not specified it will default to `axes`. Pay extra
        attention if you are training a task that has a different number of input and 
        output channels.
```

melisande-c (approved):
Looks good, just some super minor stuff

melisande-c:
> [01f2717](https://github.com/CAREamics/careamics/commit/01f27179fd242a42b529075530b1b1976b10482f): an e2e test (not implemented here) showed that we will run into issues if using `channels` and if `channels` is not compatible with `target_axes`.
> 
> Regarding the last point, either we limit `channels` to input or we only allow to use it if it is compatible with the both input and output. Currently, I simply limited it to input (simpler) and probably would need to extend the docs and docstring.

I could see it being useful to have a `target_channels` parameter, but lets just limit the `channels` to the input for now.

melisande-c:
Oh actually also there are some tests failing now

src/careamics/config/factories/care_n2n_factory.py
jdeschamps: I think I added this for mypy (otherwise I don't see why this would be there)

src/careamics/config/factories/care_n2n_factory.py
jdeschamps: Done in https://github.com/CAREamics/careamics/pull/1005/commits/1b626b62fd94a14350112446e8b7138f66bc35cd

src/careamics/config/factories/care_n2n_factory.py
melisande-c: yes for mypy but because the previous validation should ensure that `channels` cannot be `None` at this point.

jdeschamps: I know, but somehow it does not.

```python
src/careamics/config/factories/care_n2n_factory.py:68: error: Argument 1 to "len" has incompatible type "Sequence[int] | None"; expected "Sized"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

jdeschamps:
> Oh actually also there are some tests failing now

Thanks for pointing it out, adapted the tests in https://github.com/CAREamics/careamics/pull/1005/changes/f40de95d0b8cd5010fa79c9e98a1054201d30435

src/careamics/config/factories/care_n2n_factory.py
melisande-c: what do you mean? pre-commit is passing

jdeschamps: I removed the line and ran pre-commit to check that mypy indeed does not understand that `channels` has to be `None`.

jdeschamps: I guess it is just how it is build: https://mypy.readthedocs.io/en/stable/type_narrowing.html?utm_source=chatgpt.com#limitations

jdeschamps: But maybe I also misunderstood your comment and you were not discussing the presence of the `assert` at all.

melisande-c: Yes I just commented to confirm that `channels` could not accidentally be `None` at that point, because it is not clear otherwise

PR #1023, opened 2026-07-23
Docs: improve docs

jdeschamps:
## Disclaimer

Forgot to update the docs for the change of behaviour in https://github.com/CAREamics/careamics/pull/1005.

### DIFF
--- a/docs/current/configuration.md
+++ b/docs/current/configuration.md
@@ -206,8 +206,7 @@ possibility.
 !!! note "Assumptions parameters are not passed"
 
     If `target_axes` is not set, it is assumed to be the same as `axes`. If
-    `n_channels_out` is not set, then it is set to the same value as `n_channels_in` if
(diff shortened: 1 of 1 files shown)


docs/current/configuration.md
```diff
@@ -203,6 +203,14 @@ possibility.
 2. The number of output channels can be different from the number of input channels.
 
 
+!!! note "Assumptions parameters are not passed"
+
+    If `target_axes` is not set, it is assumed to be the same as `axes`. If
+    `n_channels_out` is not set, then it is set to the same value as `n_channels_in` if
+    `target_axes` has a "C" channel, otherwise 1.
```
melisande-c:
```suggestion
    `n_channels_out` is not set and `target_axes` has a "C" dimension, then it is set to the same value as `n_channels_in`, otherwise it is set to 1.
```

melisande-c (approved):

PR #1029, opened 2026-07-25
Noise model normalization

### LINKED ISSUE #857: NG MicroSplit: test normalize noise model
## Problem

Currently, data is denormalized in the loss in order to use the noise models. Chances are the noise model could be normalized once, and lead to the same results.

This could be verified mathematically or by running the methods against random values.

CatEek:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: The noise-model likelihood is now normalized once into the network's data space instead of denormalizing predictions/targets on every training step

For MicroSplit/HDN training, the noise-model likelihood was evaluated in raw pixel space: `_compute_noise_model_log_likelihood` denormalized the reconstruction and target with `data_mean`/`data_std` on every loss call, and those statistics had to be threaded through `MicroSplitModule`/`HDNModule` and the loss functions. In `MicroSplitModule.on_fit_start`, this used only the *first* target channel's statistics (`means[0]`, `stds[0]`) to denormalize **all** channels, so any channel other than 0 was evaluated by its noise model in the wrong intensity space whenever per-channel target statistics differed.

### Overview - what changed?

The GM NM can now be transformed once into normalized data space (`get_normalized_copy`), so the loss operates directly on normalized reconstructions/targets with no per-step denormalization. Each channel is normalized with its own statistics. `data_mean`/`data_std` are removed from the loss functions and Lightning modules entirely.

- mean-coefficient rows are divided by `std`
- log-variance-coefficient rows are shifted by `-ln(std²)` (they parameterize the *variance* via `exp(w)`)
- mixture-weight rows are unchanged
- `min_signal`/`max_signal` are mapped affinely; `min_sigma` (which clamps the variance) is divided by `std²`

`get_normalized_copy` returns a new model (`is_normalized=True`) and leaves the original untouched. Both `MicroSplitModule.on_fit_start` and `HDNModule.on_fit_start` **rebuild the raw-space model from the configuration** and replace `self.noise_model` with a normalized copy, rather than mutating the live model in place — this keeps the hook idempotent (e.g. on checkpoint resume, where the module's `nn.Module` state would otherwise already contain normalized weights). `save()` on a normalized model raises, since the `.npz` format stores raw-space models only.

```python
# before
loss = microsplit_loss(..., noise_model=nm, data_mean=mean, data_std=std)

# after
nm_normalized = nm.get_normalized_copy(means, stds)  # done once, in on_fit_start
loss = microsplit_loss(..., noise_model=nm_normalized)
```

### New features or files

- `tests/models/lvae/test_noise_model_normalization.py`: exactness/gradient-equivalence tests for `get_normalized_copy`, parameter-transform checks, multichannel per-channel stats, and a module-level regression test locking in the per-channel bug fix.

- `src/careamics/models/lvae/noise_models.py`: `GaussianMixtureNoiseModel.get_normalized_copy`, `is_normalized`/`normalization_mean`/`normalization_std` attributes, `save()` guard against normalized models; `MultiChannelNoiseModel.get_normalized_copy`/`is_normalized`.
- `src/careamics/losses/lvae/lvae_losses.py`: `_compute_noise_model_log_likelihood`, `hdn_loss`, `microsplit_loss` no longer take `data_mean`/`data_std`.
- `src/careamics/lightning/modules/microsplit_module.py`: `on_fit_start` rebuilds and normalizes the noise model per channel; removed `self.data_mean`/`self.data_std`.
- `src/careamics/lightning/modules/hdn_module.py`: same pattern using input statistics; the "call `trainer.fit` first" guard now checks `noise_model.is_normalized`.
- `tests/losses/test_lvae_losses.py`, `tests/likelihood_modules/test_likelihoods.py`, `tests/models/lvae/test_noise_model.py`: updated to the new normalized-model API; legacy-equivalence tests now assert equality up to the expected constant shift (`mean(ln std)`).

### Removed features or files

- `tests/noise_model/test_noise_model_trainer.py`: deleted a ~275-line dead commented-out test block referencing a removed `set_noise_model` API.

## How has this been tested?

- New unit tests for `get_normalized_copy`

## Related Issues

- Resolves #857

## Breaking changes

- `_compute_noise_model_log_likelihood`, `hdn_loss`, and `microsplit_loss` no longer accept `data_mean`/`data_std` keyword arguments; callers must pass an already-normalized `noise_model` (via `get_normalized_copy`) instead.
- `GaussianMixtureNoiseModel.save()` and config-based save paths now raise `ValueError` if called on a normalized model.

- `.pre-commit-config.yaml`: exclude patterns updated so `noise_models.py`

---

- [x] Pre-commit passes


CatEek:
This should be reviewed after #1027

veegalinova (approved):
Took me some time to get my head around it, but I think everything is correct and well executed 👍

PR #1031, opened 2026-07-28
Chore: replace mypy with pyrefly

veegalinova:
## Disclaimer

- [ ] I am an AI agent.
- [X] I have used AI and I thoroughly reviewed every line.

## Description

> [!NOTE]
> **tldr**: This PR is for switching from mypy type checker to a newer pyrefly checker

I want to test how a more modern type checker would work with our code. After some research, currently there are 3 alternatives people prefer - pyright, ty and pyrefly, the last two beeing more "hyped". I selected pyrefly, which promises to be faster than pyright and has experimental pytorch tensor shape type checking. ty is also a great alternative from astral ecosystem, but it is less mature and with less checks implemented.

To see the type errors in vscode you only need to install "Pyrefly - Python Language Tooling" extension.

### Overview - what changed?

The PR has two parts:

1. Changes to pre-commit and CI, will explain in deatil in "changes made" section
2. A lot of small changes cleaning up type-ignore annotations that are longer needed and resolving new found errors. This I will skip explaining and listing, but I'm happy to go in detail if needed.

Major difference between mypy and pyrefly - now to run the type checker we need a full environment with all dependencies, so it can check imports and fetch types from external libraries. This makes it unusable in pre-commit CI.

- `pre-commit-config.yaml` - added pyrefly. When ran locally, type checker runs with the installed environment. Skipped in CI.
- `ci.yml`
  - moved the type checker from pre-commit to a separate CI step that is ran before the testing matrix. Testing waits for it to finish successfully, so it will still fail fast as pre-commit did and prevent a more expensive matrix checking.
  - switched "cancel previous runs" from external action to github native functionality
  - switched default permissions to "read" for safety reasons (see [guide](https://github.com/github/awesome-copilot/blob/main/instructions/github-actions-ci-cd-best-practices.instructions.md#3-least-privilege-for-github_token))
- `pyproject.toml` - pyrefly settings now are defined here. The checking preset is `legacy` which is recommended for switching from mypy. Only two optional logging library imports are ignored (wandb and tensorboard)
- codecov - set patch status to informational, patch % is still posted but no longer fails CI

## How has this been tested?

After switching, ~20 type ignores were removed as unneeded, and ~60 new errors were resolved. I can not say that the new errors are magically super useful and more meaningful than mypy so far, need more experience with the tool.

Locally, I did not see any meaningful speed difference when calling checker from console, but the IDE integration is much faster than mypy.
CI timings - before, pre-commit took 15-20s, now it took 8. But, the new action takes 45 seconds, from which the checking itself is 1 second and the rest env setup etc. So purely speed-wise the new check is a net negative in terms of CI speed unfortunately.

---

- [X] Code builds and passes tests locally, including doctests
- [X] Pre-commit passes


jdeschamps:
I merged a pre-commit update that created a conflict in this PR.

src/careamics/lightning/grouped_index_sampler.py
jdeschamps, line 1: What happened to that module? Was it not used anymore?

jdeschamps (approved):
Looks good. Some fixes are a bit verbose, but we can live with that. I'd be ok to try it for a while and reevaluate if we don't like it.

src/careamics/lightning/grouped_index_sampler.py
veegalinova: It got duplicated somehow. There is the same file at `src/careamics/lightning/data/grouped_index_sampler.py` which is the one being imported.

PR #1033, opened 2026-08-02
Refac(MSplit): LVAE module

CatEek:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]
> **tldr**: Behavior-preserving cleanup of the LVAE model — dead code removal, layer blocks cleanup, and `models/lvae/` back to pre-commits.

Lot's of old commented-out code, unnecessary inheritance, dead parameters, lot's of pre-commit errors

### Overview - what changed?

- Removed dead/commented code and the resolved TODOs across `lvae.py`, `layers.py`, `stochastic.py`, `utils.py`.
- Specialized the layer blocks to the single used config: dropped the dead-configurable params (`block_type`, `batchnorm`, `conv2d_bias`, `res_block_kernel`, `merge_type`, skip-padding) and their branches.
- Collapsed unnecessary inheritance: removed `ResBlockWithResampling`, `ResidualGatedBlock`, `SkipConnectionMerger`. `BottomUpDeterministicResBlock`/`TopDownDeterministicResBlock` are now standalone and share a `_make_pre_conv` helper.
- Docstrings → numpydoc, typing cleaned up, and the `models/lvae/` pre-commit exclusions removed

- `src/careamics/models/lvae/lvae.py`, `layers.py`, `stochastic.py`, `utils.py` — cleanup + specialization + docstrings/typing.
- `src/careamics/config/architectures/lvae_config.py` — removed dead/broken `set_3D`.
- `.pre-commit-config.yaml` — narrowed the lvae exclusions to `noise_models.py` only.
- `tests/models/lvae/test_lvae_architecture.py` — updated the (already-skipped) `test_bottom_up_pass` to call `bottomup_pass` after the wrapper was inlined.

### Removed features or files

- `ResBlockWithResampling`, `ResidualGatedBlock`, `SkipConnectionMerger`, `ModelType`, `power_of_2`, `LVAEConfig.set_3D`.

---

- [x] Pre-commit passes


CatEek:
I'm very open to suggestion, maybe missed something

src/careamics/config/architectures/lvae_config.py
```diff
@@ -174,20 +174,6 @@ def validate_z_dims(cls, z_dims: tuple) -> tuple:
 
         return z_dims
 
-    def set_3D(self, is_3D: bool) -> None:
```
jdeschamps, line 177: We should also have another look at the UNet and whether this set 3D is ever used or necessary. Maybe a small clean up issue.

src/careamics/models/lvae/layers.py
```diff
@@ -18,27 +18,39 @@
 NormType = Union[nn.BatchNorm2d, nn.BatchNorm3d]
 DropoutType = Union[nn.Dropout2d, nn.Dropout3d]
 
+# Module-level default non-linearity (stateless), used as a shared default argument.
```
jdeschamps, line 21: Useless comments from LLM should b e systematically removed.

```suggestion
```

jdeschamps (commented):
I did not review the PR, just had a quick look.

Could you comment on how you tested correctness? It is part of the PR template and it would be useful to know whether the existing model weights can still be loaded, and/or whether you ran some training.

src/careamics/config/architectures/lvae_config.py
CatEek: created issue

CatEek:
> I did not review the PR, just had a quick look.
> 
> Could you comment on how you tested correctness? It is part of the PR template and it would be useful to know whether the existing model weights can still be loaded, and/or whether you ran some training.

Old checkpoints trained by Ashesh won't load in any version of careamics without some manipulation with weights. We have that in reproducibility. 
Older careamics versions will almost load here(although not with strict=True) but that doesn't matter because we don't have any useful checkpoints

jdeschamps:
Can you report on training performances?

CatEek:
> Can you report on training performances?

~ 31.1 / 32.2, given the full(400ep) original run produced 32/32.9

veegalinova:
> > Can you report on training performances?
> 
> ~ 31.1 / 32.2, given the full(400ep) original run produced 32/32.9

I think it's a substantial difference. If the performance was checked with a script similar to one in #1052, there is no seed set anywhere during training (please correct me if I'm wrong!). Can you re-run two comparison runs with the same seed, and maybe cuda deterministic and benchmark for a good measure?
I will also do the same for HDN and report here.

CatEek:
> > > Can you report on training performances?
> > 
> > 
> > ~ 31.1 / 32.2, given the full(400ep) original run produced 32/32.9
> 
> I think it's a substantial difference. If the performance was checked with a script similar to one in #1052, there is no seed set anywhere during training (please correct me if I'm wrong!). Can you re-run two comparison runs with the same seed, and maybe cuda deterministic and benchmark for a good measure? I will also do the same for HDN and report here.

seeds are very good point, will add. As for the numbers, also forgot that those were with mmse of 1, with 50 it jumps to 31.6/32.8. Running full 400ep also

src/careamics/config/architectures/lvae_config.py
jdeschamps: Following https://github.com/CAREamics/careamics/issues/1038, this should not be removed and we should see what else needs to be updated for the LVAE architecture to be set to 3D.

Indeed, we need to have an HDN napari plugin for a grant.

jdeschamps:
As discussed, before merging we are waiting for full report of the experiments (original PSNR, this branch with seeds and mmse) and for @veegalinova tests on HDN.

CatEek:
After fixing the test subset according to the paper setup the metrics on full training(400ep) are 31.87 ± 0.592 / 32.88 ± 1.162 for ht_lif24_5ms
Reference obtained with original checkpoint is 32.03 ± 0.662 / 32.93 ± 1.211

veegalinova (approved):

PR #1039, opened 2026-08-11
Refac(config): remove unused `set_3D` methods

### LINKED ISSUE #1038: Investigate set_3D in Unet
## Question

We need to have another look at the UNet and whether this set 3D is ever used or necessary

jdeschamps:
## Description

`config.set_3D` was thought as an early mechanism to ease interactions with the configuration. It is clear now that the configuration will hardly ever be used that way, and that creating is new from a factory function is easier.

This PR removes `set_3D` from `Configuration`, model configs, and data config.

Closes #1038


mese79:
Please see my comment on the issue [here](https://github.com/CAREamics/careamics/issues/1038#issuecomment-5253811670)

PR #1056, opened 2026-08-19
 Feat(MicroSplit API): Careamist piping and uncertainty prediction

veegalinova:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: Adds MMSE and uncertainty prediction for HDN and MicroSplit), and the plumbing that lets Careamist run them. 

Currently in stochastic algorithms prediction of mmse and std of multiple samples is not reachable from Careamist. 
While wiring it, I also had to implement the "piping" necessary to add both HDN and MicroSplit as possible methods in Careamist. This PR addresses two concerns at the same time, but they were coupled, so I could not separate this PR into two. 

### Overview - what changed?

Two major decisions in this PR (Please lmk if you disagree on them!): 
- The number of repeated predictions in stochastic methods is not a property of a specific algorithm, but an option of the prediction, like tile size or tile overlaps. 
- To return type from Careamist changes when uncertainty is requested. Implemented with overload if `n_predictions` parameter is set.

- `HDNModule` and `MicroSplitModule` implement `predict_sample` (one draw, denormalised), and `predict_step` feeds it to the shared `mmse_and_sample_std`, which accumulates mean and std in a running fashion. Memory doesn't grow with `n_predictions`, but individual samples aren't retrievable.
- HDN and MicroSplit algorithm configs no longer have `mmse_count` field
- Namings - `n_predictions` and `uncertainty` are user facing names, `mmse`, `n_samples`,`sample_std` stay internal. Calling a sample std "uncertainty" isn't strictly accurate, but I find it less confusing. 
- `Careamist` - `predict` method now has new overloads with a new optional parameter - `n_predictions`. If it is not set, the method returns `predictions, sources`, and if set, the return type changes to `predictions, uncertainties, sources`. I think this is an easiest way to implement this, but I don't like that the return type is different based on parameter set, open to better ideas
- HDN method saves "best" checkpoint based on loss ranking, important for prediction correctness
- Added necessary classes to allow Careamics to use HDN and MicroSplit algos, nothing interesting here 

- `careamist.py` - added new overloads for `n_predictions` and related output types, added related setup and prediction unpacking methods
- `hdn_algorithm_config.py`, `microsplit_algorithm_config.py`, `microsplit_factory.py` - removed `mmse_count` parameter
- `config/factories/config_discriminators.py` - added `HDNAlgorithm` and `MicroSplitAlgorithm`
- `training_configuration.py` - added HDN to the list of methods that save "best" checkpoint and not "last"
- `hdn_module.py`, `microsplit_module.py`, `module_util.py` - implement a mmse prediction wrapper and individual methods, change prediction method return type
- `convert_prediction.py` - functions to unstack mmse and std predictions

## How has this been tested?

I added new tests and tested Careamist HDN training on Convalaria dataset. I am doing proper benchmark experiments and adding related files on the rest of datasets reported in the paper.

## Related Issues

- Resolves #

## Breaking changes

- New possible return type from Careamist - `predictions, uncertainties, sources` if `n_predictions` parameter is set 

---

- [x] Pre-commit passes


src/careamics/lightning/callbacks/prediction/prediction_writer_callback.py
```diff
@@ -194,16 +197,18 @@ def write_on_batch_end(
         """
         Write predictions at the end of a batch.
 
-        Writing method is determined by the attribute `write_strategy`.
+        Writing method is determined by the attribute `write_strategy`. Only the
```
jdeschamps, line 200: Let's open an issue about saving the uncertainty as well using an optional parameter.

src/careamics/models/constraints/model_constraints.py
```diff
@@ -37,15 +37,15 @@ def validate_target_channels(self, n_channels: int) -> None:
...
+    def validate_spatial_shape(self, input_shape: Sequence[int]) -> None:
         """Whether the given spatial shape is compatible with the model constraints.
 
         Shape must be of length 2 (YX) or 3 (ZYX). To validate channel dimension, use
         `validate_input_channels` or `validate_target_channels` instead.
 
         Parameters
         ----------
-        shape : Sequence[int]
+        input_shape : Sequence[int]
```
CatEek, line 48: MInor but why do we need to change the name?

CatEek (approved):
The only thing I noticed is technically stitching std comes with more issues than simply visual tiling artifacts, total uncertainty won't represent exactly the same as if we take whole image std. Maybe we can investigate this further later

src/careamics/models/constraints/model_constraints.py
veegalinova: It's now `input_shape` in lvae and unet constrains and `shape` in protocol, so for consistency.

veegalinova:
> The only thing I noticed is technically stitching std comes with more issues than simply visual tiling artifacts, total uncertainty won't represent exactly the same as if we take whole image std. Maybe we can investigate this further later

Thanks for the review! 
You raised an interesting point, I investigated a bit - looked at several datasets' prediction uncertainty on full image and on tiled image side by side, and I did not see any stitching artifacts. But the totals are indeed different, if I calculated it correctly, tiled prediction adds 5-10% uncertainty. Probably each tile is just less context, so the possible value distribution is larger.
But it's certainly a good idea to investigate it further.

src/careamics/careamist.py
```diff
@@ -952,12 +1063,25 @@ def predict(
...
+
+        if n_predictions is None:
+            return predictions_output, sources
+
+        uncertainties = cast(
+            "list[ImageRegionData]",
+            [uncertainty_region(batch) for batch in predictions],
+        )
+        uncertainty_output, _ = convert_prediction(
+            uncertainties,
```
jdeschamps: I am wondering whether we could again in clarity by refactoring this away from `careamist.py` into a `careamist_utils`.

For now it is fine here, although I would add some comments because it will be hard to read without context in the future.

jdeschamps (approved):
Thanks @veegalinova !

The `CAREamist` starts to be complicated to read, but currently I am not sure it is avoidable.

src/careamics/lightning/callbacks/prediction/prediction_writer_callback.py
veegalinova: Opened #1072

src/careamics/careamist.py
veegalinova: I tried to clean it up by creating a new `convert_predict_outputs` function, can you have a look?

jdeschamps (approved):
Looks good!

PR #986, opened 2026-06-29
Feat(Seg): left-over changes

jdeschamps:
## Description

> [!IMPORTANT]  
> This PR contains the remnants of the changes that were not split into the multiple sub-PRs:
> - #1061
> - #1062
> - #1067
> - #1068
> - #1081

Introduces the backbone for semantic segmentation in CAREamics. While we are predominantly a restoration library, semantic segmentation in a codebase that we know inside out is valuable for the facility and potentially our users. In addition CAREamics has strong arguments with the zarr support, and we could benefit from that in segmentation projects.

### Input format

`targets` should be passed as class labels and are assumed to have same `axis` as the inputs, albeit without a `C` dimension.

### Internal change to model constraints

The main departure from the previous behaviour is that in a segmentation task the number of model output channels is different from the targets. The former are probabilities with `N` channels, while the `target` has class labels (singleton channel after reshaping). In order to pass the model constraints, I needed to add `model_constraints` that are algorithm-aware, meaning that `get_model_constraints` now take the `algorithm_config` rather than the `model_config`.

```python
# before
def get_model_constraints(model_config: UNetConfig) -> ModelConstraints:

# now
def get_model_constraints(algorithm_config: AlgorithmConfig) -> ModelConstraints:
```

### Bulk of changes

The bulk of changes pertains to making all the necessary segmentation classes and modules:
- Segmentation configuration
- Segmentation algorithm configuration
- Segmentation configuration factory
- Segmentation Lightning module
- Segmentation losses

### Out-of-scope changes

Note that this PR also adds the following out-of-scope features: 
- Adding missing `target_axes` to the prediction functions in `CAREamist`. 
- Added an example folder (`/examples/`), from which `.gitignore` ignores all files but `.md`, `.py` and `.ipynb`
- e2e CI running only smoke/e2e tests from tests/e2e
- Fixed pre-commit, `int` casting in `xy_flip.py`

⚠️ This PR includes #1058 and should therefore be reviewed after merging it.

### API

Nothing changed there:

```python
    cfg = create_seg_config(
        experiment_name="test_seg",
        data_type="array",
        axes="SYX",
        batch_size=4,
        patch_size=(64, 64),
        num_epochs=20,
        n_classes=n_classes,
    )

    careamist = CAREamist(config=cfg, work_dir=tmp_path)
    careamist.train(
        train_data=train_data,
        train_data_target=train_data_tar,
        val_data=val_data,
        val_data_target=val_data_tar,
    )

    # predict to memory
[...]

PR #1062, opened 2026-08-24
Feat(Segmentation): add losses

jdeschamps:
Add semantic segmentation losses.

This PR adds `DiceLoss`, `CrossEntropyLoss` and `DiceCELoss` (Dice+CrossEntropy) losses. All losses assume that the input to the `forward` method are:
- `inputs`: raw model output as logits of dims (B, C, [Z], Y, X) where `C` has dimension `num_classes` (including background)
- `targets`: class labels with singleton channel of dims (B, 1, [Z], Y, X).

Dice loss converts the input to probabilities (softmax) and the targets to one-hot encoding, while cross-entropy simply removes from the target the singleton `C` dimension and converts to `long` (necessary for `torch.nn.functional.cross_entropy`).

### Added

- Segmentation losses to `segmentation_losses`
- Corresponding tests

## Notes

Typing added in https://github.com/CAREamics/careamics/pull/1081.

### DIFF
--- a/src/careamics/losses/__init__.py
+++ b/src/careamics/losses/__init__.py
@@ -3,6 +3,7 @@
 __all__ = [
     "denoisplit_loss",
     "denoisplit_musplit_loss",
+    "get_seg_loss",
     "hdn_loss",
     "lvae_loss_factory",
     "musplit_loss",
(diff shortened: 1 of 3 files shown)


jdeschamps:
Note: this PR was merged in #1068 as it was needed there, then it is modified by #1081.

jdeschamps:
Superseded by https://github.com/CAREamics/careamics/pull/1068

PR #1063, opened 2026-08-24
Feat(API): add `target_axes` to prediction

jdeschamps:
`target_axes` is missing from `CAREamist` prediction functions, so when changing `axes` at prediction time users may not be able to predict since `axes` and `target_axes` can disagree. 

This was missed during the PRs that introduced `target_axes`.

### DIFF
--- a/src/careamics/careamist.py
+++ b/src/careamics/careamist.py
@@ -796,6 +796,7 @@ class CAREamist:
         tile_size: tuple[int, ...] | None = None,
         tile_overlap: tuple[int, ...] | None = (48, 48),
         axes: str | None = None,
+        target_axes: str | None = None,
         data_type: Literal["array", "tiff", "zarr", "czi", "custom"] | None = None,
         # ADVANCED PARAMS
         num_workers: int | None = None,
@@ -815,6 +816,7 @@ class CAREamist:
         tile_size: tuple[int, ...] | None = None,
         tile_overlap: tuple[int, ...] | None = (48, 48),
         axes: str | None = None,
+        target_axes: str | None = None,
         data_type: Literal["array", "tiff", "zarr", "czi", "custom"] | None = None,
         # ADVANCED PARAMS
         num_workers: int | None = None,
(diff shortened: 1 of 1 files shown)


src/careamics/careamist.py
```diff
@@ -871,6 +874,9 @@ def predict(
             Overlap between tiles, can be None.
         axes : str, optional
             Axes of the input data, by default None.
+        target_axes : str | None, default=None
+            Axes for prediction target data. If None, uses target axes from the training
+            configuration.
```
veegalinova, line 879: This got me a bit confused. There is no target data for the prediction, so this parameter will be used for determining the order of axes for the final output, right? If so, the docstring is a bit misleading.

src/careamics/careamist.py
```diff
@@ -1078,6 +1088,9 @@ def predict_to_disk(
             Overlap between tiles.
         axes : str, optional
             Axes of the input data, by default None.
+        target_axes : str | None, default=None
+            Axes for prediction target data. If None, uses target axes from the training
+            configuration.
```
veegalinova, line 1093: And here it is different, there is optional `pred_data_target`, but I could not find where / if it is used anywhere? Looks like here the target is loaded, validated, then dropped. And then the axis order is not fixed before writing I think? (`restore_shape` is defaulted to false in writing strategies)

veegalinova (commented):
I have questions about this parameter meaning in `predict` vs `predict_to_disk` functions.

src/careamics/careamist.py
veegalinova: Ah, I now see that you fixed the restore_shape=True in #1065

jdeschamps: I did not even remember the `pred_data_target`. I opened an issue: https://github.com/CAREamics/careamics/issues/1080

src/careamics/careamist.py
jdeschamps: yeah I see how that is confusing. I renamed the user facing parameter to `output_axes`: https://github.com/CAREamics/careamics/pull/1063/changes/d643e832735bad0662ee84517e45b5a1da4ca531

veegalinova (approved):

PR #1065, opened 2026-08-24
Feat(CI): add e2e CI

jdeschamps:
This PR adds e2e CI and normalizes behaviour of writing arrays to disk.

## Main features

> [!WARNING]  
> Changes from #986 touched on these particular tests and were included here.

- New e2e CI that only runs `tests/e2e` in a single CI, these tests being excluded from other CIs.
- Tile and image writing strategies now restore the shape of the arrays/tiles before writing. 
    - Without these changes, the resulting arrays written to disk will have singleton channels. 
    - A consequence is that `predict` and `predict_to_disk` results will have different shape.
    - If channels were present in the original data (or target), then the singleton channels are kept as is the behaviour of `restore_shape`.

### Added

- `.github/ci_e2e.yml` gh action
- `tests/e2e/conftest.py` to mark all e2e tests

### Modified

- `pyproject.toml` declares the `e2e` pytest mark
- `smoke_tests` renamed `lightning_api` e2e tests
- `image_write_strategy` and `tile_write_strategy` now do `restore_shape=True`
- Related tests are updated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         run: uv sync --all-extras --dev
 
       - name: 🧪 Run tests
-        run: uv run pytest --color=yes -m "not lvae and not compat"
+        run: uv run pytest --color=yes -m "not lvae and not e2e"
 
 
   deploy:
--- a/.github/workflows/ci_e2e.yml
+++ b/.github/workflows/ci_e2e.yml
@@ -1,0 +1,48 @@
+name: CI-e2e
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    # run every week (for --pre release tests)
+    - cron: "0 0 * * 0"
+
+jobs:
+  test:
+    name: e2e py3.12 (ubuntu-latest)
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12"]
+
+    steps:
+      - name: 🛑 Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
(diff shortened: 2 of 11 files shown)


.github/workflows/ci_e2e.yml
```diff
@@ -0,0 +1,48 @@
+name: CI-e2e
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+  pull_request:
```
veegalinova, line 9: Is the intent here to run these tests on every PR but only on one platform ?

.github/workflows/ci.yml
veegalinova, line 1: Should these tests also be excluded from the codecov ci?

.github/workflows/ci_e2e.yml
jdeschamps: Yes. I go back and forth between either considering that all the deep features are tested on all platforms and there is no need to also run the e2e on all platforms, and thinking that it is not much more expensive to also run on one python and all platforms.

.github/workflows/ci.yml
jdeschamps: I kept them in the codecov because they may increase coverage.

In general we would need to rediscuss our CI strategy, the CI takes in general too long for my taste.

veegalinova: True, but also if I understand correctly it will trigger an extra e2e test run, which might not be desirable if they would require a lot of time

veegalinova (approved):

.github/workflows/ci.yml
jdeschamps: I can remove it from the codecov indeed, but I think it is more the CodeCov CI itself that I'd like to change in the near future.

jdeschamps: Done in https://github.com/CAREamics/careamics/pull/1065/changes/01049efa4af99dfb28f5c75a34f4c370d92bcd6e

PR #1070, opened 2026-08-25
Fix: stitching back tiles when input and output have different number of channels

### LINKED ISSUE #948: Stitching tiled predictions fails when prediction channels are different from input's channel

jdeschamps:
#993 failed to close #948 because it did not address tile stitching (outside of Zarr) but only combining entire samples. This PR introduces a new function to adjust data shapes and use it in the stitching utilities to create arrays of the correct size.

### Added

- `adjust_shape_for_channels` in `reshape_array.py`: a method to update a data shape with a new channel dimension size given the axes
- Relevant unit and e2e tests

### Modified

- `dataaset.py`: a similar step was performed in the dataset, it has been updated to use the new `adjust_shape_for_channels` function.
- `stitch_single_prediction` and `stitch_single_sample` were updated to adjust the channels of the receiving array shape.

## Resolves

Closes #948

### DIFF
--- a/src/careamics/dataset/dataset.py
+++ b/src/careamics/dataset/dataset.py
@@ -15,6 +15,7 @@ from careamics.config.data.data_config import (
 )
 from careamics.dataset.augmentation.compose import Compose
 from careamics.models.constraints import ModelConstraints
+from careamics.utils.reshape_array import adjust_shape_for_channels
 
 from .image_region_data import ImageRegionData
 from .image_stack import GenericImageStack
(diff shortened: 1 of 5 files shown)


jdeschamps:
I may have missed one additional function in dataset to refactor. Will fix later.

[EDIT]: done in https://github.com/CAREamics/careamics/pull/1070/changes/29a850c89c5842f54b1b48df3a05ff649d5ce1c4

mese79 (approved):
I've tested the lambdaSplit prediction with this branch. All went ok!

PR #1084, opened 2026-09-09
Feat(MicroSplit API): HDN model fixes

veegalinova:
## Disclaimer

- [ ] I am an AI agent.

## Description

> [!NOTE]  
> **tldr**: adding back parameters that are important for correct HDN training

While refactoring, some of the parameters that were meaningful for HDN training were removed. I re-introduced them one by one, checking that the metric on the paper dataset improved for each one. I made sure that the MicroSplit defaults are not changed.

### Overview - what changed?

Parameters re-introduced: 
- `encoder_first_conv_kernel` - 5 for HDN, 3 for MicroSplit. 
- `analytical_kl` - kl loss formula differs between HDN and MicroSplit 
- `enable_topdown_normalize_factor` - deeper layers in MicroSplit are normalized with a coefficient, should be off in HDN

Fixes: 
- `process_q_params` cropped odd latents with a center_crop with single int size, applied to both spatial dimensions. A non-square latent became square, and a 1x1 latent cropped to size 0. The crop now handles height and width separately and skips a dimension of size 1
- removed dead `allow_oddsizes` parameter 
- `LVAEConfig` now sets extra="forbid", so unknown fields are not accepted silently. Was bug inducing cause of the parameters renaming and refactoring. 

Configs: all three fields are new on LVAEConfig and are set by default. The HDN and MicroSplit factories set all three and a user cannot override them through the factories (happy to revisit this if you see the need).

- `lvae_config.py`: added the params
- `hdn_factory.py`, `microsplit_factory.py` - set the new parameter in non-editable way 
- `stochastic.py` - fix the center cropping issue

## How has this been tested?

Added tests for the model parameters verification, tested training for HDN.

---

- [x] Pre-commit passes

### DIFF
--- a/src/careamics/models/lvae/lvae.py
+++ b/src/careamics/models/lvae/lvae.py
@@ -60,6 +60,9 @@ class LadderVAE(nn.Module):
     analytical_kl : bool
         Whether to compute the KL divergence analytically instead of by a single-sample
         Monte Carlo estimate.
+    enable_topdown_normalize_factor : bool
+        Whether to scale the inference parameters of the i-th top-down layer for depth
+        stabilization. Only applied when `z_dims` holds more than 4 entries.
 
     Raises
     ------
@@ -84,6 +87,7 @@ class LadderVAE(nn.Module):
         decoder_blocks_per_layer: int = 1,
         encoder_first_conv_kernel: int = 5,
         analytical_kl: bool = False,
+        enable_topdown_normalize_factor: bool = True,
     ):
         super().__init__()
 
@@ -104,6 +108,7 @@ class LadderVAE(nn.Module):
         self.nonlin = nonlinearity
         self.predict_logvar = predict_logvar
         self.analytical_kl = analytical_kl
+        self.enable_topdown_normalize_factor = enable_topdown_normalize_factor
         # -------------------------------------------------------
 
         # -------------------------------------------------------
@@ -128,7 +133,6 @@ class LadderVAE(nn.Module):
         self.mode_pred = False
         self._var_clip_max = 20
         self._stochastic_use_naive_exponential = False
-        self._enable_topdown_normalize_factor = True
 
         # Attributes that handle LC -> Hardcoded
         self.enable_multiscale = self._multiscale_count > 1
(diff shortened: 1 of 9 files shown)


src/careamics/models/lvae/lvae.py
```diff
@@ -56,6 +56,14 @@ class LadderVAE(nn.Module):
...
     decoder_blocks_per_layer : int
         The number of residual blocks per decoder layer.
+    encoder_first_conv_kernel : int
+        The kernel size of the first bottom-up convolution.
+    analytical_kl : bool
+        Whether to compute the KL divergence analytically instead of by a single-sample
+        Monte Carlo estimate.
+    enable_topdown_normalize_factor : bool
+        Whether to scale the inference parameters of the i-th top-down layer for depth
+        stabilization. Only applied when `z_dims` holds more than 4 entries.
```
jdeschamps:
```suggestion
    encoder_first_conv_kernel : int, default=3
        The kernel size of the first bottom-up convolution.
    analytical_kl : bool, default=False
        Whether to compute the KL divergence analytically instead of by a single-sample
        Monte Carlo estimate.
    enable_topdown_normalize_factor : bool, default=True
        Whether to scale the inference parameters of the i-th top-down layer for depth
        stabilization. Only applied when `z_dims` holds more than 4 entries.
```

jdeschamps: also there's another occurrence later

tests/config/architectures/test_lvae_config.py
```diff
@@ -102,3 +102,24 @@ def test_parameters_wrong_values_by_assigment():
...
+    model = LVAEConfig(architecture="LVAE")
+
+    assert model.enable_topdown_normalize_factor is True
+
+
+def test_encoder_first_conv_kernel_default():
+    """Test that encoder_first_conv_kernel defaults to 3."""
+    model = LVAEConfig(architecture="LVAE")
+
+    assert model.encoder_first_conv_kernel == 3
```
jdeschamps: I would kick out these tests, just testing defaults does not bring much

jdeschamps (approved):
Just very small comments that can be ignored!

